PR #13, opened 2015-04-24
Updated to the latest schema 

TheAzureGuy:


singhkay:
@TheAzureGuy Unrelated to this PR, but for your templates I suggest the following itemDisplayName so that they show up better in the Azure.com index column and the text is not cut off

Redis cluster on Ubuntu VMs
MongoDB on Ubuntu VMs
Nagios Core on Ubuntu VMs
PostgreSQL 9.3 on Ubuntu VMs

and use the descriptions or summary for more details

PR #19, opened 2015-04-25
Updating for new api and metadata

gbowerman:


apache2-on-ubuntu-vm/metadata.json
```diff
@@ -0,0 +1,7 @@
+{
+  "itemDisplayName": "Deploy an Ubuntu VM with Apache webserver",
```
singhkay: Suggestion for name: Apache Webserver on Ubuntu VM

PR #50, opened 2015-04-27
Moved to new repo and updated to the new API version

GONZALORUIZ:


couchbase-ansible/metadata.json
```diff
@@ -0,0 +1,7 @@
+{
+  "itemDisplayName": "Advanced Linux : Deploy a Multi VM Couchbase Cluster with Ansible",
```
singhkay: Suggested name: Couchbase cluster with Ansible so that it displays better in the Azure.com index title and the text is not cut off.

couchbase-ansible/metadata.json
```diff
@@ -0,0 +1,7 @@
...
+  "description": "This advanced template deploys N  Linux VMs (CentsOS or Ubuntu) and it configures a Multi VM Couchbase Cluster.",
+  "summary": "This advanced template deploys a Multi VM Couchbase cluster.It uses the Ansible Playbooks provided by couchbaselabs in the GitHub repo ansible-couchbase-server to deploy Couchbase.",
```
singhkay: Summary looks too long according to the validation character limit, but description looks right if you just want to switch them

couchbase-ansible/metadata.json
GONZALORUIZ: Is Multi VM Couchbase cluster with Ansible Ok?

singhkay: Sure. I think Ansible won't get cut off with your title, but we can change later if it does.

PR #54, opened 2015-04-27
Create python proxy template 

yuezh:
A template that allow you to quickly deploy a http proxy server to azure.

### DIFF
--- a/python-proxy-on-ubuntu/README.md
+++ b/python-proxy-on-ubuntu/README.md
@@ -1,0 +1,24 @@
+# Install Python Proxy on a Ubuntu Virtual Machine using Custom Script Linux Extension
+
+<a href="https://azuredeploy.net/" target="_blank">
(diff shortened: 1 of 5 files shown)


singhkay:
Hi @yuezh couple of requests because we're deprecating the old API 2014-12-01-preview

• Update API Version for all Compute, Network and Storage resources to 2015-05-01-preview on all templates
• Update Load Balancer API the latest design
• Remove Default values for storage account etc., since it fails when using it with PowerShell
• Update Source Image specification model to follow - https://github.com/azurermtemplates/azurermtemplates/blob/master/101-simple-vm-from-image/azuredeploy.json 
• Add OsDisk specification even when creating from an image and use the "createoperation" element
• Update DataDisk specification to have "CreatOption" element

Here's another guide for an example https://github.com/Azure/azure-quickstart-templates/blob/master/breaking-change.md

yuezh:
Close this PR and will reopen it after test the new API

PR #62, opened 2015-04-27
Create python proxy template 

yuezh:
A template that allow you to quickly deploy a http proxy server to azure.

### DIFF
--- a/python-proxy-on-ubuntu/azuredeploy.json
+++ b/python-proxy-on-ubuntu/azuredeploy.json
@@ -28,12 +28,26 @@
                 "Description": "Password for the Virtual Machines" ,
             } 
         },
-        "vmSourceImageName": {
+        "imagePublisher": {
             "type": "string",
-            "defaultValue": "b39f27a8b8c64d52b05eac6a62ebad85__Ubuntu-14_04_2_LTS-amd64-server-20150309-en-us-30GB",
-            "metadata": { 
-                "Description": "Source Image Name for the VM. Example: b39f27a8b8c64d52b05eac6a62ebad85__Ubuntu-12_04_5-LTS-amd64-server-20140927-en-us-30GB" 
-            } 
+            "defaultValue": "Canonical",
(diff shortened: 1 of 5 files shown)


python-proxy-on-ubuntu/azuredeploy.json
```diff
@@ -0,0 +1,263 @@
+{
+    "$schema": "http://schema.management.azure.com/schemas/2014-04-01-preview/deploymentTemplate.json",
+    "contentVersion": "1.0.0.0",
+    "parameters": {
+        "newStorageAccountName": {
+            "type": "string",
+            "defaultValue": "uniquestorageaccountName",
```
singhkay: Can you remove the default value for storage account since these need to be unique DNS names?

python-proxy-on-ubuntu/azuredeploy.json
```diff
@@ -0,0 +1,263 @@
...
+        "vmSize": {
+            "type": "string",
+            "defaultValue": "Standard_A0",
+            "metadata": { 
+                "Description": "Size of the Virtual Machine"
+            } 
+        },
+        "publicIPAddressName": {
+            "type": "string",
+            "defaultValue": "myPublicIP",
```
singhkay: Same regarding the default value for this since these need to be unique

python-proxy-on-ubuntu/azuredeploy.json
```diff
@@ -0,0 +1,263 @@
...
+        "newStorageAccountName": {
+            "type": "string",
+            "defaultValue": "uniquestorageaccountName",
+            "metadata": { 
+                "Description": "Unique DNS Name for the Storage Account where the Virtual Machine's disks will be placed" 
+            } 
+        },
+        "dnsNameForPublicIP": {
+            "type": "string",
+            "defaultValue": "uniquednsnameforpublicip",
```
singhkay: Same for this as well since this needs to be unique

PR #65, opened 2015-04-27
Adding Docker swarm template

kmouss:


docker-swarm-cluster/azuredeploy.json
```diff
@@ -0,0 +1,220 @@
+{
+    "$schema": "http://schema.management.azure.com/schemas/2014-04-01-preview/deploymentTemplate.json",
+    "contentVersion": "1.0.0.0",
+    "parameters" : {
+        "newStorageAccountName": {
+            "type": "string"
```
singhkay: @kmouss Can you add metadata description tags to all the parameters?

kmouss:
Thank you for merging, I will push a lower priority fix PR to add the metadata for other fields.

ahmetalpbalkan:
@kmouss 

    This template sets up a Docker Swarm cluster in a VM.

this statement seems like inaccurate. Do any of those VMs actually join the Swarm cluster or form a cluster?

 Looks like it doesn't but I might be wrong as I just took a peek. We may add some capabilities to Docker extension to make slave nodes join to a swarm cluster. Check that w/ us.

PR #75, opened 2015-04-27
migrate symantec to new repo and api

sung-msft:


symantec-extension-windows-vm/metadata.json
```diff
@@ -0,0 +1,7 @@
+{
+  "itemDisplayName": "Create a Windows VM with Symantec Endpoint Protection extension enabled",
```
singhkay: Suggested name: Symantec Endpoint Protection extension on Windows VM

PR #76, opened 2015-04-27
updating to may header

KundanaP:


windows-puppet-agent/azuredeploy.json
```diff
@@ -3,51 +3,89 @@
...
+             "type": "String",
+             "metadata" : {
+               "Description" : "Puppet Master URL"
+             }
+         }
+
     },
     "variables": {
-        "vmSourceImageName" : "a699494373c04fc0bc8f2bb1389d6106__Windows-Server-2012-R2-201412.01-en.us-127GB",
+        "vmSourceImageName" : "b39f27a8b8c64d52b05eac6a62ebad85__Ubuntu-14_10-amd64-server-20141204-en-us-30GB",
```
singhkay: can you remove tis since you don't need it with image reference?

windows-puppet-agent/azuredeploy.json
```diff
@@ -3,51 +3,89 @@
...
         "sourceImageName" : "[concat('/',subscription().subscriptionId,'/services/images/',variables('vmSourceImageName'))]",
```
singhkay: Same with this, not needed anymore

PR #97, opened 2015-04-28
Integrated changes to blueprints for hortonworks-on-centos template

predeekc:


hortonworks-on-centos/azuredeploy.json
```diff
@@ -126,7 +126,7 @@
             "nameNodeCount": 1,
             "dataNodeCount": 3,
             "masterVmSize": "Standard_A3",            
-            "masterNodeTemplateUri": "/cluster-node-16disk.json",
+            "masterNodeTemplateUri": "/cluster-node-8disk.json",
             "workerVmSize": "Standard_A3",
```
johngossman: Do we want to use D sizes?

hortonworks-on-centos/scripts/mountDisk.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+DRIVE=$1
+INDEX=$2
+
+echo "Creating filesystem on $DRIVE..."
+mke2fs -F -t ext4 -b 4096 -O sparse_super,dir_index,extent,has_journal -m1 $DRIVE
+
+echo "Mountint disk $DRIVE at /disks/$INDEX"
```
johngossman: Mounting?

hortonworks-on-centos/azuredeploy.json
predeekc: Not sure what you mean by D, but this change is correct.  It should have been using only 8 disks with the A3 VM

hortonworks-on-centos/scripts/mountDisk.sh
predeekc: Fixing this now, good catch

predeekc:
The spelling error is fixed

PR #98, opened 2015-04-28
Updated template URL

dennisan:


johngossman:
lgtm -- I am CR-ing all the templates, so watch for other feedback

PR #104, opened 2015-04-28
Include parameter docs in README of drone-ubuntu-vm template

anweiss:
Update README to include parameter docs

Signed-off-by: Andrew Weiss <andrew.weiss@microsoft.com>


johngossman:
lgtm

PR #106, opened 2015-04-28
Updating readme and fixing an issue with deploy.json

KundanaP:


johngossman:
Can you also update the descriptions of the parameters in the template itself? And change "Description" to "description"?

PR #109, opened 2015-04-28
Fixing deployment problems

gbowerman:


lamp-app/azuredeploy.json
```diff
@@ -1,181 +1,186 @@
...
-          }
+                "Description": "Unique DNS Name for the Storage Account where the Virtual Machine's disks will be placed."
+            }
         },
         "adminUsername": {
             "type": "string",
             "metadata": {
-              "Description": "User name of the Linux admin user who can log on to the machine."
-          }
+               "Description": "User name for the Virtual Machine."
```
johngossman: nit: should be "description" instead of "Description"

lamp-app/azuredeploy.json
```diff
@@ -1,181 +1,186 @@
...
-                    },
-                    "subnet": {
-                        "id": "[variables('subnetRef')]"
+            }
+        },
+        {
+            "apiVersion": "2015-05-01-preview",
+            "type": "Microsoft.Network/virtualNetworks",
+            "name": "[variables('virtualNetworkName')]",
+            "location": "[parameters('location')]",
```
johngossman: There is no such parameter...should be variables.

PR #117, opened 2015-04-28
Moved to new repo

simongdavies:
Updated VM SKUs and added storage options


kenazk:
Can you remove the main README changes so we can add the templates? We can fix the central README later.

singhkay:
@simongdavies  looks like we might have moved your templates already https://github.com/Azure/azure-quickstart-templates/tree/master/activedirectorynewdomain-ha-2-dc
https://github.com/Azure/azure-quickstart-templates/tree/master/activedirectorynewdomain

Do you want to remove these or override the files in place?

PR #120, opened 2015-04-28
move gluster fs template to new azure repo

liupeirong:


gluster-file-system/azuredeploy.json
```diff
@@ -0,0 +1,318 @@
...
+                "description": "ssh user name"
+            }
+        },
+        "adminPassword": {
+            "type": "securestring",
+            "metadata": {
+                "description": "ssh password"
+            }
+        },
+        "imagePublisher": {
```
johngossman: Since all these have only one allowed value, just hardcode them rather than make them parameters.

gluster-file-system/azuredeploy.json
```diff
@@ -0,0 +1,318 @@
...
+            "type": "string",
+            "defaultValue": "6.5",
+            "allowedValues": [ 
+                "6.5"
+            ],
+            "metadata": {
+                "description": "the SKU name of the image offering typically major OS version"
+            }
+        },
+        "vmSize": {
```
johngossman: Should these be D sizes?

gluster-file-system/azuredeploy.json
```diff
@@ -0,0 +1,318 @@
...
+                "Standard_A1",
+                "Standard_A2",
+                "Standard_A3",
+                "Standard_A4"
+            ],
+            "metadata": {
+                "description": "VM size for the nodes"
+            }
+        },
+        "vmNamePrefix": {
```
johngossman: This doesn't need to be a parameter

gluster-file-system/azuredeploy.json
```diff
@@ -0,0 +1,318 @@
...
+            }
+        },
+        "vnetAddressPrefix": {
+            "type": "string",
+            "defaultValue": "10.0.0.0/16",
+            "metadata": {
+                "description": "virtual network CIDR in which the cluster will be created"
+            }
+        },
+        "gfsSubnetName": {
```
johngossman: Just hardcode

gluster-file-system/azuredeploy.json
```diff
@@ -0,0 +1,318 @@
...
+            }
+        },
+        "gfsSubnetPrefix": {
+            "type": "string",
+            "defaultValue": "10.0.0.0/24",
+            "metadata": {
+                "description": "CIDR of the subnet for the Gluster nodes"
+            }
+        },
+        "customScriptFilePath": {
```
johngossman: Just make this a variable

gluster-file-system/azuredeploy.json
```diff
@@ -0,0 +1,318 @@
...
+            "type": "string",
+            "defaultValue": "https://raw.githubusercontent.com/azure/azure-quickstart-templates/master/gluster-file-system/azuregfs.sh",
+            "metadata": {
+                "description": "GitHub location of the custom script that configures Gluster"
+            }
+        },
+        "customScriptCommandToExecute": {
```
johngossman: variable or hardcode

gluster-file-system/azuredeploy.json
```diff
@@ -0,0 +1,318 @@
...
+                        {
+                            "id": "[resourceId('Microsoft.Network/networkInterfaces',concat(variables('nicName'), copyindex()))]"
+                        }
+                    ]
+                }
+            }
+        },
+        {
+            "type": "Microsoft.Compute/virtualMachines/extensions",
+            "name": "[concat(parameters('vmNamePrefix'), copyindex(), '/extention')]",
```
johngossman: typo: "extension"

gluster-file-system/azuredeploy.json
liupeirong: If user already has VMs with the same name in this resource group, the deployment will fail, right?

gluster-file-system/azuredeploy.json
liupeirong: if I hardcode it, can a user use an existing subnet with the matching CIDR?

gluster-file-system/azuredeploy.json
liupeirong: My bash script doesn't necessarily configure the linux VM exactly the user wanted.  User should be able to copy the bash script to another location, and customize it. If I make it a variable, they can't do that any more.

gluster-file-system/azuredeploy.json
liupeirong: User can customize the bash script, or rename it, or change its parameters.

liupeirong:
Fixed all the issues other than those I commented.  Will separate OS disk to another storage account later.

PR #126, opened 2015-04-29
Corrected incorrect Default - keep minimum 1.2.1

CognosysTech:


johngossman:
lgtm...I wondered about this, but don't know enough about Spark

PR #127, opened 2015-04-29
Moved discover private ip template

singhkay:


singhkay:
Moving over the template for now, more fixes coming to readme

PR #129, opened 2015-04-29
Added SQL - Django Project

meet-bhagdev:
This template uses the Azure Linux CustomScript extension to deploy a Django application by creating an Ubuntu VM, doing a silent install of Python and Apache, then creating a simple Django application. The template also creates a SQL Database, with a sample table with some sample data which displayed in the web browser using a query

Once the template is deployed, just grab the FQDN of the VM and open it in a browser and you'll see the sample data displayed on your screen. Make sure you wait 10-15 minutes after the tempalate is deployed before you access the DNS name. There might be a lag when accessing the dns in your browser. Please refresh your browser a few times to mitigate it.

### DIFF
--- a/sqldb-django-on-ubuntu/azuredeploy.json
+++ b/sqldb-django-on-ubuntu/azuredeploy.json
@@ -1,0 +1,249 @@
+{
+    "$schema": "http://schema.management.azure.com/schemas/2014-04-01-preview/deploymentTemplate.json",
+    "contentVersion": "1.0.0.0",
+    "parameters": {
+        "newStorageAccountName": {
+            "type": "string",
+            "metadata": {
+                "description": "Name of the Storage Account"
+            }
+        },
+        "location": {
+            "type": "string",
+            "metadata": {
+                "description": "Location of all resources created by this template."
+            }
+        },
+        "adminUsername": {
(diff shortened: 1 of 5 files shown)


sqldb-django-on-ubuntu/azuredeploy.json
```diff
@@ -0,0 +1,249 @@
...
+    {
+        "type": "Microsoft.Storage/storageAccounts",
+        "name": "[parameters('newStorageAccountName')]",
+        "apiVersion": "2014-12-01-preview",
+        "location": "[parameters('location')]",
+        "properties": {
+            "accountType": "[variables('storageAccountType')]"
+        }
+    }, {
+        "apiVersion": "2014-12-01-preview",
```
singhkay: Please move to 2015-05-01-preview API version header for storage, network and compute resources

sqldb-django-on-ubuntu/metadata.json
```diff
@@ -0,0 +1,7 @@
+{
+  "itemDisplayName": "Deploy a Django app with SQL Databases",
```
singhkay: Suggestion: Django app with SQL Databases would look better in Azure.com index

PR #133, opened 2015-04-29
Submitting two templates

varunmaker:


101-vm-simple-linux-vm-data-disk/README.md
```diff
@@ -0,0 +1,21 @@
+# Very simple deployment of an Windows VM with a single empty Data Disk and VM Diagnostic extension
```
singhkay: Linux VM here

varunmaker: Thanks for the catch! Removed the extension.

PR #156, opened 2015-04-29
Fixed "String" should be "string"

singhkay:


singhkay:
closing will resubmit

PR #166, opened 2015-04-29
Build Environment of Vnet/LoadBal/PubIp/NSG/Nat rules/VM/Disks

danSthearm:
This is the first template version, building an environment (PubIP, LB,
NSG, Av Sets) to create a VM where it can be RDP'd externally.


danSthearm:
problem with metadata...fixing

danSthearm:
Reopening - problem with schema validation....

PR #171, opened 2015-04-29
move docker swarm template to stage repo

kmouss:
move docker swarm template to stage repo until we support automatically
adding nodes to the cluster


johngossman:
lgtm thanks

PR #186, opened 2015-04-29
Removed Extraneous commas

simongdavies:


singhkay:
Merging for testing. Will still CR

PR #202, opened 2015-04-30
Moved location of security group attachment in NIC

predeekc:


singhkay:
lgtm merging but @predeekc Please submit a metadata.json file. I had removed it because Mahesh requested for removing but we have a better way of blacklisting the template now until it's ready. So the metadata file can be added back in.

PR #217, opened 2015-05-03
moved MySQL HA cluster from personal repo to new Azure repo

liupeirong:
Updated with latest schema and metadata.


liupeirong:
The automatic validtor isn't working properly - 
1. Every parameter already has a description property, 
2. the template is tested live, so it's valid.  
3. only the readme.md is updated in the root

mysql-ha-pxc/azuredeploy.json
```diff
@@ -0,0 +1,796 @@
...
+                                    "id": "[variables('sshNatRuleID3')]"
+                                }
+                            ]
+                        }
+                    }
+                ]
+            }
+        },
+        {
+            "name": "[variables('vmName1')]",
```
singhkay: Not a blocker but if these are similar VMs, could you use a copyIndex function and make this more than a 3 node cluster if that's a valid scenario?

singhkay:
@liupeirong Please ignore the bot for now, it's looking for "Description" instead of "description". The team is working on a fix to the pipeline

PR #224, opened 2015-05-04
Apprenda 6.0 Small Cluster

dutronlabs:
The small cluster template provides the ability for Azure customers to provision a cross-platform-supported environment, within the confines of the Apprenda Express license. This allows for both .NET and generic linux (ie. Java, Docker) applications to run.

Large cluster is incoming, need to finalize how the license is on-boarded during provisioning (otherwise, it's the same template, just larger).


apprenda60-small/metadata.json
```diff
@@ -0,0 +1,7 @@
+{
+  "itemDisplayName": "Apprenda 5.5 Single Node",
```
singhkay: Does this need to be updated?

apprenda60-small/metadata.json
```diff
@@ -0,0 +1,7 @@
...
+  "description": "This is the express version of Apprenda, with a single Windows VM.",
+  "summary": "Apprenda 5.5 Express",
```
singhkay: This as well?

singhkay:
@dutronlabs Template looks good but you might want to update the metadata.json to reflect Apprenda 6.0

dutronlabs:
should be updated for 6.0, also has some changes in here that represent fixes based on feedback from Kundana.

singhkay:
@dutronlabs Looks good, merged

PR #231, opened 2015-05-05
Split data disks into containers and fixed script logging error on cloud...

predeekc:
...era-on-centos


singhkay:
@predeekc Is this and hortonworks template ready to roll? We should probably remove the "not ready" message from readme and add metadata.json file https://github.com/Azure/azure-quickstart-templates/tree/master/cloudera-on-centos

PR #238, opened 2015-05-06
Added new Sharepoint HA template

simongdavies:
Added a new template that deploys a 9 VM HA Sharepoint farm


sharepoint-server-farm-ha/azuredeploy.json
```diff
@@ -0,0 +1,1985 @@
...
+				"East US",
+				"West Europe",
+				"East Asia",
+				"Southeast Asia"
+			],
+			"metadata": {
+				"Description": "The region to deploy the resources into"
+			}
+		},
+		"virtualNetworkName": {
```
johngossman: This doesn't need to be a parameter.

sharepoint-server-farm-ha/azuredeploy.json
```diff
@@ -0,0 +1,1985 @@
...
+			"defaultValue":"10.0.0.5"
+		},
+		"sqlLBIPAddress": {
+			"type": "string",
+			"metadata": {
+				"Description": "The IP address of the new SQL ILB"
+			},
+			"defaultValue":"10.0.0.6"
+		},
+		"adVMPrefix": {
```
johngossman: Why parameterize this? Why not just hardcode it?

sharepoint-server-farm-ha/azuredeploy.json
```diff
@@ -0,0 +1,1985 @@
...
+			"type": "string",
+			"metadata": {
+				"Description": "The Prefix of the AD VM name"
+			},
+			"defaultValue":"dc"
+		},
+		"sqlVMPrefix": {
```
johngossman: Hardcode?

sharepoint-server-farm-ha/azuredeploy.json
```diff
@@ -0,0 +1,1985 @@
...
+			"type": "string",
+			"metadata": {
+				"Description": "The Prefix of the SQL VM Name"
+			},
+			"defaultValue":"sql"
+		},
+		"spVMPrefix": {
```
johngossman: Hardcode?

sharepoint-server-farm-ha/azuredeploy.json
```diff
@@ -0,0 +1,1985 @@
...
+				"Standard_DS13",
+				"Standard_D14",
+				"Standard_DS14"
+			],
+			"metadata": {
+				"Description": "The size of the SP VMs Created"
+			},
+			"defaultValue": "Standard_DS2"
+		},
+		"adImagePublisher": {
```
johngossman: Take a look at https://github.com/Azure/azure-quickstart-templates/blob/master/101-simple-windows-vm-data-disk/azuredeploy.json
And consider simplifying the parameterization here.

sharepoint-server-farm-ha/azuredeploy.json
```diff
@@ -0,0 +1,1985 @@
...
+				"Description": "The DNS Prefix for the Public IP Address for RDP"
+			}
+		},
+		"spCentralAdminDNSPrefix": {
+			"type":"string",
+			"metadata": {
+				"Description": "The DNS Prefix for the Public IP Address for the Central Admin site"
+			}
+		},
+		"rdpPort":{
```
johngossman: How many people actually change this? For those people does it make sense to have them edit the template?

sharepoint-server-farm-ha/azuredeploy.json
```diff
@@ -0,0 +1,1985 @@
...
+			{
+				"name": "[variables('spAppSubnetName')]",
+				"properties": {
+					"addressPrefix": "[parameters('spAppSubnet')]"
+				}
+			}
+		],
+		"vnetwithDNSTemplateUri":"[concat(parameters('assetLocation'),'/vnet-with-dns-server.json')]",
+		"nicTemplateUri": "[concat(parameters('assetLocation'),'/nic.json')]",
+		"adPDCModulesURL":"[concat(parameters('assetLocation'),'/CreateADPDC.ps1.zip')]",
```
johngossman: Total nit: some of these are listed as Uris, some as URLs.

sharepoint-server-farm-ha/azuredeploy.json
```diff
@@ -0,0 +1,1985 @@
...
+			"location": "[parameters('location')]",
+			"dependsOn":[
+				"[concat('Microsoft.Compute/virtualMachines/', variables('adBDCVMName'))]"
+			],
+			"properties": {
+				"publisher": "Microsoft.Powershell",
+				"type": "DSC",
+				"typeHandlerVersion": "1.7",
+				"settings": {
+					"ModulesUrl": "[variables('adBDCModulesURL')]",
```
johngossman: nit: camelCase (assuming DSC extension supports camelCase).

sharepoint-server-farm-ha/vnet-with-dns-server.json
```diff
@@ -0,0 +1,64 @@
+{
+	"$schema": "http://schema.management.azure.com/schemas/2014-04-01-preview/deploymentTemplate.json",
```
johngossman: Update the schema version here too

sharepoint-server-farm-ha/vnet-with-dns-server.json
```diff
@@ -0,0 +1,64 @@
...
+				"Description": "The DNS address(es) of the DNS Server(s) used by the VNET"
+			}
+		},
+	},
+	"resources": [
+		{
+			"name": "[parameters('virtualNetworkName')]",
+			"type": "Microsoft.Network/virtualNetworks",
+			"location": "[parameters('deploymentLocation')]",
+			"apiVersion": "2014-12-01-preview",
```
johngossman: Should use the 2015 api version

sharepoint-server-farm-ha/nic.json
```diff
@@ -0,0 +1,51 @@
+{
```
johngossman: Just curious why you choose to split out these two as subtemplates when they are so small, but not some of the VMs for example.

sharepoint-server-farm-ha/azuredeploy.json
```diff
@@ -0,0 +1,1985 @@
+{
+	"$schema": "http://schema.management.azure.com/schemas/2014-04-01-preview/deploymentTemplate.json",
+	"contentVersion": "1.0.0.0",
+	"parameters": {
+		"newStorageAccountName": {
```
johngossman: You are running all the VMs and their data disks out of one storage account. Is that sufficient for the IO?

sharepoint-server-farm-ha/azuredeploy.json
```diff
@@ -0,0 +1,1985 @@
...
+							}
+						}
+					}
+				]
+			}
+		},
+		{
+			"apiVersion": "2015-05-01-preview",
+			"type": "Microsoft.Compute/virtualMachines",
+			"name": "[variables('adPDCVMName')]",
```
johngossman: PDC == Primary Domain Controller?

sharepoint-server-farm-ha/azuredeploy.json
```diff
@@ -0,0 +1,1985 @@
...
+						"value": [
+								"[parameters('adPDCNICIPAddress')]","[parameters('adBDCNICIPAddress')]"
+						]
+					}
+
+				}
+			}
+		},
+		{
+			"type": "Microsoft.Compute/virtualMachines/extensions",
```
johngossman: For readability does it make sense to move these closer the VM they are working on?

sharepoint-server-farm-ha/azuredeploy.json
```diff
@@ -0,0 +1,1985 @@
...
+						"DomainName": "[parameters('domainName')]",
+						"AdminCreds":{
+							"UserName": "[parameters('adminUserName')]",
+							"Password": "PrivateSettingsRef:AdminPassword"
+						},
+						"SharePointSetupUserAccountcreds":{
+							"UserName": "[parameters('sharePointSetupUserAccountUserName')]",
+							"Password": "PrivateSettingsRef:SharePointSetupUserAccountPassword"
+						},
+						"SharePointFarmAccountcreds":{
```
johngossman: Why is this null?

sharepoint-server-farm-ha/azuredeploy.json
```diff
@@ -0,0 +1,1985 @@
...
+							"UserName": "[parameters('sqlServerServiceAccountUserName')]",
+							"Password": "PrivateSettingsRef:SqlServerServiceAccountPassword"
+						},
+						"DatabaseName":"[parameters('configDatabaseName')]",
+						"AdministrationContentDatabaseName":"[parameters('administrationContentDatabaseName')]",
+						"DatabaseServer":"[concat(variables('sqlVMName'),'1.',parameters('domainName'))]",
+						"SqlAlwaysOnAvailabilityGroupName":"[variables('sqlAOAGName')]",
+						"PrimaryReplica":"[concat(variables('sqlVMName'),'1')]",
+						"SecondaryReplica":"[concat(variables('sqlVMName'),'0')]",
+						"databaseNames":null,
```
johngossman: I take it the secondary has to have this null, but not the primary. Can you have one template and call it with two different databaseNames properties? You could even have an array in the variables section and base whether this is null on the copyIndex.

sharepoint-server-farm-ha/nic.json
simongdavies: I did this out of necessity rather than to improve modularity\reuse etc. In order to update a resource that I have previously created or updated in a template then I have to have a linked template to do this as it is not possible to specify the same resource twice in a template.

I can look at breaking this up into more subtemplates but I would rather do that as part of the refactoring once the issues with DSC extension are addressed

sharepoint-server-farm-ha/azuredeploy.json
simongdavies: Good feedback, I don't think we are going to hit any issues as there are only 11 disks in use but it wont do any harm - other than use quota - to separate storage accounts for AD, SQL and SP. I plan to add the capability to specify the number of SP VMs once DSC is fixed so this will help there.

sharepoint-server-farm-ha/azuredeploy.json
simongdavies: Yes, maybe my terminology is a little out of date :-) I will check.

sharepoint-server-farm-ha/azuredeploy.json
simongdavies: This is another artefact of DSC issues forcing me to combine resource operations that really should be separate, the parameter is mandatory on the script and the DSC resource it invokes and I didn't want to create a different configuration script for different, I will be able to make this go away once DSC extension works correctly

sharepoint-server-farm-ha/azuredeploy.json
simongdavies: Same issue as the previous null argument

sharepoint-server-farm-ha/azuredeploy.json
simongdavies: I made all the image related parameters variables, as I have only tested this with the default values it doesn't make much sense for these to be parameters

sharepoint-server-farm-ha/nic.json
johngossman: Ah, yes. I remember this issue from when we first tried to do Sharepoint long ago.

PR #246, opened 2015-05-08
updated region/location default values and options

trentmswanson:


singhkay:
@trentmswanson looks good, thanks for the defaults. Our bot overlord is happy now :) Merging

PR #251, opened 2015-05-08
update template create-hpc-cluster (add one more template to give user more options, and supply new function)

justintian:
update template create-hpc-cluster 
1, add one more template to give user more options, 
2, supply new function: user can specify post config script for head node and compute node


justintian:
have updated to solve the issue

PR #259, opened 2015-05-08
Update azuredeploy.json

squillace:
Removing bad default values. If you specify a default value, two things happen. 1) you will not be prompted for one on the command line; 2) if the value -- in this case "" -- does not match one of the accepted values, you're never, ever going to be able to deploy it. :-) and :-(

with these two changes, locally, template deployment succeeds.

### DIFF
--- a/redis-high-availability/azuredeploy.json
+++ b/redis-high-availability/azuredeploy.json
@@ -16,14 +16,12 @@
         },
         "storageAccountName": {
             "type": "string",
-            "defaultValue": "",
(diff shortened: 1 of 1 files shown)


squillace:
@singhkay that's a strange validation. Unless it's just saying that I can't make a direct change, and you can.... :-)

singhkay:
@squillace That's fine :) We're looking into adding a contributor tag into the metadata and updating the bot that will do away with this

PR #260, opened 2015-05-10
fix one issue in README.md for create-hpc-cluster

justintian:


singhkay:
@justintian How is the second template run? Can these two be independent templates in different folders?

justintian:
yes, the second template can run standalone, the first one use published image, the second one can let user to use their private(custom image), the custom script extension are same for both template, so I think it is better to put them together, so user can see two options in one page

singhkay:
@justintian So in the second one, you can run a custom user image on compute node right? The reason for having two folders is so that these templates can be indexed and run separately from the Azure.com index http://azure.microsoft.com/en-us/documentation/templates/

justintian:
@singhkay thanks, I got it, I will create separate folders for the second template, will let you know when ready

justintian:
@singhkay  I already add separate folder, please take a look to see whether can be merged, thanks

singhkay:
@justintian Left few comments. One other nit, can you rename the folder create-hpc-cluster-custom-image for more clarity

justintian:
@singhkay  thanks, just update the name

singhkay:
@justintian Looks good, thanks!

PR #267, opened 2015-05-12
remove unused paramters

qingfuwang:


qingfuwang:
@johngossman please take a look, cfip and boship will be used in future.

johngossman:
I don't see boship, microboship or ip parameters being used anywhere

PR #273, opened 2015-05-14
Linux vm w cache

vflorusso:


linux-vm-with-cache/Microsoft.Cache.1.0.4.json
```diff
@@ -0,0 +1,46 @@
+{
```
singhkay: This looks like it can be a part of the main azurdeploy.json template?

linux-vm-with-cache/azuredeploy.json
```diff
@@ -0,0 +1,230 @@
+{
+    "$schema": "https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#",
+    "contentVersion": "1.0.0.0",
+    "parameters": {
+        "location": {
+            "type": "string",
+            "defaultValue": "West US",
```
singhkay: Remove default location and add allowed values like in https://github.com/Azure/azure-quickstart-templates/blob/master/101-create-availability-set/azuredeploy.json

linux-vm-with-cache/azuredeploy.json
```diff
@@ -0,0 +1,230 @@
...
+                    "networkInterfaces": [
+                        {
+                            "id": "[resourceId('Microsoft.Network/networkInterfaces',variables('nicName'))]"
+                        }
+                    ]
+                }
+            }
+        },
+        { 
+            "apiVersion": "2015-01-01",
```
singhkay: This can be removed and the cache template integrated directly here

linux-vm-with-cache/metadata.json
```diff
@@ -0,0 +1,7 @@
+{
+  "itemDisplayName": "Deploy a simple Linux VM in West US",
```
singhkay: Suggestion: Something along the lines of "Redis Cache with a Linux VM" so that that the app name is used first in the Azure.com index

PR #275, opened 2015-05-15
Add lazy init flag for the mke2fs

thomas1206:
Add lazy init flag to speed up the disk formatting.
Use localhost instead of hostname.


singhkay:
@thomas1206 Looks like there are merge conflicts. Can you resolve?

thomas1206:
The change will be submitted by Newdesic.

PR #280, opened 2015-05-18
Deis template updated for 2015-05-01-preview API

HaishiBai:


deis/azuredeploy.json
```diff
@@ -0,0 +1,225 @@
...
+            } 
+        },
+		"numberOfNodes":{
+			"type": "int",
+            "metadata": { 
+                 "description": "Number of member nodes"
+            },
+			"defaultValue": "3"
+		},
+		"imagePublisher": {
```
singhkay: Consider making these image related parameters into variables if this template takes a hard dependency on CoreOS. You could also have pub and offer hardocoded to CoreOS as variables and give the user a choice of SKU flavors as allowed values.

deis/azuredeploy.json
```diff
@@ -0,0 +1,371 @@
...
+            }
+        },
+        "vmNamePrefix": {
+            "type": "string",
+            "defaultValue": "coreos",
+            "metadata": {
+                "description": "Prefix for the names of the VMs"
+            }
+        },
+        "virtualNetworkName": {
```
singhkay: This can become a variable and one less parameter for the user to enter since it is only RG unique

deis/azuredeploy.json
```diff
@@ -0,0 +1,371 @@
...
+        "lbID": "[resourceId('Microsoft.Network/loadBalancers',variables('loadBalancerName'))]",
+        "lbAPIRuleID": "[concat(variables('lbID'),'/loadBalancingRules/',variables('loadBalancerAPIRuleName'))]",
+        "lbBuilderRuleID": "[concat(variables('lbID'),'/loadBalancingRules/',variables('loadBalancerBuilderRuleName'))]",
+        "lbIPConfig": "[concat(variables('lbID'),'/frontendIPConfigurations/',variables('loadBalancerIPConfigName'))]",
+        "apiProbeID": "[concat(variables('lbID'),'/probes/',variables('apiProbeName'))]",
+        "builderProbeID": "[concat(variables('lbID'),'/probes/',variables('builderProbeName'))]",
+        "lbPoolID": "[concat(variables('lbID'),'/backendAddressPools/',variables('lbBackendAddressPoolName'))]"
+    },
+    "resources": [
+        /* Availability Set Definition */
```
singhkay: No comments in JSON please, Azure CLI breaks

deis/azuredeploy.json
```diff
@@ -0,0 +1,371 @@
...
+            "name": "[parameters('newStorageAccountName')]",
+            "location": "[resourceGroup().location]",
+            "apiVersion": "2015-05-01-preview",
+            "properties": {
+                "accountType": "[variables('storageAccountType')]"
+            }
+        },
+        /* Public IP Address for the LB */
+        {
+            "apiVersion": "2014-12-01-preview",
```
singhkay: Networking resources need 2015-05-01-preview API as well

singhkay:
@HaishiBai Can you rename the template folder to something like "deis-cluster-coreos" so that it is more clear in case we have another deis template submission in the future

deis-cluster-coreos/azuredeploy.json
```diff
@@ -0,0 +1,361 @@
...
+    "$schema": "http://schema.management.azure.com/schemas/2014-04-01-preview/deploymentTemplate.json",
+    "contentVersion": "1.0.0.0",
+    "parameters": {
+        "newStorageAccountName": {
+            "type": "string",
+            "metadata": {
+                "description": "Name of the of the storage account for VM OS Disk"
+            }
+        },
+        "vmNamePrefix": {
```
johngossman: This doesn't need to be a parameter. Can just hardcode. VMnames are relative to the resource group and only provide addresses.

johngossman: I see. You are using it for the DNS name of the load balancer. It does need to be a parameter, but should be called something like "DNSname" or "Loadbalancer IP DNS name"

deis-cluster-coreos/deploy-deis.ps1
```diff
@@ -0,0 +1,28 @@
+﻿#Requires -Version 3.0
```
johngossman: cli instead?

deis-cluster-coreos/azuredeploy.json
```diff
@@ -0,0 +1,361 @@
...
+                "description": "Public key for SSH authentication"
+            }
+        },
+        "customData": {
+            "type": "string",
+            "metadata": {
+                "description": "Base64-encoded cloud-config.yaml file to deploy and start Fleet"
+            }
+        },
+        "numberOfNodes": {
```
johngossman: Will this work with a number other than 3? The docs say 3. The lbrules seems to imply 3.

HaishiBai: At this point it's limited to 3 because of the inbound NAT rule syntax. I'm following this up on Yammer and will fix if a solution is available.

singhkay: @HaishiBai Can you make the number of nodes a variable? so that the user doesn't accidentally specify another number and the template breaks

deis-cluster-coreos/azuredeploy.json
```diff
@@ -0,0 +1,361 @@
...
+        "newStorageAccountName": {
+            "type": "string",
+            "metadata": {
+                "description": "Name of the of the storage account for VM OS Disk"
+            }
+        },
+        "publicDomainName": {
+            "type": "string",
+            "metadata": {
+                "description": "Prefix for the names of the VMs"
```
johngossman: description should be updated

deis-cluster-coreos/README.md
```diff
@@ -0,0 +1,155 @@
...
+
+##Other parameters
+
+Below are the parameters that the template expects
+
+| Name   | Description    |
+|:--- |:---|
+| location | location where the resources will be deployed |
+| newStorageAccountName | new storage account for the VMs OS disk |
+| vmNamePrefix | prefix for the names of each VM |
```
johngossman: This should be publicDomainName

PR #324, opened 2015-06-10
adds a basic dns name to the single ssh-secured linux vm with ubuntu

squillace:
this is very close to a simple, default v1 vm create but secured by default.


azurecla:
Hi __@squillace__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <p>
        It looks like you're working at Microsoft (rasquill). If you're full-time, we DON'T require a contribution license agreement.
    </p>
    <p>
    </p>

TTYL, AZPRBOT;

squillace:
LGTM. @singhkay, what do you think?

squillace:
LGTM, deploy button targets the azure portal and not the azuredeploy website, which is.... just strange. We need to convert those over. It's very pretty, but it's not what you want.

PR #359, opened 2015-06-27
Initial commit of octopusdeploy3-single-vm-windows.

paulbouwer:
This template allows you to deploy a single **Octopus Deploy 3.0** server with a trial license. This will deploy on a single Windows Server 2012R2 VM (Standard D2) and SQL DB (S1 tier).

The deploy button url in `README.md` is already configured to point to the `Azure / azure-quickstart-templates` repo.

The `Install-OctopusDeploy.ps1` script is also already referenced in the `Azure / azure-quickstart-templates` repo via the `installerUri` variable in `azuredeploy.json`.


azurecla:
Hi __@paulbouwer__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <p>
        It looks like you're working at Microsoft (pabouwer). If you're full-time, we DON'T require a contribution license agreement.
    </p>
    <p>
    </p>

TTYL, AZPRBOT;

octopusdeploy3-single-vm-windows/azuredeploy.json
```diff
@@ -0,0 +1,355 @@
...
+        "Japan East",
+        "Japan West",
+        "North Europe",
+        "South Central US",
+        "Southeast Asia",
+        "West Europe",
+        "West US"
+      ],
+      "metadata": {
+        "Description": "Location of the Octopus Deploy server."
```
singhkay: lower case "Description"

octopusdeploy3-single-vm-windows/azuredeploy.json
```diff
@@ -0,0 +1,355 @@
+{
+  "$schema": "https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#",
+  "contentVersion": "1.0.0.0",
+  "parameters": {
+    "location": {
+      "type": "string",
+      "defaultValue": "West US",
+      "allowedValues": [
```
singhkay: Please remove the default location. Don't believe the VMs are available in all these locations. Choosing some of these will result in an error. Please look at another template with allowed values to see the valid region list atm.

octopusdeploy3-single-vm-windows/azuredeploy.json
```diff
@@ -0,0 +1,355 @@
...
+    },
+    "octopusAdminPassword" : {
+      "type": "securestring",
+      "metadata": {
+        "description": "Admin password for the Octopus Deploy web application."
+      }
+    }
+  },
+  "variables": {
+    "location": "[parameters('location')]",
```
singhkay: This seems redundant. Isn't it better to just use parameters(location) where needed?

octopusdeploy3-single-vm-windows/azuredeploy.json
```diff
@@ -0,0 +1,355 @@
...
+    "vmImagePublisher": "MicrosoftWindowsServer",
+    "vmImageOffer": "WindowsServer",
+    "vmOSDiskName": "osdiskforwindowssimple",
+    "vmWindowsOSVersion": "2012-R2-Datacenter",
+    "vmStorageAccountName": "[parameters('vmStorageAccountName')]",
+    "vmStorageAccountType": "Standard_LRS",
+    "vmStorageAccountContainerName": "vhds",
+    "vmName": "OctopusDeploy",
+    "vmSize": "Standard_D2",
```
singhkay: This should be a parameter giving the user an option to select a size

octopusdeploy3-single-vm-windows/azuredeploy.json
paulbouwer: done.

octopusdeploy3-single-vm-windows/azuredeploy.json
paulbouwer: Removed the default location. 

I built the allowed values for the location by ensuring that the Resource Providers and Resource Types were available using the following PowerShell:

```powershell
PS> $resourceProviders = @("Microsoft.Storage/storageAccounts", "Microsoft.Network/publicIPAddresses", "Microsoft.Network/virtualNetworks", "Microsoft.Network/networkInterfaces", "Microsoft.Compute/virtualMachines", "Microsoft.Sql/servers")
PS> $hash = @{}; Get-AzureLocation |? Name -in $resourceProviders |% { $_.Locations |% { $hash[$_] = $hash[$_] + 1 } }; $hash.Keys |% { New-Object PSObject -Property @{Location = $_; Count = $hash[$_] } } |? { $_.Count -eq $resourceProviders.Count } | Select Location | Sort -Property Location

Location
--------
Central US
East Asia
East US
East US 2
Japan East
Japan West
North Europe
South Central US
Southeast Asia
West Europe
West US
```

octopusdeploy3-single-vm-windows/azuredeploy.json
paulbouwer: This is an approach to abstracting the method of obtaining a value for later use in the template. This allows you to simply use variables in the template for reference. You can decide in the variable section whether you define the value there or obtain it from a parameter. 

This definitely helped me in the stages when I was defining the template and switching between obtaining a value from a parameter or a variable. It may still have value beyond that and it may still be worth showing this as an alternate method for defining parameters and variables.

octopusdeploy3-single-vm-windows/azuredeploy.json
paulbouwer: I initially had this as a parameter, but the user experience of having to understand the **magic strings** when deploying via the Azure Portal felt bad to me. 

It would be great to have access to a lookup function in the ARM Templates that would map the user input (from allowed values list) **Standard D2 (2 cores, 7GB) ** to **Standard_D2** or **Standard A2 (2 cores, 3.5GB)** to **medium**.

In the end I decided to fix the value. This is the suggested spec for a single Octopus VM and this template is aimed at **kicking the tires** on an Octopus 3.0 environment.

I'm happy to make it a limited list but not sure the flexibility (and extra parameter) makes sense here.

paulbouwer:
Renamed `azuredeploy.parameters.json` to `azuredeploy-parameters.json`. If this is a new requirement we should probably walk the templates in this repo and update the examples.

paulbouwer:
When using an empty `azuredeploy-parameters.json` file to validate the templates I get the same error as above ...

```
> armclient post "https://management.azure.com/subscriptions/{subscription-id}/resourceGroups/OctopusTrial01/providers/microsoft.resources/deployments/OctopusTrialDeployment01/validate?api-version=2015-01-01" @validate.json

{
  "error": {
    "code": "InvalidTemplate",
    "message": "Deployment template validation failed: 'The provided value for the template parameter 'location' at line '5' and column '18' is not valid.'."
  }
}
```

But when using a `azuredeploy-parameters.json` with values to validate the template I get succes ...

```
> armclient post "https://management.azure.com/subscriptions/{subscription-id}/resourceGroups/OctopusTrial01/providers/microsoft.resources/deployments/OctopusTrialDeployment02/validate?api-version=2015-01-01" @validate.json

{
  "id": "/subscriptions/{subscription-id}/resourceGroups/OctopusTrial01/providers/Microsoft.Resources/deployments/OctopusTrialDeployment02",
  "name": "OctopusTrialDeployment02",
  "properties": {
    "templateLink": {
      "uri": "https://xxxx.blob.core.windows.net/test/azuredeploy.json",
      "contentVersion": "1.0.0.0"
    },
    "parametersLink": {
      "uri": "https://xxxx.blob.core.windows.net/test/azuredeploy-parameters.json",
      "contentVersion": "1.0.0.0"
    },
    ...
```

Is there something I'm missing ?

octopusdeploy3-single-vm-windows/azuredeploy.json
singhkay: You can have a list of allowed sizes that the user can select from so that the exact size string is visible for the user. In Portal deploy experience, this would be a drop down box.

octopusdeploy3-single-vm-windows/azuredeploy.json
singhkay: But how is this different from just using "[parameters('location')]" in the template? If you were doing some operations on the parameter value to derive a variable (for e.g. concat, toLower, replace etc.) I can understand the value of using "[variables('location')]" instead of "[parameters('location')]" but I'm not seeing what it accomplishes in this scenario. Looking closely there are other variables following the same pattern as well. 

"licenseFullName": "[parameters('licenseFullName')]",
"licenseOrganisationName": "[parameters('licenseOrganisationName')]",
"licenseEmailAddress": "[parameters('licenseEmailAddress')]",
"octopusAdminUsername": "[parameters('octopusAdminUsername')]",
 "octopusAdminPassword": "[parameters('octopusAdminPassword')]",

Adding @johngossman for his input as well

singhkay:
@paulbouwer The error is due to empty values/string in the parameters file. Can you fill out some dummy values in the parameters file for the template validation?

singhkay:
Thanks @paulbouwer! Merged!

PR #374, opened 2015-07-06
Adding a template for customized Minecraft server on Ubuntu

gbowerman:


azurecla:
Hi __@gbowerman__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <p>
        It looks like you're working at Microsoft (guybo). If you're full-time, we DON'T require a contribution license agreement.
    </p>
    <p>
    </p>

TTYL, AZPRBOT;

gbowerman:
I'm not sure if the numbering is funny but I don't see anything wrong with line 152, and this template works fine in my test branch. I do not see why this template is not valid.

gbowerman:
I don't see anything wrong with this template

gbowerman:
To double check I made an exact replica of this template here (https://github.com/gbowerman/azure-minecraft). It deploys perfectly. I do not understand why acomghbot is saying there is an invalid resource.

rgardler:
I don't know how the bot does it's verification but it looks to me like the line number is off by 1 (the error refers to parameter "name" which is 151 not 152). Assuming this is correct then perhaps the problem might be that there is no default value set for the dnsNameForPublicIP parameter.

You could test this by providing a value for the name property in 151.

gbowerman:
Thanks Ross. It likes the pull request once I set a default parameter in the azuredeploy-parameters file. Good catch.
GuyDate: Sun, 5 Jul 2015 22:36:10 -0700

gbowerman:
Hi guys, any update on this pull request?

minecraft-on-ubuntu/azuredeploy.json
```diff
@@ -0,0 +1,281 @@
...
+            "metadata": {
+                "description": "Add a seed for your world"
+            }
+        }
+    },
+    "variables": {
+        "imagePublisher": "Canonical",
+        "imageOffer": "UbuntuServer",
+        "OSDiskName": "osdiskforlinuxsimple",
+        "newStorageAccountName": "[parameters('dnsNameForPublicIP')]",
```
singhkay: Probably make this a parameter. Otherwise somebody could enter myuniquednsname, then it appears as the name of the storage account and the vm

gbowerman:
Amended storage account name to not be confused with DNS name

PR #404, opened 2015-07-17
 Removing resource group name dependency

NikhilRanjanKumar:
1) Removing resource group name dependency in parameter file as this is not required
2) Adding DNS name as well for VM, without this you can't get right certificate as IP of the dynamic machine is not constant

### DIFF
--- a/201-winrm-windows-vm/azuredeploy-parameters.json
+++ b/201-winrm-windows-vm/azuredeploy-parameters.json
@@ -2,6 +2,9 @@
     "newStorageAccountName" : {
         "value":""
     },
+    "dnsNameForPublicIP": {
+        "value": ""
+    },
     "adminUsername": {
         "value": ""
     },
@@ -18,9 +21,6 @@
       "value" : ""
     },
     "vaultName" : {
-      "value" : ""
-    },
-    "vaultResourceGroup" : {
       "value" : ""
     },
     "certificateUrl" : {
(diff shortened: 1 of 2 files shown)


azurecla:
Hi __@NikhilRanjanKumar__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
    </span>

TTYL, AZPRBOT;

201-winrm-windows-vm/azuredeploy-parameters.json
```diff
@@ -20,9 +23,6 @@
     "vaultName" : {
       "value" : ""
     },
-    "vaultResourceGroup" : {
```
singhkay: This is actually required when a vault is in a different resource group. Think about when you want to delete a resource group and it deletes all the resources within, you don't want to accidentally delete your vault and have to set it up again.

201-winrm-windows-vm/azuredeploy.json
```diff
@@ -167,7 +170,7 @@
         "adminPassword": "[parameters('adminPassword')]",
         "secrets": [{
           "sourceVault": {
-            "id": "[resourceId(parameters('vaultResourceGroup'), 'Microsoft.KeyVault/vaults', parameters('vaultName'))]"
```
singhkay: Same here, this is required.

PR #427, opened 2015-07-28
Describe valid chars for newStorageAccountName

aculich:
According to an `AccountNameInvalid` error message I received in the
Azure portal, a "Storage account name must be between 3 and 24
characters in length and use numbers and lower-case letters only."

It would be even more helpful if the templates were validated before
allowing them to be submitted, but at least including this in the
description might help a bit.

### DIFF
--- a/CentOS-2nics-lb-cluster/azuredeploy.json
+++ b/CentOS-2nics-lb-cluster/azuredeploy.json
@@ -17,7 +17,7 @@
         "newStorageAccountName": {
             "type": "string",
             "metadata": {
-                "description": "Unique storage account name"
+                "description": "Unique storage account name. Must be between 3 and 24 characters in length and use numbers and lower-case letters only."
             }
         },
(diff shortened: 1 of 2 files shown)


azurecla:
Hi __@aculich__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
    </span>

TTYL, AZPRBOT;

singhkay:
@aculich Appreciate the fix! Please leave the githubUsername field unchanged and ignore the bot for the moment. I've filed a bug against the bot owners to get it fixed for exactly this scenario :)

rgardler:
The checks from the bot are very misleading and need to be changed. I'll raise an issue in a moment if there isn't one already.

singhkay:
@aculich Since I haven't heard back from you in a while, I decided to address this in #461

PR #445, opened 2015-08-02
High-availability Swarm cluster with Consul

ahmetalpbalkan:
- Multiple swarm-master replicas setup (hardcoded: 3)
- Nodes and managers now use Consul for service discovery
  (removed Docker Hub token:// discovery)
- Statically allocated private IPs for masters
- Add TLS authentication to Swarm and Docker engines
- Masters and slave are now in an availability sets
- Use SSH key for authentication instead of password
- Slaves don't have public IPs, they are accessible via SSH tunneling
  through masters however they are load balanced publicly.
- Masters are behind a public Azure Load Balancer, clients should hit this
  Load Balancer's DNS address.
- Template output now conforms DOCKER_HOST URI format
- Only masters have a public IP address
- Update slave VM size from D1 to A2 (Docker use case generally does not use
  /mnt which D1 gets an SSD for. A2 has the same memory, 2-core CPU and
  twice more IOPS =4x500).
- Read 'location' from resource group instead of hardcoded value

Known issues
- `swarm-slaves-lb` is not getting an IP address allocated. Following up with
  networking team.

Docker Swarm logo

### DIFF
--- a/docker-swarm-cluster/README.md
+++ b/docker-swarm-cluster/README.md
@@ -1,0 +1,102 @@
+# Docker Swarm Cluster
+
+<a href="https://portal.azure.com/#create/Microsoft.Template/uri/https%3A%2F%2Fraw.githubusercontent.com%2FAzure%2Fazure-quickstart-templates%2Fmaster%2Fdocker-swarm-cluster%2Fazuredeploy.json" target="_blank">
+    <img src="http://azuredeploy.net/deploybutton.png"/>
+</a>
+
+*(Built by: [ahmetalpbalkan](https://github.com/ahmetalpbalkan), [garimakhulbe](https://github.com/garimakhulbe))*
+
+This template deploys a [Docker Swarm](http://docs.docker.com/swarm) cluster
+on Azure with multiple manager nodes and specified number of slave nodes in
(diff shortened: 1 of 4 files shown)


azurecla:
Hi __@ahmetalpbalkan__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <p>
        It looks like you're working at Microsoft (ahmetb). If you're full-time, we DON'T require a contribution license agreement.
    </p>
    <p>
    </p>

TTYL, AZPRBOT;

ahmetalpbalkan:
@singhkay any ideas what this error is about?:

> The file azuredeploy.json is not valid. Response from ARM API: BadRequest - {"error":{"code":"InvalidTemplate","message":"Deployment template validation failed: 'The template resource '' at line '102' and column '6' is not valid. The name property cannot be null or empty'."}}

ahmetalpbalkan:
@singhkay PTAL

ahmetalpbalkan:
@singhkay PTAL. should be ready to :shipit: now.

ahmetalpbalkan:
Thanks.

PR #487, opened 2015-08-17
feat(site-to-site-vpn) - Site to Site VPN Connection

sedouard:
I was messing around with cross-cloud integration with AWS and came up with this template which should be reusable for any IPSec Site-to-Site VPN connection. I found that [this virtual network gateway example](https://github.com/Azure/azure-quickstart-templates/tree/master/arm-asm-s2s) was lacking in completeness.

This one does:
- Creates a VNet, Subnet, VNetGateway & VNetConnection
- Provisions a test Ubuntu instance
- Tested to work with AWS VPC

The thing I'm not really certain on is which api version I should use. The current versions I'm using seems to work but I could use some guidance here.

### DIFF
--- a/201-site-to-site-vpn/README.md
+++ b/201-site-to-site-vpn/README.md
@@ -1,0 +1,38 @@
+# Site to Site VPN Connection
+
+<a href="https://portal.azure.com/#create/Microsoft.Template/uri/https%3A%2F%2Fraw.githubusercontent.com%2FAzure%2Fazure-quickstart-templates%2Fmaster%2F201-site-to-site-vpn%2Fazuredeploy.json" target="_blank">
+    <img src="http://azuredeploy.net/deploybutton.png"/>
+</a>
+
+This template will create a Virtual Network, a subnet for the network, a Virtual Netowork Gateway and Connection to your network outside of Azure (defined as your `local` network). This could be anything such as your on-premises network and can even be used with other cloud networks such as [AWS Virtual Private Cloud](https://github.com/sedouard/aws-vpc-to-azure-vnet). It also provisions an Ubuntu instance attached to the Azure Virtual Network so that you can test connectivity.
+
+Please note that you must have a Public IP for your other network's VPN gateway and cannot be behind an NAT.
+
+Although only the parameters in [azuredeploy-parameters.json](./azure-deploy-parameters.json) are necessary, you can override the defaults of any of the template parameters below:
+
+| Name   | Description    |
+|:--- |:---|
+| location | Region where the resources will be deployed |
+| vpnType | Route based, or policy based |
+| subscriptionId | Subscription ID |
+| localGatewayName | Name for gateway connected to other Network |
+| localGatewayIpAddress | Public IP address of other network Gateway |
+| localAddressPrefix | CIDR block of other network address space |
+| virtualNetworkName | Name for new virtual network |
+| azureVNetAddressPrefix | CIDR block for new Azure VNet |
+| subnetName | Name for Azure VM subnet |
+| subnetPrefix | CIDR block for Azure VM subnet |
+| gatewaySubnet | Name for gatway subnet |
+| gatewaySubnetPrefix | CIDR block for Azure gateway subnet |
+| gatewayPublicIPName | Name for public IP resource for the Azure gateway |
+| gatewayName | Name for the gateway connected to the new VNet |
+| connectionName | Name for the new connection between Azure VNet and other network |
+| vmName | Virtual Machine Name |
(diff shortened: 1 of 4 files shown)


azurecla:
Hi __@sedouard__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
    </span>

TTYL, AZPRBOT;

sedouard:
Other things I'm unsure about are supported regions. 

Cool thing about this is that it's friendly for our non-windows friends, too since today you can only do this kind of thing via powershell.

site-to-site-vpn/azuredeploy.json
```diff
@@ -0,0 +1,406 @@
+{
+    "$schema": "http://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json",
+    "contentVersion": "1.0.0.0",
+    "parameters": {
+        "location": {
+            "type": "String",
+            "metadata": {
+                "description": "Region where the resources will be deployed"
+            },
+            "defaultValue": "Central US",
```
singhkay: Please let the user pick a location instead of having a default value

site-to-site-vpn/azuredeploy.json
```diff
@@ -0,0 +1,406 @@
...
+            "metadata": {
+                "description": "Route based or policy based"
+            },
+            "defaultValue": "RouteBased",
+            "allowedValues": [
+                "RouteBased",
+                "PolicyBased"
+            ]
+        },
+        "enableBgp": {
```
singhkay: Looks like there is only one allowed value which doesn't give the user much choice. Might be a better idea to just have this as a variable.

site-to-site-vpn/azuredeploy.json
```diff
@@ -0,0 +1,406 @@
...
+                "Standard_D11",
+                "Standard_D12",
+                "Standard_D13",
+                "Standard_D14"
+            ],
+            "metadata": {
+                "description": "Size of the Virtual Machine."
+            }
+        },
+        "adminUsername": {
```
singhkay: Please let the user choose a username rather than a default value

site-to-site-vpn/azuredeploy.json
```diff
@@ -0,0 +1,406 @@
...
+                "description": "Username for sample VM"
+            }
+        },
+        "adminPassword": {
+            "type": "securestring",
+            "metadata": {
+                "description": "User password for sample VM"
+            }
+        },
+        "imagePublisher": {
```
singhkay: Image related properties could become variables since this is a networking focussed template unless there is a scenario where Ubuntu might not work and the user needs to change the OS.

singhkay:
@sedouard Could you please rename this template to 201-site-to-site-vpn to indicate it's level. Other than that can you take a look at any parameters that could become variables to reduce the number of parameters a user has to enter while deploying a template? 

I'll try to follow up internally with the networking team to see who can help with the api-version question.

site-to-site-vpn/README.md
```diff
@@ -0,0 +1,39 @@
+# Site to Site VPN Connection
+
+<a href="https://azuredeploy.net/" target="_blank">
```
singhkay: The deploy button code needs to have the Ibiza portal deeplink. See here 

https://raw.githubusercontent.com/Azure/azure-quickstart-templates/2d4665806c105354b2ea9dccc19ad5b8b952579e/101-create-key-vault/README.md

sedouard:
Thanks for the fast review @singhkay! I tried to keep the parameters number down by providing default values so it doesn't become to annoying.

I made the changes you mentioned above. Let me know what you hear about the api versions, as well as supported data centers.

singhkay:
@sedouard You can run a command like this to figure out the locations the resource providers are in. This is an example of all resources provider by Microsoft.Network provider

    kay@UbuntuVM:~$ azure provider show -n Microsoft.Network

For gateways current available locations are 

    West US, East US, North Europe, West Europe, East Asia, Southeast Asia, South Central US, Central US, East US 2, Japan East, Japan West

For API Versions, can you try 2015-06-15 for Network, storage, vm resources? For lookup use something like this 
    
    azure provider show -n Microsoft.Network --json

sedouard:
Cool I didn't know about that command. I tested everything out with 2015-06-15 looks like everything but storageAccount resources don't work with that API version just yet (strange), so I just left that one at `2015-05-01-preview`. Besides that works just fine.

Thanks!

singhkay:
@sedouard Looks good, merged!

PR #501, opened 2015-08-19
new template for creating VMs with user specified OS and data images

iandow:
I'm submitting a new template for creating VMs with user specified OS and data images.


azurecla:
Hi __@iandow__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
    </span>

TTYL, AZPRBOT;

azurecla:
__@iandow__, Thanks for signing the contribution license agreement so quickly! Actual humans will now validate the agreement and then evaluate the PR.
<br />Thanks, AZPRBOT;

101-vm-from-user-image-and-datadisk/README.md
```diff
@@ -0,0 +1,25 @@
...
+    <img src="http://azuredeploy.net/deploybutton.png"/>
+</a>
+
+Prerequisite - The Storage Account with the User Image VHD should already exist in the same resource group.
+
+This template allows you to create a Virtual Machines from user specified images for OS and Data disks. This template also deploys a Virtual Network, Public IP addresses and a Network Interface.
+
+Below are the parameters that the template expects
+
+| Name   | Description    |
```
singhkay: This table can be removed. We'll render this on Azure.com from the template

iandow: @singhkay I removed the table and pushed that change back to my master branch. Do I need to resubmit this PR?

101-vm-from-user-image-and-datadisk/README.md
```diff
@@ -0,0 +1,10 @@
+# Create a Virtual Machine from a User Image
+
+<a href="https://portal.azure.com/#create/Microsoft.Template/uri/https%3A%2F%2Fraw.githubusercontent.com%2Fiandow%2Fazure-quickstart-templates%2Fmaster%2F101-vm-from-user-image-and-datadisk%2Fazuredeploy.json" target="_blank">
```
singhkay: This location should be updated to this repo (replace iandow with Azure)

101-vm-from-user-image-and-datadisk/azuredeploy.json
```diff
@@ -0,0 +1,197 @@
...
+  "$schema": "https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#",
+  "contentVersion": "1.0.0.0",
+  "parameters": {
+    "vmName": {
+      "type": "string",
+      "metadata": {
+        "description": "This is the name of your virtual machine"
+      }
+    },
+    "userImageStorageAccountName": {
```
singhkay: It'd be easier to just take in a URI of the image from the user and have them enter less parameters

101-vm-from-user-image-and-datadisk/azuredeploy.json
```diff
@@ -0,0 +1,197 @@
...
+        "description": "This is the name of the container in your storage account"
+      }
+    },
+    "userImageVhdName": {
+      "type": "string",
+      "metadata": {
+        "description": "This is the name of the your customized VHD"
+      }
+    },
+    "userDataImageVhdName": {
```
singhkay: Same for data disk, take in a URI

101-vm-from-user-image-and-datadisk/azuredeploy.json
```diff
@@ -0,0 +1,197 @@
...
+    },
+    "vmSize": {
+      "type": "string",
+      "metadata": {
+        "description": "This is the size of your VM"
+      }
+    }
+  },
+  "variables": {
+    "location": "West US",
```
singhkay: Please allow the user to select a location

101-vm-from-user-image-and-datadisk/azuredeploy.json
```diff
@@ -0,0 +1,197 @@
...
+    "publicIPAddressName": "[concat(parameters('vmName'),'PublicIP')]",
```
singhkay: It'd be better to prefix the template scenario name rather than the vmName for all these variables. See this example https://github.com/Azure/azure-quickstart-templates/blob/master/201-ospatching-extension-on-ubuntu/azuredeploy.json#L161

101-vm-from-user-image-and-datadisk/azuredeploy.json
```diff
@@ -0,0 +1,197 @@
...
+    "publicIPAddressType": "Dynamic",
+    "vnetID": "[resourceId('Microsoft.Network/virtualNetworks',variables('virtualNetworkName'))]",
+    "subnet1Ref": "[concat(variables('vnetID'),'/subnets/',variables('subnet1Name'))]",
+    "userImageName": "[concat('http://',parameters('userImageStorageAccountName'),'.blob.core.windows.net/',parameters('userImageStorageContainerName'),'/',parameters('userImageVhdName'))]",
+    "dataDisk1VhdName": "[concat('http://',parameters('userImageStorageAccountName'),'.blob.core.windows.net/',parameters('userImageStorageContainerName'),'/',parameters('userDataImageVhdName'))]",
+    "osDiskVhdName": "[concat('http://',parameters('userImageStorageAccountName'),'.blob.core.windows.net/',parameters('userImageStorageContainerName'),'/',variables('vmName'),'osDisk.vhd')]"
+  },
+  "resources": [
+    {
+      "apiVersion": "2015-05-01-preview",
```
singhkay: It's a good idea to make this a variable. Makes its easier to move to newer version headers.

101-vm-from-user-image-and-datadisk/azuredeploy.json
```diff
@@ -0,0 +1,197 @@
...
+        },
+        "storageProfile": {
+          "dataDisks": [
+          {
+            "name": "[concat(variables('vmName'),'-dataDisk')]",
+            "lun": 0,
+            "createOption": "FromImage",            
+            "vhd": {
+              "Uri": "[variables('dataDisk1VhdName')]"
+            }
```
singhkay: I think you need to specify the image->uri object for data disks the same way as for the osdisk below

iandow:
I made the suggested changes in my master branch. Is this PR acceptable now?

singhkay:
@iandow Can you address the other comments as well?

iandow:
The disks are specified with URI parameters now. I also parametrized the API version. However, I didn't prefix the template scenario name because I wanted to stay consistent with the format of the 101-vm-from-user-image template, which does not use a scenario name, as shown here:

https://github.com/Azure/azure-quickstart-templates/blob/master/101-vm-from-user-image/azuredeploy.json

singhkay:
@iandow I don't see the URI parameter and API parameter changes in the current files in this PR. The latest change only had the Azure link https://github.com/iandow/azure-quickstart-templates/commit/edbbd4709574bf3cd07076cafdb2866af768e66f 

Since the template name is long, I also suggest shortening it so it fits inside the github filename column. Something like 101-vm-user-image-data-disks.

iandow:
@singhkay I pushed the changes to my local master branch. I thought this PR would automatically include my new commits.  How do I get them included here, or do I need to submit another PR?

iandow:
@singhkay I see I forgot to push my changes.  Now you should see them.  If possible, I'd prefer to keep the template name as is, even though it is a bit long.  I can't think of a better, shorter name.

101-vm-from-user-image-and-datadisk/azuredeploy.json
```diff
@@ -0,0 +1,218 @@
...
+    "virtualNetworkName": "[concat(parameters('vmName'),'VNet')]",
+    "nicName": "[concat(parameters('vmName'),'Nic')]",
+    "addressPrefix": "10.0.0.0/16",
+    "subnet1Name": "Subnet-1",
+    "subnet1Prefix": "10.0.0.0/24",
+    "publicIPAddressType": "Dynamic",
+    "vnetID": "[resourceId('Microsoft.Network/virtualNetworks',variables('virtualNetworkName'))]",
+    "subnet1Ref": "[concat(variables('vnetID'),'/subnets/',variables('subnet1Name'))]",
+    "_comment1": "['http://iantest09.blob.core.windows.net/appinternals/appinternals1001ga-certifiable-20150910-os-2015-09-10.vhd']",
+    "_comment2": "['http://iantest09.blob.core.windows.net/appinternals/appinternals1001ga-certifiable-20150910-datadisk-0-2015-09-10.vhd']",
```
singhkay: Do these comment need to be taken out of the public version of the template?

101-vm-from-user-image-and-datadisk/azuredeploy.json
```diff
@@ -0,0 +1,218 @@
...
+          "computername": "[variables('vmName')]",
+          "adminUsername": "[parameters('adminUsername')]",
+          "adminPassword": "[parameters('adminPassword')]"
+        },
+        "storageProfile": {
+          "dataDisks": [
+          {
+            "name": "[concat(variables('vmName'),'-dataDisk')]",
+            "lun": 0,
+            "createOption": "FromImage",
```
singhkay: This is missing the image->uri property. I don't think the template works without that.

101-vm-from-user-image-and-datadisk/metadata.json
```diff
@@ -0,0 +1,7 @@
+{
+  "itemDisplayName": "Create a VM from User Image",
+  "description": "This template allows you to create a Virtual Machines from a User image. This template also deploys a Virtual Network, Public IP addresses and a Network Interface.",
+  "summary": "Create a Virtual Machine from a User Image",
+  "githubUsername": "idownard",
+  "dateUpdated": "2015-08-19"
```
singhkay: Date can be updated

singhkay:
For the name, how about 101-vm-user-image-data-disks? It should help "data disks" not get cut off by the github column width and diffrentiate from the other template without data disks

iandow:
Yes, thanks for finding the comments. I removed them, and renamed the project.

singhkay:
@iandow The data disk object seems off to me. Can you deploy and see if there is an error from server side?

iandow:
I have tested the template. It works. It will copy the user-specified os disk image to a new image which will be used for the VM, but the VM will be tied to the data disk that the user specifies.  This seemed the best approach because the time it takes to copy large data disks can cause a timeout in the VM provisioning.

singhkay:
@iandow got it. Can you include this in the readme and any other conditions regarding the location of disks the user needs to take care of before deploying? I always believed if you specified createOption: "FromImage", you needed the image->uri property.

iandow:
@singhkay I've included the image URI property with the data disk, so it's more consistent with the OS disk definition. The README has also been updated to describe prerequisite disk image locations and that the VM will use *copies* of those disk images.  Upon further testing, I don't think copying the data disk will cause the VM provisioning to timeout, afterall.

singhkay:
@iandow Thanks for the update! merged

PR #517, opened 2015-08-25
Move head node out of availability set

sunbinzhu:
Move head node out of availability set so that the size of the head node doesn't need to be A8/A9 when the size of linux compute nodes is A8/A9


azurecla:
Hi __@sunbinzhu__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
    </span>

TTYL, AZPRBOT;

PR #564, opened 2015-09-02
Update diagnostics extension config so it works correctly with portal

sbtron:
Updated the diagnostics extension config to include additional performance counters and metrics that can be displayed in the Azure Portal.
Also switched out the configuration to a variable so it can be read instead of using the based 64 encoded version.


azurecla:
Hi __@sbtron__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
    </span>

TTYL, AZPRBOT;

PR #590, opened 2015-09-09
fixed issue with default parameters

simongdavies:
Default parameter value for SharePointFarm Name caused template to fail
with computer names longer than 15 characters


azurecla:
Hi __@simongdavies__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
    </span>

TTYL, AZPRBOT;

PR #631, opened 2015-09-21
Install and configure mesos-dns on masters and agents

artemharutyunyan:


azurecla:
Hi __@artemharutyunyan__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
    </span>

TTYL, AZPRBOT;

anhowe:
If the master was an agent, the /etc/resolv.conf will be overwritten.  Is that ok?

anhowe:
for the zk config don't you want to use: zk://10.0.0.5:2181,10.0.0.6:2181,10.0.0.7:2181/mesos with all ip addresses.  If so replace that with $zkmesosconfig

artemharutyunyan:
We've established that the patch will not be actually rewriting resolv.conf, but will be amending it, as for the zk configuration, it's safer for the time being to just set it to 127.0.0.1, because if something is wrong with the box and the zk is not functioning on it we'd rather see mesos-dns fail too.

azurecla:
__@artemharutyunyan__, Thanks for signing the contribution license agreement so quickly! Actual humans will now validate the agreement and then evaluate the PR.
<br />Thanks, AZPRBOT;

PR #634, opened 2015-09-22
Add SSH rsa public key parameter for all linux vms, and docker to linux jumpbox

anhowe:
- enable the user to specify an ssh public key through the parameter.  The reason this is done through the script is to provide the user the option, and the normal method highlighted in https://github.com/anhowe/scratch/tree/master/linuxvm does not provide this option.
- add docker to linux jumpbox

@ahmetalpbalkan, @rgardler

### DIFF
--- a/mesos-swarm-marathon/agent-0.json
+++ b/mesos-swarm-marathon/agent-0.json
@@ -115,6 +115,24 @@
       "metadata": {
         "description": "Flag for enabling the Chronos framework."
       }
+    },
+    "sshRSAPublicKey": {
+      "type": "string",
+      "metadata": {
+        "description": "Configure all linux machines with the SSH rsa public key string.  Use 'disabled' to not configure access with SSH rsa public key."
+      }
+    },
+    "omsStorageAccount": {
+      "type": "string",
+      "metadata": {
+        "description": "The storage account for OMS log data."
+      }
+    },
+    "omsStorageAccountKey": {
+      "type": "securestring",
+      "metadata": {
+        "description": "The storage account primary or secondary key for OMS log data."
+      }
     }
   },
   "variables": {},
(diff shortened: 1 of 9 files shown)


azurecla:
Hi __@anhowe__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
    </span>

TTYL, AZPRBOT;

ahmetalpbalkan:
@anhowe I think the way you add the SSH public key to the VM is inconvenient. It will work but it is vague and you have a lot of code around it just to handle/parse/place it. Please check the swarm template about how it is done. Our infrastructure already makes this easy to specify keys in the provisioning configuration.

anhowe:
The swarm template doesn't solve the problem I highlighted in the description to this pull.  How do you solve that problem?

ahmetalpbalkan:
@anhowe I am not sure I understand the problem. Is it because you want to give user an option to choose password or public key?

anhowe:
yes, to provide the user an option to enable ssh access

anhowe:
let's take this discussion offline

rgardler:
Please post a summary of that discussion here. That way this issue becomes a useful piece of documentation for those facing the same issue.

ahmetalpbalkan:
It's not trivial to easily make templates use password for Virtual Machines and at the same time "give an option" to use SSH public keys. Therefore the scripts are there to write the public key contents to `>> $HOME/.ssh/authorized_keys`. The script is trivial, as I also said will probably just work fine. The point I didn't see here initially was not many people just trying to kick the tires will have SSH key pairs, so we still need to have a password option as well.

rgardler:
Yeah, that makes sense. Thanks for following up here.

PR #639, opened 2015-09-22
Update Key Vault template to API version 2015-06-01

seanbamsft:
This change updates the API version to 2015-06-01. I have also changed the template to deploy the vault to the same location as the resource group and allow the user to specify permissions as parameters.

### DIFF
--- a/101-create-key-vault/azuredeploy.json
+++ b/101-create-key-vault/azuredeploy.json
@@ -46,6 +46,7 @@
     },
     "enableVaultForDeployment": {
       "type": "bool",
+      "defaultValue": false,
       "allowedValues": [
         true,
         false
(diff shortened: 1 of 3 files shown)


singhkay:
Looks good overall @seanbamsft. Few additional questions

* Can the key vaults now be deployed to all locations? 
* I see you changed the GitHub username in the metadata file. Will you be owning the template going forward?

seanbamsft:
Yes, Key Vault is now available in all public regions (including Australia, but not China or the US government regions).

I'll be the owner of the Key Vault templates and keeping them up to date as we add features.

singhkay:
@seanbamsft Sounds good!

PR #704, opened 2015-10-12
Refactoring and added support for premium storage

trentmswanson:
* changed region to location
* removed scale grouping configurations and replaced with storage bin packing mapping
* included build templates and scripts for generating the data templates
* included the reference template test mocks
* added initial structure for simplifying the template usage


azurecla:
Hi __@trentmswanson__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!

TTYL, AZPRBOT;

elasticsearch/test/client-nodes-resources.json
```diff
@@ -0,0 +1,101 @@
+{
+  "$schema": "https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#",
```
singhkay: Is test an artifact left over from testing?

trentmswanson: No these are test resources when developing/testing some things in the top level azuredeploy.json template, they are essentially mocks that are generated from the referenced templates and deploy no resources.  The top level azuredeploy.json template can be changed to use the /test base path, and then you run a deployment you can test the parameter manipulation logic.  This makes editing the top level template much easier and faster.

trentmswanson: I guess after thinking about it here, I would have to say yes they are left over from test but intentionally included them.

PR #715, opened 2015-10-15
UI fix

liupeirong:
Use builtin UI controls


azurecla:
Hi __@liupeirong__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
    </span>

TTYL, AZPRBOT;

mysql-replication/azuredeploy.json
```diff
@@ -207,7 +212,8 @@
         "sshIPConfig": "[concat(variables('lbID'),'/frontendIPConfigurations/',variables('sshIPConfigName'))]",
         "nicName": "[concat(parameters('dnsName'), '-nic')]",
         "availabilitySetName": "[concat(parameters('dnsName'), '-set')]",
-        "vhdContainer": "[concat('http://',parameters('storageAccountName'),'.blob.core.windows.net/vhds/')]",
+        "storageAccountName": "[concat(parameters('storageAccountPrefix'), '0')]",
```
singhkay: @liupeirong Looks like this'll likely cause storage name collisions in case a user uses the default value. I suggest appending a random string using the uniqueString function

wordpress-mysql-replication/mainTemplate.json
```diff
@@ -0,0 +1,384 @@
+{
+    "$schema": "http://schema.management.azure.com/schemas/2014-04-01-preview/deploymentTemplate.json#",
```
singhkay: Update schema to 2015-01-01

wordpress-mysql-replication/mainTemplate.json
```diff
@@ -0,0 +1,384 @@
+{
```
singhkay: This file looks like it's the same as azuredeploy.json ?

liupeirong: This will be integrated with marketplace.  If this repo only takes azuredeploy.json, and marketplace only takes mainTemplate.json, what would you recommend other than having 2 duplicate files?

singhkay: Will azuredeploy be going to marketplace as well? I suggest having just azuredeploy while we sort out the direction on integrating marketplace templates here

liupeirong:
Removed mainTemplate for now.

PR #727, opened 2015-10-20
Updated Title and custom deployment link

nilaydshah:
Updated Title and custom deployment link

### DIFF
--- a/101-create-backup-vault-policy/README.md
+++ b/101-create-backup-vault-policy/README.md
@@ -1,6 +1,6 @@
-# Create Backup Vault
+# Create Backup Vault and Policy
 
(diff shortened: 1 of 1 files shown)


azurecla:
Hi __@nilaydshah__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
    </span>

TTYL, AZPRBOT;

singhkay:
@nilaydshah We're testing the automated deployment bot. Does this template require any pre-setup before executing the template? If not can you edit the parameters file to include a valid set of values that can deploy this template successfully? Thanks!

nilaydshah:
@singhkay , i tried deploying this template, and found that - policy is getting created successfully. But the response isn't proper. It returns empty guid. whereas we expect id, name etc mandatory fields returned as part of the response. That's why even we have got the policy created, template deployment says Failed with following message: "The response for resource had empty or invalid content." 
What should we do in such case?

singhkay:
@nilaydshah Is the empty response returned from the backup service?

nilaydshah:
@singhkay yes.

singhkay:
@nilaydshah I think this should be fixed on the service side so that the template responds as succeeded

PR #772, opened 2015-10-29
Changed PowerShell DSC extension versions from 1.0-2.3 to v2.8 

slapointe:
This is to prevent templates deployment failure because these versions have been removed 3 days ago.

http://blogs.msdn.com/b/powershell/archive/2015/10/26/azure-dsc-extension-versions-1-0-2-3-no-longer-available.aspx

This pull request also contain fixes for parameters file that were visibly not validated before.

PR #777, opened 2015-10-30
Bug fixes: Updated DSC typeHandlerVersion to 2.8 and update storageAc…

rchaganti:
- DSC typeHandlerVersion was set to 1.9 which is not available anymore. I updated this to 2.8 which is the latest version.
- storageAccount parameter was getting referenced as newStorageAccountName on line 235. I updated this to storageAccount.

### DIFF
--- a/201-vm-domain-join/azuredeploy.json
+++ b/201-vm-domain-join/azuredeploy.json
@@ -232,7 +232,7 @@
         "diagnosticsProfile": {
           "bootDiagnostics": {
              "enabled": "true",
-             "storageUri": "[concat('http://',parameters('newStorageAccountName'),'.blob.core.windows.net')]"
+             "storageUri": "[concat('http://',parameters('storageAccount'),'.blob.core.windows.net')]"
           }
         }
       },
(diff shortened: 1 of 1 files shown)


azurecla:
Hi __@rchaganti__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
    </span>

TTYL, AZPRBOT;

singhkay:
Hi @rchaganti, appreciate you taking the time to fix this! but there's another PR that includes a fix for this #772. I'm going to close this but please review the other PR and suggest any changes you'd like to see. Thanks!

rchaganti:
Thanks @singhkay! I have been multiple other issues with this template. For example, DC gets created but the second VM fails to join domain and it fails to find the domain specified. My initial investigation showed that the DC VM did not reboot after DSC configuration to create the AD forest. After rebooting DC VM manually, I was able to add the second VM to domain. I am validating the fix where I configure LCM on DC VM to Reboot as needed. I will submit another pull request with all fixes.

PR #784, opened 2015-11-01
Updated vm-domain-join to fix a few things

rchaganti:
Sorry about the churn with multiple pull requests. But, I had to make this template work and that involved multiple runs. 

Here is the final update:
* Added restartdc.ps1 and custom script extension to DC-VM. Without
this, the second VM will never join the domain as the DC wouldn't have
restarted. Ideally, we should use LCM setting RebootNodeIfNeeded for
this. However, for some reason, setting LCM completed but it was never
reflecting in the LCM configuration.
* Updated DSC extension typeHandlerVersion to 2.8.
* Updated readme.md to add visualize button.
* Updated metadata.json.
* Updated azuredeploy.parameter.json to change the default values.


azurecla:
Hi __@rchaganti__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!

TTYL, AZPRBOT;

slapointe:
Hi @rchaganti 

You are right, the problem is because the DC need to reboot. I wouldn't go with the CustomScriptExtension to perform the reboot as this is supported in DSC. A fix is on the way that will make it work with the DSC extension.

PR #796, opened 2015-11-03
Azure Traffic Manager ARM examples

GarethBradshawMSFT:


101-traffic-manager-external-endpoint/azuredeploy.json
```diff
@@ -0,0 +1,58 @@
+{
+    "$schema": "https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json",
+    "contentVersion": "1.0.0.0",
+    "parameters": {
+        "dnsname": {
```
singhkay: better to call this uniqueDnsName so the right expectations are set

GarethBradshawMSFT: That's a fair suggestion, I'm trying to keep the names as similar to the portal as possible, where we call the field DNS Name.

singhkay: I think in Portal a feedback is generated if the name exists already. Sadly in our template deploy UI, that's not available :(

101-traffic-manager-external-endpoint/azuredeploy.parameters.json
```diff
@@ -0,0 +1,5 @@
+﻿{
+    "dnsname": {
+	  "value": "something"
```
singhkay: Please enter the value "#####" so we can test this automatically with the bot. The bot looks for "#####" and replaces with random auto-genned strings

101-traffic-manager-virtual-machines/metadata.json
```diff
@@ -0,0 +1,7 @@
+﻿{
+  "itemDisplayName": "Azure Traffic Manager external endpoint example",
```
singhkay: Looks like copy-pasta from previous template :) Needs to be updated

101-traffic-manager-virtual-machines/azuredeploy.json
```diff
@@ -0,0 +1,271 @@
+{
+    "$schema": "https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json",
```
singhkay: This looks like a 201 template for your service as it is a bit more involved than 101 with vms and all. Also consider renaming to 201-traffic-manager-vm-load-balance

101-traffic-manager-virtual-machines/azuredeploy.json
```diff
@@ -0,0 +1,271 @@
...
+    "contentVersion": "1.0.0.0",
+    "parameters": {
+        "dnsname": {
```
singhkay: uniqueDnsName

101-traffic-manager-virtual-machines/azuredeploy.json
```diff
@@ -0,0 +1,271 @@
...
+                "description": "User name for the Virtual Machines."
+            }
+        },
+        "adminPassword": {
+            "type": "securestring",
+            "metadata": {
+                "description": "Password for the Virtual Machines."
+            }
+        },
+        "dnsNameForPublicIP": {
```
singhkay: maybe unique prefix here as well?

101-traffic-manager-virtual-machines/azuredeploy.parameters.json
```diff
@@ -0,0 +1,17 @@
+﻿{
+    "dnsname": {
+        "value": "somethingunique"
```
singhkay: For these unique values, please replace with "#####" to enable automated testing

101-traffic-manager-webapp/README.md
```diff
@@ -0,0 +1,16 @@
+# Azure Traffic Manager with web apps
```
singhkay: This looks like 201 template as well

101-traffic-manager-webapp/azuredeploy.json
```diff
@@ -0,0 +1,99 @@
+{
+    "$schema": "https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json",
+    "contentVersion": "1.0.0.0",
+    "parameters": {
+        "dnsname": {
```
singhkay: unique

101-traffic-manager-webapp/azuredeploy.json
```diff
@@ -0,0 +1,99 @@
...
+            "metadata": {
+                "description": "Relative DNS name for the WebApps, must be globally unique.  An index will be appended for each Web App."
+            }
+        }
+    },
+
+    "variables": {
+        "tmApiVersion": "2015-04-28-preview",
+        "apiVersion": "2015-08-01",
+        "locations": [ "westus", "eastus" ],
```
singhkay: Suggest replacing with readable names West US and East US. Also are there any other valid locations?

101-traffic-manager-webapp/azuredeploy.parameters.json
```diff
@@ -0,0 +1,8 @@
+﻿{
+    "dnsname": {
+        "value": "somethingunique"
```
singhkay: Replace with "#####"

101-traffic-manager-webapp/azuredeploy.parameters.json
```diff
@@ -0,0 +1,8 @@
...
+    },
+    "dnsNameForWebApp": {
+        "value": "somethingunique"
```
singhkay: Replace with "#####"

101-traffic-manager-webapp/metadata.json
```diff
@@ -0,0 +1,7 @@
+﻿{
+  "itemDisplayName": "Azure Traffic Manager external endpoint example",
```
singhkay: Needs to change

101-traffic-manager-external-endpoint/azuredeploy.json
GarethBradshawMSFT: updated.

101-traffic-manager-external-endpoint/azuredeploy.parameters.json
GarethBradshawMSFT: updated

101-traffic-manager-virtual-machines/metadata.json
GarethBradshawMSFT: updated

101-traffic-manager-virtual-machines/azuredeploy.json
GarethBradshawMSFT: the scenario is a basic one, the template unfortunately looks complicated because we have to create the VMs first.  we looked at using a separate template from the gallery to deploy the VMs but then we would have an ugly dependency on resource names used within that template so we opted to do it all in one place.

101-traffic-manager-virtual-machines/azuredeploy.json
GarethBradshawMSFT: updated.

101-traffic-manager-virtual-machines/azuredeploy.parameters.json
GarethBradshawMSFT: updated.

101-traffic-manager-webapp/README.md
GarethBradshawMSFT: Same case as for the VMs, it contains a few extra bits because we have to create the webapp but it's still a simple scenario for us.

101-traffic-manager-webapp/azuredeploy.json
GarethBradshawMSFT: updated.

101-traffic-manager-webapp/azuredeploy.json
GarethBradshawMSFT: Updated the region names, also renamed the element to enforce that these are the chosen locations and not a restriction.

101-traffic-manager-webapp/azuredeploy.parameters.json
GarethBradshawMSFT: updated.

101-traffic-manager-webapp/azuredeploy.parameters.json
GarethBradshawMSFT: updated.

101-traffic-manager-virtual-machines/azuredeploy.json
singhkay: True but think about it from the perspective of the user who is starting to learn about IaaS. Asking them to learn about complex VM scenarios in 101 seems a bit much :) Which is why we usually reserve more complex VM scenarios for 201

GarethBradshawMSFT: I see.  From a traffic manager perspective, the external endpoints example is a more complex example than either the VM or WebApp one because user needs to specify more cryptic parameters such as the location of the nearest Azure dc.  Maybe they should all be 201s.

singhkay: Could be but would there be a 101 for traffic manager template?

GarethBradshawMSFT: Thinking about it more, I've decided to make the web app and vm templates into 201 series.  While the customer scenario is less complex them template itself is more difficult to understand and grasp.  Keeping the external endpoint as a 101 because there is nothing other than the traffic manager profile resource.

singhkay:
@template-deploy-bot test this please

singhkay:
@GarethBradshawMSFT Templates look good are ready to merge. coupld of quick suggestions before merging

Can you rename this one 201-traffic-manager-virtual-machines to 201-traffic-manager-vm since Github column only allows 27 characters before cutting off the name? 
Also the sample.parameters.json files are not needed.

GarethBradshawMSFT:
@singhkay I've renamed the vm example.  the sample.parameters.json are used for the powershell so are safe to leave there.

singhkay:
@GarethBradshawMSFT I understand but the guidance has been that the azuredeploy.parameters.json does represent the sample parameters file in which the values should be changed before running the template. In the current state the user will need to change the values in the sample parameters file in the same way they have to do with azuredeploy parameters file before being able to run a successful deployment due to the unique nature of storage account and dns names. 

So if the user has to change the values in the parameters file anyway, IMO it's better to keep just one parameters file to avoid confusion and setting a pattern where others are submitting two parameters files as well.

GarethBradshawMSFT:
Agree, but I think the ##### gives the impression that # might be a valid character, which is not the case for dns.

Sent from my Windows Phone
________________________________

singhkay:
@GarethBradshawMSFT I'll be editing the guidance around "#####" and when to use it in the contributing guide but I think in general folks understand to replace values in the parameters file for their scenario. Another idea I can think of to avoid the above confusion is to include a note in the readme file and powershell scripts to replace the values in the parameters file. That should remove ambiguity :)

singhkay:
Thanks for the update @GarethBradshawMSFT! merged

PR #828, opened 2015-11-11
SLURM cluster on SLES 12 HPC SKU

smith1511:
Creates a fully configured SLURM cluster based on the SLES 12 HPC SKU.  This cluster is ready to run Intel MPI workloads when deployed on A8/A9 instances.

### DIFF
--- a/slurm-on-sles12-hpc/azuredeploy.json
+++ b/slurm-on-sles12-hpc/azuredeploy.json
@@ -157,7 +157,7 @@
     },
     "vnetID": "[resourceId('Microsoft.Network/virtualNetworks',variables('networkSettings').virtualNetworkName)]",
     "subnetRef": "[concat(variables('vnetID'),'/subnets/',variables('networkSettings').subnet.dse.name)]",
-    "templateBaseUrl": "https://raw.githubusercontent.com/smith1511/hpc/master/slurm-on-sles12/",
+    "templateBaseUrl": "https://raw.githubusercontent.com/Azure/azure-quickstart-templates/master/slurm-on-sles12-hpc/",
     "installationCLI": "[concat('bash azuredeploy.sh ', variables('masterVMName'), ' ', variables('workerVMName'), ' ', parameters('workerNodeCount'), ' ', parameters('hpcUserName'), ' ', variables('templateBaseUrl'))]",
     "storageAccountType": "Standard_LRS"
   },
--- a/slurm-on-sles12-hpc/azuredeploy.json
+++ b/slurm-on-sles12-hpc/azuredeploy.json
@@ -8,12 +8,6 @@
         "description": "Unique public dns name where the master node will be exposed"
       }
     },
-    "newStorageAccountName": {
-      "type": "string",
-      "metadata": {
-        "description": "Unique DNS Name for the Storage Account where the Virtual Machine's disks will be placed."
-      }
-    },
     "adminUserName": {
       "type": "string",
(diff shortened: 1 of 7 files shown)


azurecla:
Hi __@smith1511__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
    </span>

TTYL, AZPRBOT;

azurecla:
__@smith1511__, Thanks for signing the contribution license agreement so quickly! Actual humans will now validate the agreement and then evaluate the PR.
<br />Thanks, AZPRBOT;

slurm-on-sles12-hpc/azuredeploy.json
```diff
@@ -0,0 +1,559 @@
...
+    "subnetRef": "[concat(variables('vnetID'),'/subnets/',variables('networkSettings').subnet.dse.name)]",
+    "templateBaseUrl": "https://raw.githubusercontent.com/Azure/azure-quickstart-templates/master/slurm-on-sles12-hpc/",
+    "installationCLI": "[concat('bash azuredeploy.sh ', variables('masterVMName'), ' ', variables('workerVMName'), ' ', parameters('workerNodeCount'), ' ', parameters('hpcUserName'), ' ', variables('templateBaseUrl'))]",
+    "storageAccountType": "Standard_LRS"
+  },
+  "resources": [
+    {
+      "type": "Microsoft.Storage/storageAccounts",
+      "name": "[parameters('newStorageAccountName')]",
+      "apiVersion": "2015-05-01-preview",
```
singhkay: I suggest making the api version a variable and using it for all resources here and below. Also use the GA api 2015-06-15

slurm-on-sles12-hpc/azuredeploy.json
```diff
@@ -0,0 +1,559 @@
+{
+  "$schema": "https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#",
+  "contentVersion": "1.0.0.0",
+  "parameters": {
+    "dnsName": {
```
singhkay: dnsname and storageaccounts cause a bunch of template failures due to strict requirements, I suggest creating variables for these and generating these using the uniqueString() function. See usage here https://github.com/Azure/azure-quickstart-templates/blob/bfb0c40965969a4884ccee65ff75c7ee13e3195b/301-2fe-lb80-rdp-1be-nsg-rdp/azuredeploy.json#L32

smith1511:
Comments addressed...

singhkay:
Thanks for the update @smith1511. Can you check the comment on using uniquestring for dns name as well?

smith1511:
Can you point me at another example using uniqueString for DNS name?  

I'm unsure about this change to be honest.  I can understand storage as it's essentially an internal impl detail.  However the DNS name is the 'public face' to the customers service/RG and they may in fact want to choose it.  (Unless the example allows it to be overridden?)

singhkay:
@smith1511 Good point, I don't have an example atm but this is something we're looking at solving since we're seeing so many folks running into failures that look like below where they entered the fqdn instead of a prefix. Maybe renaming the parameter to say dnsPrefix would be a good solution for now?

The domain name label test.somedomain.rd.us is invalid. It must conform to the following regular expression: ^[a-z][a-z0-9-]{1,61}[a-z0-9]$

smith1511:
done

PR #830, opened 2015-11-11
Promote "VNet name" and "storage account name/type" as parameters

sunbinzhu:
Let the user to specify "VNet name" and "storage account name/type" as parameters. Previously they are pre-defined in the template as variables.


singhkay:
@sunbinzhu We're trying to move away from storage account name as an input since it causes the most failures and use the uniqueString function to auto-gen it. Any reason to promote it? Also what about Vnet as a parameter?

sunbinzhu:
@singhkay We are going to publish a solution template on Azure marketplace. In the "Template Solutions in the Marketplace Best Practices" document, It is mandatory to provide VNet name and storage account name as parameters. To keep it consistent, i promote them as parameters here as well. I agree with you that it may cause failures if the user input a storage account name which is already used by others. I will keep them as variables here then.

singhkay:
@sunbinzhu What you could also do is take in a 11 char prefix for storage account and the other 13 chars can come from the uniqueString function. That should ensure uniqueness and also satisfy storage 24 char name requirement. Since this is part of the marketplace document, I'm OK with promoting VNET parameter.

PR #843, opened 2015-11-12
Critical bug fixes and data folders setup for elasticsearch settings on data nodes

kirpasingh:
1. Fixed an issue with alternate translation of discovery hosts that jeopardize master discovery and throws cluster block or master not discovered exceptions.
2. Script now sets the data folder paths else elasticsearch stores data on system drive on data nodes
3. Reduce noise in the output by null redirection in the script


elasticsearch/azuredeploy.json
```diff
@@ -174,7 +174,7 @@
     }
   },
   "variables": {
-    "templateBaseUrl": "https://raw.githubusercontent.com/Azure/azure-quickstart-templates/master/elasticsearch/",
+    "templateBaseUrl": "https://raw.githubusercontent.com/kirpasingh/azure-quickstart-templates/master/elasticsearch/",
```
singhkay: Update back to Azure path?

PR #848, opened 2015-11-13
Add createUIDefinition.json and change naming rules

sunbinzhu:
1. Add createUIDefinition.json and mainTemplate.json
2. Change the naming rules for storage accounts and VMs
3. Change the algorithm of VM disks store location.
4. Change some parameter descriptions.


sunbinzhu:
@singhkay Following you suggestion, i keep vnet and storage account names as variables, and leverage the uniqueString function to generate storage account name prefix.

singhkay:
@sunbinzhu We're still working on consuming mainTemplate.json in the same way as azuredeploy.json files. In the meantime, can you just commit the azuredeploy.json instead of mainTemplate.json?

sunbinzhu:
@singhkay I have commited both azuredeploy.json and mainTemplate.json, I know createUiDifinition.json and mainTemplate.json is not used here currently, i just put them here, won't break anything.

PR #863, opened 2015-11-18
WAD 5 Node

ChackDan:


service-fabric-cluster-5-node-secure-1-nodetype-wad/azuredeploy.json
```diff
@@ -0,0 +1,761 @@
...
+    "clusterLocation": {
+      "type": "string",
+      "metadata": {
+        "description": "Location of the Cluster"
+      }
+    },
+    "clusterName": {
+      "type": "string",
+      "defaultValue": "Cluster",
+      "metadata": {
```
singhkay: You can enforce the length like below

"parameters": {
   "<parameterName>" : {
     "type" : "<type-of-parameter-value>",
     "defaultValue": "<optional-default-value-of-parameter>",
     "allowedValues": [ "<optional-array-of-allowed-values>" ],
     "minValue": <optional-minimum-value-for-int-parameters>,
     "maxValue": <optional-maximum-value-for-int-parameters>,
     "minLength": <optional-minimum-length-for-string-secureString-array-parameters>,
     "maxLength": <optional-maximum-length-for-string-secureString-array-parameters>,
     "metadata": {
         "description": "<optional-description-of-the parameter>" 
     }
   }
}

service-fabric-cluster-5-node-secure-1-nodetype-wad/azuredeploy.json
```diff
@@ -0,0 +1,761 @@
...
+      }
+    },
+    "loadBalancedAppPort2": {
+      "type": "int",
+      "defaultValue": 8081,
+      "metadata": {
+        "description": "Input end point2 for the application to use"
+      }
+    },
+    "certificateStoreValue": {
```
singhkay: This can be a variable since there is only one allowed value?

service-fabric-cluster-5-node-secure-1-nodetype-wad/azuredeploy.json
```diff
@@ -0,0 +1,761 @@
...
+        "description": "The Store name where the cert will be deployed in the VM"
+      }
+    },
+    "certificateThumbprint": {
+      "type": "string",
+            "metadata": {
+        "description": "Certificate Thumbprint"
+      }
+    },
+    "sourceVaultValue": {
```
singhkay: should be called key vault resource id

service-fabric-cluster-5-node-secure-1-nodetype-wad/azuredeploy.json
```diff
@@ -0,0 +1,761 @@
...
+    "lbIPConfig0": "[concat(variables('lbID0'),'/frontendIPConfigurations/LoadBalancerIPConfig')]",
+    "lbPoolID0": "[concat(variables('lbID0'),'/backendAddressPools/LoadBalancerBEAddressPool')]",
+    "lbProbeID0": "[concat(variables('lbID0'),'/probes/FabricGatewayProbe')]",
+    "lbHttpProbeID0": "[concat(variables('lbID0'),'/probes/FabricHttpGatewayProbe')]",
+    "vmNodeType0Name": "[concat('NT1', variables('vmName'))]",
+    "vmNodeType0Size": "Standard_A2"
+  },
+  "resources": [
+    {
+      "apiVersion": "2015-05-01-preview",
```
singhkay: Please make the api version a variable and also use the GA versions 2015-06-15

singhkay:
@ChackDan  It's a good idea to name this template different than the other one. Right now it isn't clear which template is which from the repo view

image 054

ChackDan:
@singhkay incorporated your comments.

singhkay:
@ChackDan Looks like there are merge conflicts on this PR. Can you resolve these?

service-fabric-secure-cluster-5-node-1-nodetype-wad/azuredeploy.json
```diff
@@ -0,0 +1,761 @@
+{
+  "$schema": "http://schema.management.azure.com/schemas/2014-04-01-preview/deploymentTemplate.json",
```
singhkay: This should be updated to 2015-01-01 schema

service-fabric-secure-cluster-5-node-1-nodetype-wad/azuredeploy.json
```diff
@@ -0,0 +1,761 @@
...
+  "contentVersion": "1.0.0.0",
+  "parameters": {
+    "clusterLocation": {
+      "type": "string",
+      "metadata": {
+        "description": "Location of the Cluster"
+      }
+    },
+    "clusterName": {
```
singhkay: If this needs to be unique, then it'd be good to prefix this parameter with a "unique" prefix to set the right expectations

service-fabric-secure-cluster-5-node-1-nodetype-wad/azuredeploy.json
```diff
@@ -0,0 +1,761 @@
...
+    "applicationEndPort": "30000",
+    "ephemeralStartPort": "49152",
+    "ephemeralEndPort": "65534",
+    "publicIPAddressName": "PublicIP-VM",
+    "publicIPAddressType": "Dynamic",
+    "vmStorageAccountContainerName": "vhds",
+    "virtualNetworkName": "VNet",
+    "addressPrefix": "10.0.0.0/16",
+    "subnet1Name": "Subnet-1",
+    "subnet2Name": "Subnet-2",
```
singhkay: I don't see this subnet being used anywhere in the template

service-fabric-secure-cluster-5-node-1-nodetype-wad/azuredeploy.json
```diff
@@ -0,0 +1,761 @@
...
+    "lbIPConfig0": "[concat(variables('lbID0'),'/frontendIPConfigurations/LoadBalancerIPConfig')]",
+    "lbPoolID0": "[concat(variables('lbID0'),'/backendAddressPools/LoadBalancerBEAddressPool')]",
+    "lbProbeID0": "[concat(variables('lbID0'),'/probes/FabricGatewayProbe')]",
+    "lbHttpProbeID0": "[concat(variables('lbID0'),'/probes/FabricHttpGatewayProbe')]",
+    "vmNodeType0Name": "[concat('NT1', variables('vmName'))]",
+    "vmNodeType0Size": "Standard_A2"
+  },
+  "resources": [
+    {
+      "apiVersion": "2015-05-01-preview",
```
singhkay: This should become a variable and also move to the GA API version - 2015-06-15

service-fabric-secure-cluster-5-node-1-nodetype-wad/azuredeploy.json
```diff
@@ -0,0 +1,761 @@
...
+        },
+        "subnets": [
+          {
+            "name": "[variables('subnet1Name')]",
+            "properties": {
+              "addressPrefix": "[variables('subnet1Prefix')]"
+            }
+          },
+          {
+            "name": "[variables('subnet2Name')]",
```
singhkay: This subnet isn't being used anywhere, should be removed

ChackDan: Wont Fix - This is so that customers can add any other resources they want to that subnet.. We want to avoid folks adding other azure resources to the subnet1 that we use for the nodeType1.

service-fabric-secure-cluster-5-node-1-nodetype-wad/azuredeploy.json
ChackDan: Wont fix – Any changes to the schema will need to be validated by the product team, so if we do end up moving to a GA version. I will submit a change then. I do not want to make it a variable for now.

service-fabric-secure-cluster-5-node-1-nodetype-wad/azuredeploy.json
ChackDan: Since this is a DNS name, the only way to truly figure out if the selected cluster name will go through is by deploying it, so adding unique does not really reduce the chances that it the deployment will go through or not. So wont fix this. This parameter name is carried though out the product, so making a cosmetic change here may not be the right thing to do anyway.

singhkay: @ChackDan But it could be the difference between the user entering the value "testcluster" or "uniqueString>ClusterName". The first will lead to deployment failures and user will need to start again.

service-fabric-secure-cluster-5-node-1-nodetype-wad/azuredeploy.json
singhkay: Not sure I understand this. IaaS resources can use the 2015-06-15 API version. Is there validation required on service fabric end for this API version?

service-fabric-secure-cluster-5-node-1-nodetype-wad/azuredeploy.json
ChackDan: I think the probabilities of failing are the same… any how,, can you send me the exact Json (snippet) on what you are talking about and how we can validate in JSON that a certain VIP will be unique.

Also – while we try and resolve these differences, I would recommend that you merge what we have got now to main… Folks have been asking about them And I have now started to point them to my on Git repo.. and are using them without issues... I do not want to make that the main place folks go to for SF templates..

singhkay: Sure we can merge today but I think it's important to take the time to recognize the potential point of failures and address them since the friction and cost of failing and having to enter parameters and deploy again is pretty high. I see lines 134, 135 as two potential points of failure since this parameter is being used as is for things that need to be globally unique. Also storage account has another limitation that it needs to have all lower case. So any upper case letters in cluster name will cause the deployment to fail.

    "dnsName": "[parameters('clusterName')]", 
    "vmStorageAccountName": "[concat(parameters('clusterName'),'1')]", 
 
The simplest solution is to give the user an indication to enter a unique value for this parameter. Doing that by renaming this parameter to "uniqueClusterName" should be good. Then you could do

    "storageAccountName" = "[toLower(parameters('uniqueClusterName'))]"

in variables section to prevent failures due to upper case.

But the truly ideal solution which requires more work would be to concat this parameter value with the 13 character output of the uniqueString() function but then you need to limit the input for clusterName to 11 characters since storage accounts can only be a max of 24 characters. This template is a good example of how to that https://github.com/Azure/azure-quickstart-templates/blob/bfb0c40965969a4884ccee65ff75c7ee13e3195b/301-2fe-linux-lb80-ssh-1be-win-nsg-rdp-datadisk-ssd/azuredeploy.json#L90

    "newStorageAccountName": "[concat('pst', uniqueString(resourceGroup().id))]",

ChackDan:
#sign-off

PR #884, opened 2015-11-23
docker-swarm: Hard-code admin username

ahmetalpbalkan:
One more step towards minimalism in the Docker Swarm template.
Having a default username does not impose any significant security
risks yet simplifies the process even better.

Usernames such as "admin" are not allowed in Azure, however portal
does not do any validations about that. So that still leaves a room
for user mistake which actually is our mistake.

The actual value is still in one place in case someone wants to replace
and roll their own username.

cc: @anhowe


singhkay:
@ahmetalpbalkan Looks like there is a conflict. Can you take a look?

ahmetalpbalkan:
@singhkay rebased. PTAL

singhkay:
merged

PR #901, opened 2015-11-30
Add a template to deploy Concourse CI on Azure

bingosummer:
This template prepares Azure resources to deploy Concourse CI on Azure, including 1 storage account, 1 virtual network with 2 subnets, 2 public IPs and 1 VM.

### DIFF
--- a/concourse-ci/azuredeploy.json
+++ b/concourse-ci/azuredeploy.json
@@ -1,0 +1,284 @@
+{
+  "$schema": "https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#",
+  "contentVersion": "1.0.0.0",
+  "parameters": {
+    "virtualNetworkName": {
+      "type": "string",
+      "defaultValue": "vnet",
+      "metadata": {
+        "description": "name of the virtual network"
+      }
+    },
+    "virtualNetworkAddressPrefix": {
+      "type": "string",
+      "defaultValue": "10.0.0.0/16",
+      "metadata": {
+        "description": "address prefix of the virtual network"
+      }
+    },
+    "subnetNameForBosh": {
+      "type": "string",
+      "defaultValue": "Bosh",
+      "metadata": {
+        "description": "name of the subnet for Bosh"
+      }
+    },
+    "subnetAddressPrefixForBosh": {
+      "type": "string",
+      "defaultValue": "10.0.0.0/24",
+      "metadata": {
+        "description": "address prefix of the subnet for Bosh"
+      }
+    },
+    "subnetNameForConcourse": {
+      "type": "string",
+      "defaultValue": "Concourse",
+      "metadata": {
+        "description": "name of the subnet for Concourse"
+      }
+    },
+    "subnetAddressPrefixForConcourse": {
+      "type": "string",
+      "defaultValue": "10.0.16.0/24",
+      "metadata": {
+        "description": "address prefix of the subnet for Concourse"
+      }
+    },
+    "vmName": {
+      "type": "string",
+      "metadata": {
+        "description": "name of the vm, will be used as DNS Name for the Public IP used to access the Virtual Machine"
+      }
+    },
+    "vmSize": {
+      "type": "string",
+      "defaultValue": "Standard_D1",
+      "allowedValues": [
+        "Standard_A1",
+        "Standard_A2",
+        "Standard_A3",
+        "Standard_A4",
+        "Standard_D1",
+        "Standard_D2",
+        "Standard_D3",
+        "Standard_D4"
+      ],
+      "metadata": {
+        "description": "Size of vm"
+      }
+    },
+    "adminUsername": {
+      "type": "string",
+      "metadata": {
+        "description": "User name for the Virtual Machine."
+      }
+    },
+    "adminPassword": {
+      "type": "securestring",
+      "metadata": {
+        "description": "Password for the Virtual Machine."
+      }
+    },
+    "tenantID": {
+      "type": "string",
+      "defaultValue": "TENANT-ID",
+      "metadata": {
+        "description": "ID of the tenant."
+      }
+    },
+    "clientID": {
+      "type": "string",
+      "defaultValue": "CLIENT-ID",
+      "metadata": {
+        "description": "ID of the client"
+      }
+    },
+    "clientSecret": {
+      "type": "securestring",
+      "defaultValue": "CLIENT-SECRET",
+      "metadata": {
+        "description": "secret of the client"
+      }
+    }
+  },
+  "variables": {
+    "apiVersion": "2015-06-15",
+    "location": "[resourceGroup().location]",
+    "newStorageAccountName": "[concat('concourse', uniqueString(resourceGroup().id))]",
+    "storageAccountType": "Standard_RAGRS",
+    "vmStorageAccountContainerName": "vhds",
+    "storageid": "[resourceId('Microsoft.Storage/storageAccounts', variables('newStorageAccountName'))]",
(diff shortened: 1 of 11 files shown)


concourse-ci/azuredeploy.json
```diff
@@ -0,0 +1,284 @@
...
+      "metadata": {
+        "description": "name of the subnet for Concourse"
+      }
+    },
+    "subnetAddressPrefixForConcourse": {
+      "type": "string",
+      "defaultValue": "10.0.16.0/24",
+      "metadata": {
+        "description": "address prefix of the subnet for Concourse"
+      }
```
singhkay: Can the above vnet & subnet parameters become variables instead? to reduce the number of parameters

concourse-ci/azuredeploy.json
```diff
@@ -0,0 +1,284 @@
...
+    },
+    "adminPassword": {
+      "type": "securestring",
+      "metadata": {
+        "description": "Password for the Virtual Machine."
+      }
+    },
+    "tenantID": {
+      "type": "string",
+      "defaultValue": "TENANT-ID",
```
singhkay: Please remove the default values. This is going to cause  PS to fail using the Uri deployment method

concourse-ci/azuredeploy.json
```diff
@@ -0,0 +1,284 @@
...
+          "SUBNET-NAME-FOR-CONCOURSE": "[parameters('subnetNameForConcourse')]",
+          "SUBSCRIPTION-ID": "[subscription().subscriptionId]",
+          "STORAGE-ACCOUNT-NAME": "[variables('newStorageAccountName')]",
+          "STORAGE-ACCESS-KEY": "[listKeys(variables('storageid'),variables('apiVersion')).key1]",
+          "RESOURCE-GROUP-NAME": "[resourceGroup().name]",
+          "TENANT-ID": "[parameters('tenantID')]",
+          "CLIENT-ID": "[parameters('clientID')]",
+          "CLIENT-SECRET": "[parameters('clientSecret')]",
+          "CONCOURSE-IP": "[reference('concourse').ipAddress]",
+          "username": "[parameters('adminUserName')]"
```
singhkay: Didn't know custom script had so many other setting properties. How do these work? Can they be accessed inside the script?

concourse-ci/azuredeploy.json
bingosummer: Thanks @singhkay for your comments.
Yes. We can move them into variables.
However, I have a question about parameters and variables. How to determine which is a parameter and which is a variable?
If users want to change the subnet address prefix for some reason, they have to modify the variables in `azuredeploy.json`. Right?

concourse-ci/azuredeploy.json
bingosummer: @singhkay Could you explain more about why the default values cause PS failure deployment? It sounds like a bug.

singhkay: If you try to deploy a template like below without specifying a parameters file, PS doesn't prompt you to enter the value and uses the default value specified in the template 

New-AzureResourceGroupDeployment -Name <deployment-name> -ResourceGroupName <resource-group-name> -TemplateUri https://raw.githubusercontent.com/azure/azure-quickstart-templates/master/checkpoint-multi-nic/azuredeploy.json

concourse-ci/azuredeploy.json
bingosummer: It is a tricky way to leverage CustomScript to pass parameters.
Actually, CustomScript don't understand so many other settings. But these settings will be put into a configuration file. And `setup_devbox.py` in this template will parse these settings from the configuration file.

concourse-ci/azuredeploy.json
singhkay: Usually things that don't need to be globablly unique should become variables. Users can change these values in "Deploy to Azure" experience in the Portal before deploying.

concourse-ci/azuredeploy.json
singhkay: Do these get put into a config file on the VM?

concourse-ci/azuredeploy.json
bingosummer: If PS failure with the default values is by design, not a bug. I prefer to remove tenantId, clientId and clientSecret from parameters instead of removing the default values.

singhkay: If these properties are important to the scenario then they should remain parameters instead of variables so that the user knows what absolutely needs to be entered for a successful deployment

concourse-ci/azuredeploy.json
bingosummer: Got it. Will move vnet & subnet parameters into variables.
But I'm not sure whether users know the difference between parameters and variables. They may think only the parameters are configurable.

singhkay: Very true. We're trying to balance first time experience for new users who might be overwhelmed by the number of parameters versus advanced users who know variables can be changed as well :) 

Think of parameter vs variables as something that is needed for a successful deployment and needs to change vs something that is OK not changing between different deployments.

concourse-ci/azuredeploy.json
bingosummer: > If you try to deploy a template like below without specifying a parameters file, PS doesn't prompt you to enter the value and uses the default value specified in the template

Actually, the above behavior is expected. It won't break the scenario of this template. So I think, we can reserve these three parameters.

concourse-ci/azuredeploy.json
bingosummer: Yes. In a config file like `/var/lib/waagent/Microsoft....CustomScript/config/0.config`

concourse-ci/azuredeploy.json
singhkay: What happens in the case where values passed are TENANT-ID, CLIENT-ID, CLIENT-SECRET? What capability is the user missing out on?

concourse-ci/azuredeploy.json
singhkay: Ah interesting!

concourse-ci/azuredeploy.json
bingosummer: These three values should be configured in `bosh.yml`.
If users leave the default values, we just pass TENANT-ID, CLIENT-ID, CLIENT-SECRET into `bosh.yml`. And users need to update them manually after they login the dev box.

Q: Why I need to set the default values?
A: I want to make these three values optional. AFAIK, I can't set these parameters are required and those are optional except providing the default values. If the default values are provided, these parameters are optional. Right?

concourse-ci/azuredeploy.json
bingosummer: Make sense. :)

sedouard:
@bingosummer this probably should eventually use [template linking](https://azure.microsoft.com/en-us/documentation/articles/resource-group-linked-templates/) as the `azuredeploy.json` is super similar to the `azuredeploy.json` found in the `bosh-setup` template. This way as updates are made to the bosh setup template, you get them here.

bingosummer:
@sedouard There are still some changes from the `azuredeploy.json` in bosh-setup. For now, we do not have the plan to link to the `azuredeploy.json` in bosh-setup.

sedouard:
I just did a [quick diff](http://www.diff-online.com/view/5666f1c34d761) and there aren't any significant changes that would necessitate a different template. The changes are really just around variable naming. You can parametrize the startup script and avoid having all that redundant template code.

Its obviously not necessary now but I would urge that you do that since concourse is meant to be deployed against a BOSH director on azure which is exactly what the `bosh-init` template does. It doesn't make sense to repeat that portion. Instead, reusing the same consistent BOSH deployment template would make future changes apply to this template, too.

PR #920, opened 2015-12-04
Update to include Brazil South Region

pateixei:
Update including Brazil South region, which already supports ARM & IaaS v2 deployments; also added clarification regarding the type of gateway to be used in the deployment


azurecla:
Hi __@pateixei__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
    </span>

TTYL, AZPRBOT;

PR #949, opened 2015-12-11
changed dependsOn to storageLoop instead of individual sa's

gatneil:


singhkay:
@gatneil Formatting looks off. Can you run this through a JSON formatter?

PR #968, opened 2015-12-15
moved commandToExecute to protectedSettings to protect passwords and keys

gatneil:
moved commandToExecute to protectedSettings to protect passwords and keys
renamed improperly-named files
fixed formatting in some templates
converted parameter files to use Travis CI


singhkay:
New one opened here https://github.com/Azure/azure-quickstart-templates/pull/977

PR #969, opened 2015-12-15
Availability set strategy and some minor updates

sunbinzhu:
1. Availability set strategy change to meet the best practice: (I) Always create compute nodes in availability sets. (2) In principal, head node will not be put in an availability set, except that when its VMSize and the compute node VM size are both A8/A9, because only VMs in one availability set can communicate with Infiniband nic.
3. Change maximum VMs in one storage account from 10 to 20 to save storage accounts
4. record resource group name in head node for future use.
5. Modify some strings.


sunbinzhu:
The commit removes linuxnode-rdma-disabled.json and linuxnode-rdma-enabled.json and add one file linuxnode.json, because previously the only difference between the former two files is put in an availability set or not. Now the compute nodes will always be put in availability set, so i combined these two files into one.
The removal of availabilityset-disabled.json and availabilityset-enabled.json and the adding of availabilitySets.json is because we will always create availability sets.
Btw, we also removed some deprecated codes from a PowerShell script and update the DSC extension version to 2.10

sunbinzhu:
@singhkay Is the check failure due to missing linked resource template? In the azuredeploy.json, i have hardcode the baseUrl of the linked resource templates as "https://raw.githubusercontent.com/Azure/azure-quickstart-templates/master/create-hpc-cluster/". The deployment will fail in this case before the commit is merged to Azure/master branch because the linked resource templates were changed. 
When i did test, I changed "Azure" to "sunbinzhu" to make it run.

singhkay:
I believe the CI is taking into account hard coded URLs. Adding @sedouard to take a look

sunbinzhu:
@sedouard could you help to check?

sedouard:
Hey @sunbinzhu, the CI system automatically takes into account for any links in your `azuredeploy.json` that points to `raw.githubusercontent.com/Azure/azure-quickstart-templates/master`. Even though the script may not exist in this repository, we are smart enough to recalculate the link to your fork, where we expect it to exist. *So keep your template links pointing here and not your fork*

Your last run was running some code we accidentally merged, and also had a bug which made it run lots of other unrelated templates. If you resolve your merge conflict and update your branch, you should get a new run with just your template.

sedouard:
See https://github.com/Azure/azure-quickstart-templates/blob/master/CONTRIBUTING.md#rawgithubusercontentcom-links

sunbinzhu:
@singhkay the CI check passed now, please help to merge

singhkay:
lgtm, merged

PR #975, opened 2015-12-15
Dynamic Web Tests

sedouard:


sedouard:
@acomghbot lies. the template is good

singhkay:
@sedouard Ya, looks like it's only a validation issue due to name but at deployment time a different name is generated. Not sure if @rjmax has any guidance but FYI anyways. Merging in the meantime.

PR #992, opened 2015-12-17
rename linux-puppet-agent to puppet-agent-linux

sebastus:


singhkay:
merging since we already checked this last

Issue #966, opened 2015-12-14
[HELP WANTED] DNS names best guidance


singhkays:
Currently for DNS names the user must provide a prefix rather than a FQDN. The provided prefix is concatenated with <location>.cloudapp.azure.com by the platform to provide the FQDN for the Public IP. We see users trying to enter a FQDN (abc.com, def.com etc.) which causes failures. The DNS prefix must match the following regex or it will cause an error 

```
`^[a-z][a-z0-9-]{1,61}[a-z0-9]$`
```

We'd like to change current templates to provide this guidance as part of the parameter name and description. Here's how to do this
1. Make sure any template parameter for a DNS label has a `Prefix` at the end. This will give the user a hint of the value that's needed for this parameter. 
2. Add to the `metadata.description` information about the regex the DNS labels are expected to meet.

Here's how a good dns label parameter will look

```
"dnsLabelPrefix": {
   "type": "string",
   "metadata": {
      "description": "DNS Label for the Public IP. Must be lowercase. It should match with the following regular expression: ^[a-z][a-z0-9-]{1,61}[a-z0-9]$ or it will raise an error."
   }
},
```

---

UPDATE 12/17/2015
@humbertcostas is working on this now

humbertcostas:
Hi @singhkay , could you assign me this issue?

singhkay:
@humbertcostas For some reason, Github doesn't let me assign it to you :|
![image](https://cloud.githubusercontent.com/assets/6164178/11877242/dc6fb0ca-a4a2-11e5-9a20-ba47e0a93bf2.png)

sedouard:
@singhkay you can't assign to non-collaborators. You'll just have to note in the comment's that @humbertcostas is working on it. :-)

singhkay:
@sedouard Ah ok! 

@humbertcostas is working on this now

humbertcostas:
Hi @sedouard , I signed the CLA days ago...
https://github.com/Azure/azure-quickstart-templates/pull/942#issuecomment-164702940

Issue #944, opened 2015-12-10
[HELP WANTED] Sanitize storage account name input in existing templates


singhkays:
Currently the storage account names have the most strict requirements for the input. These are needed to be globally unique, all lower case and 24 characters or less. Most templates do not currently sanitize the user input which leads to a bunch of failures. We can use your help to improve the existing templates by making sure the storage account inputs follow the rules. Here's how to do this
### New Proposal (based on @bingosummer's suggestion below)
1. Remove any storage account name parameters. Create a variable instead. Only get a storage account name input from user if **absolutely** required.
2. Create a storage account variable and assign it's value like below

```
"storageAccountName": "[concat(uniquestring(resourceGroup().id), '<scenario name or a common suffix>')]",
```

An example template is found here https://github.com/Azure/azure-quickstart-templates/blob/master/101-storage-account-create/azuredeploy.json#L20

bingosummer:
@singhkay 
Does it make sense that I concat a fixed prefix `cf` and `uniqueString(resourceGroup().id)` as the storage name? Then I can remove the storage account parameter.

singhkay:
@bingosummer That makes sense. Actually, after discussions with the team that might be the direction we move in and eliminate the storage account parameter to avoid any chance of failure (unless it really is needed to be entered from the user). Also having a different prefix for different account names ensures they land on different racks which could help with HA scenarios.

Here's the proposed solution

PROPOSED:
``` concat(uniqueString() + <scenario name>)```

**Comment by boutquin** [comment]
@singhkay 
I think I followed the above recipe in pull request #1087 and the template fails validation, seemingly because the variable value defined in
```
"storageAccountNamePrefix": {
   "value": "GEN_UNIQUE_11"
 },
```
conflicts with the string length specified in
```
"storageAccountNamePrefix": {
  "type": "string",
  "maxLength": 11
  "metadata": {
    "description": "Name prefix of the Storage Account"
  }
},
```
Any advice, short of adding .ci_skip?

boutquin:
@singhkay 

Never mind. Found the issue: using"GEN-UNIQUE-11" instead of "GEN_UNIQUE_11" helps. ;)

ryanreich:
This fails for me: `The resource name 'dwjbicumxp2fy' is not allowed.`.  It seems that `uniqueString` produces alpha *numeric* strings, while storage account names must be alphabetic only...

PR #995, opened 2015-12-17
rename folder

sebastus:
rename windows-puppet-agent to puppet-agent-windows and fixed up the
readme.  left windows-puppet-agent in place for a bit with a notice of
the change in the readme.


singhkay:
@sedouard Let's make checking the README.md filename case insensitive

singhkay:
@sedouard Do you know what this error means? 

```Server Error:undefined```

**Comment by sedouard** [comment]
Generally that means the server was unreachable. Strange we don't get an error object in that scenario. I restarted the run. The server logs show no problems...

**Comment by sedouard** [comment]
@singhkay readme files should always be README.md by convention, like every other git repository. Any reason we would like to allow names like `ReAdMe.mD`?

**Comment by singhkay** [comment]
@sedouard Hmm, didn't know there was a convention in place for the case. Let's keep it upper case then. Btw, I was working with @sebastus offline on this template and looks like the extension fails but is not represented in the error here which I assume is CLI not bubbling it into the console.

File: windows-puppet-agent/azuredeploy.parameters.json [hunk 1]
```diff
@@ -12,25 +12,16 @@
...
     "vmDnsName": {
-      "value": "unique dns name"
-    },
-    "imagePublisher": {
-      "value": "image Publisher"
-    },
-    "imageOffer": {
-      "value": "image Offer"
-    },
-    "imageSKU": {
```
sedouard: I'm just going to guess that `imageSKU` should not be a randomly generated value?

windows-puppet-agent/azuredeploy.parameters.json
```diff
@@ -12,25 +12,16 @@
       "value": "admin password"
```
sedouard: this probably has to be `GEN-PASSWORD`

puppet-agent-windows/azuredeploy.json
```diff
@@ -0,0 +1,220 @@
...
+        "[concat('Microsoft.Compute/virtualMachines/', variables('vmName'))]"
+      ],
+      "properties": {
+        "publisher": "PuppetLabs",
+        "type": "PuppetEnterpriseAgent",
+        "typeHandlerVersion": "3.2",
+        "settings": {
+          "PUPPET_MASTER_SERVER": "[parameters('puppet_master_server')]"
+        },
+        "protectedSettings": {
```
singhkay: This looks pretty weird. What if you just remove protectedSettings?

windows-puppet-agent/azuredeploy.parameters.json
singhkay: He's essentially doing a rename of the folder windows-puppet-agent to puppet-windows-agent. But we need to keep the windows-puppet-agent folder around with a readme pointing to newer folder for a week or so to let people updated their URLs. So failing windows-puppet-agent is fine. 

@sebastus I think you can go ahead and just remove all template files from windows-puppet-agent and just leave a readme

puppet-agent-windows/azuredeploy.json
```diff
@@ -0,0 +1,220 @@
+{
+  "$schema": "https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#",
+  "contentVersion": "1.0.0.0",
+  "parameters": {
+    "newStorageAccountName": {
```
singhkay: Should be storageAccountNamePrefix see #944

puppet-agent-windows/azuredeploy.json
```diff
@@ -0,0 +1,220 @@
...
+      "type": "string",
+      "maxLength": 11,
+      "metadata": {
+        "description": "Name prefix of the storage account to hold your VM disk."
+      }
+    },
+    "vmDnsName": {
```
singhkay: Should be called dnsLabelPrefix see #966

puppet-agent-windows/azuredeploy.json
```diff
@@ -0,0 +1,220 @@
...
+    },
+    "adminPassword": {
+      "type": "securestring",
+      "metadata": {
+        "description": "Admin password name for the Virtual Machines"
+      }
+    },
+    "location": {
+      "type": "string",
+      "allowedValues": [
```
singhkay: It might be better to remove the allowed values since this list is out of date with the number of locations now

puppet-agent-windows/azuredeploy.json
```diff
@@ -0,0 +1,220 @@
...
+      }
+    },
+    "vmSize": {
+      "type": "string",
+      "defaultValue": "Standard_D2",
+      "metadata": {
+        "description": "VM Size for creating the Virtual Machine"
+      }
+    },
+    "puppet_master_server": {
```
singhkay: probably suffix a 'url' at the end to make it clear a url is needed here

puppet-agent-windows/azuredeploy.json
```diff
@@ -0,0 +1,220 @@
...
+    "imagePublisher": "MicrosoftWindowsServer",
+    "imageOffer": "WindowsServer",
+    "imageSKU": "2012-R2-Datacenter",
+    "storageAccountName": "[replace(replace(tolower(concat(parameters('newStorageAccountName'), uniquestring(resourceGroup().id))), '-',''),'.','')]"
+  },
+  "resources": [
+    {
+      "type": "Microsoft.Storage/storageAccounts",
+      "name": "[variables('storageAccountName')]",
+      "apiVersion": "2015-06-15",
```
singhkay: It'd be good to make apiVersion a variable to move to future versions quickly

puppet-agent-windows/azuredeploy.json
```diff
@@ -0,0 +1,220 @@
...
+    "storageAccountType": "Standard_LRS",
+    "publicIPAddressName": "myPublicIP",
+    "publicIPAddressType": "Dynamic",
+    "vmStorageAccountContainerName": "vhds",
+    "virtualNetworkName": "MyVNET",
+    "addressPrefix": "10.0.0.0/16",
+    "subnet1Name": "subnet-1",
+    "subnet2Name": "subnet-2",
+    "subnet1Prefix": "10.0.0.0/24",
+    "subnet2Prefix": "10.0.1.0/24",
```
singhkay: subnet 2 isn't used anywhere, can be removed

puppet-agent-windows/metadata.json
```diff
@@ -0,0 +1,7 @@
+{
+  "itemDisplayName": "Puppet agent on Windows VM",
+  "description": "Deploy a windows VM with Puppet Agent",
+  "summary": "Deploy an Windows VM With Puppet Agent",
+  "githubUsername": "kundanap",
+  "dateUpdated": "2015-04-23"
```
singhkay: Don't forget to update this date on the final commit

puppet-agent-windows/azuredeploy.json
sebastus: The deployment fails without the protectedSettings.  This is the subject of that big email thread with Kundana.

puppet-agent-windows/metadata.json
sebastus: @singhkay 

Should I also change the githubUsername?

sebastus:
@singhkay   I tested this version, it works in local deployment with actual parameters.

sebastus:
@singhkay 

@sedouard  pointed out that we could add a .ci_skip file to the olde directory so travis would ignore it.

sedouard:
@sebastus +1000 points for squashing :-).

sebastus:
@sedouard 

The notes on squashing never say "save the text file after editing picks to fixups".  And it calls it a screen instead of an open notepad.  Squashing is scary enough without the docs being a bit off center.  ;)

singhkay:
@sebastus Sounds good, thanks for bearing with us :)

PR #1000, opened 2015-12-17
Typo tempalte->template

gbowerman:


singhkay:
I thought I was the only one who made that mistake!

PR #1006, opened 2015-12-18
Master

amitsriva:
Template for configuring Azure Point-to-Site connectivity.


101-point-to-site/azuredeploy.json
```diff
@@ -0,0 +1,171 @@
+{  
+   "$schema":"http://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json",
+   "contentVersion":"1.0.0.0",
+   "parameters":{  
+      "location":{
```
singhkay: If this isn't different from the resource group location, then the new recommendation is to make a variable called location and it's value is resourceGroup().location. Otherwise, the users have to pick a location twice in the "Deploy to Azure" xp

101-point-to-site/azuredeploy.json
```diff
@@ -0,0 +1,171 @@
...
+            "description":"Location where VNet, Gateway, PublicIP will be deployed."
+         }
+      },
+      "virtualNetworkName":{  
+         "type":"string",
+         "metadata":{  
+            "description":"Name of the VNet."
+         }
+      },
+      "addressPrefix":{
```
singhkay: vnetAddressPrefix would be a better name

101-point-to-site/azuredeploy.json
```diff
@@ -0,0 +1,171 @@
...
+      },
+      "revokedCertThumbprint":{  
+         "type":"string",
+         "metadata":{  
+            "description":"Thumbprint of the revoked certificate. This would revoke VPN client certificates matching this thumbprint from connecting to the VNet."
+         }
+      }
+   },
+   "variables":{  
+      "apiVersion":"2015-05-01-preview",
```
singhkay: Can this be 2015-06-15?

singhkay:
looks good mostly, only few minor changes needed

singhkay:
CI passed, merging

PR #1015, opened 2015-12-20
Added CI to few more DSC using templates

singhkay:


singhkay:
@sedouard Might need to go over 120 minutes. 

Tested the third template works locally. merging

sedouard:
Not really possible to extended besides [begging the folks at Travis](https://github.com/travis-ci/travis-ci/issues/5346)

sedouard:
In the meantime it may make sense to just put a `.ci_skip` file in that Sharepoint template as the other two templates took about 25 minutes to deploy compared to **over 2 hours** for sharepoint (geez that is, so, so long)

singhkay:
@sedouard Ya that one is reaaaallly long! Let's see if we continue to see more templates like this

PR #1020, opened 2015-12-21
More CId templates

singhkay:


singhkay:
looks like the sendgrid account needs additional work on CI. Other templates pass, merging.

PR #1021, opened 2015-12-21
Merged and removed duplicate/non-working templates, CId others

singhkay:
* Merged duplicate data disk templates into simple linux and windows template
* Removed backup-vault policy template due to not being designed for guidelines
* Merged VNET template with 2 subnets template


singhkay:
ci passed, merging

PR #1023, opened 2015-12-22
re-fact the HPC Pack cluster templates

sunbinzhu:
re-fact the HPC Pack cluster templates to keep consistent with that in Marketplace.
The azuredeploy.json links to mainTemplate.json


sunbinzhu:
Hello, @sedouard @singhkay , the CI check fail again while i can succeed to deploy manually. Can you help to check, and how to self-analyze the error log of CI? I found the error message is "AssertionError: Template Validiation Failed. .... Server Error:undefined", no more detailed error message in the log.

sedouard:
Sorry about that @sunbinzhu we're still looking into an issue which causes intermittent connection failures to the validation server. Re-ran the PR and it passes

create-hpc-cluster-linux-cn/azuredeploy.json
```diff
@@ -159,400 +159,122 @@
     }
   },
   "variables": {
-    "apiVersion": "2015-05-01-preview",
-    "location": "[resourceGroup().location]",
     "virtualNetworkName": "[concat(parameters('clusterName'), 'VNet')]",
     "storageAccountNamePrefix": "[toLower(concat('hpc', uniqueString(resourceGroup().id, parameters('clusterName'))))]",
```
singhkay: You'll need to limit the clusterName parameter 8 characters otherwise the concat string will be more than 24 chars of the allowed storage account name. The other option is to do the below

```[concat('hpccluster', uniqueString(resourceGroup().id))]```

File: create-hpc-cluster/azuredeploy.json [hunk 2]
```diff
@@ -158,377 +158,97 @@
     }
   },
   "variables": {
-    "apiVersion": "2015-05-01-preview",
-    "location": "[resourceGroup().location]",
     "virtualNetworkName": "[concat(parameters('clusterName'), 'VNet')]",
     "storageAccountNamePrefix": "[toLower(concat('hpc', uniqueString(resourceGroup().id, parameters('clusterName'))))]",
```
singhkay: Same comment for storage account name here as above

create-hpc-cluster/mainTemplate.json
```diff
@@ -0,0 +1,559 @@
...
+      }
+    }
+  },
+  "variables": {
+    "apiVersion": "2015-06-15",
+    "computeNodeNumber": "[int(parameters('computeNodeNumber'))]",
+    "cnStorageAccountNumber": "[add(div(sub(variables('computeNodeNumber'), 1), parameters('nbrCNPerStorageAccount')), 1)]",
+    "cnStorageAccountStartIndex": "[add(parameters('storageAccountStartIndex'), 1)]",
+    "storageAccountIndexWidth": "[length(string(sub(add(variables('cnStorageAccountNumber'), variables('cnStorageAccountStartIndex')), 1)))]",
+    "hnStorageAccountName": "[concat(parameters('storageAccountNamePrefix'), padLeft(string(parameters('storageAccountStartIndex')), variables('storageAccountIndexWidth'), '0'))]",
```
singhkay: Make sure the inputs don't cause this to be more than 24 chars

create-hpc-cluster/azuredeploy.json
sunbinzhu: @singhkay please note clusterName is an input for function uniqueString, but not for function concat. The output for function uniqueString is 13 characters, so the value of storageAccountNamePrefix is always with 16 characters.

create-hpc-cluster/mainTemplate.json
sunbinzhu: It is guaranteed in azuredeploy.json. The storageAccountNamePrefix is 16 characters, there is enough name space (a maximum 8 characters for index).

create-hpc-cluster-linux-cn/azuredeploy.json
sunbinzhu: @singhkay Thanks for the advice for storage account naming. But the name here is for storageAccountNamePrefix, so whatever the prefix is, we will create a series of storage accounts with same prefix. Putting the uniqueString at the beginning of the nameprefix doesn't really help to distribute the storage accounts on different racks.

singhkay:
ltgm, merged

PR #1025, opened 2015-12-22
CId templates

singhkay:
* 101 vm ssh
* 101 vm multiple data disks


singhkay:
Merging, user image template is expected to fail without existing setup

PR #1026, opened 2015-12-22
CI continues

singhkay:


singhkay:
@sedouard Any ideas why the error is 
AssertionError: 101-webapp-with-golang/azuredeploy.parameters.json -  Parameter "enableAlwaysOn" is missing its value field

The parameters file has the value for the parameter enableAlwaysOn and this template deploys fine locally with PS.

sedouard:
ugh tiny bug. opened #1027 I'll fix it real quick.

singhkay:
@sedouard Awesome! Thanks!

Issue #1022, opened 2015-12-22
End of year template cleanup log...[DEVELOPING]


singhkay:
It's been a great year with the repo receiving an amazing number of templates across an even more amazing breadth of Azure services. As we've progressed there's been some templates that have become repetitive and don't serve a good purpose. This logs the templates we're removing/merging or moving to different paths.

------------------
REMOVED
* 101-create-backup-vault-policy -- Fails consistently
* 101-simple-windows-vm-data-disk -- Merged with 101-simple-windows-vm
* 101-vm-simple-linux-vm-data-disk -- Merged with 101-vm-simple-linux-vm
* 101-virtual-network -- Merged into 101-vnet-two-subnets
* 101-create-availability-set -- Merged into 101-create-availability-set-3FDs-20UDs

-----------------

Developing...

PR #1036, opened 2015-12-28
End of the year template cleanup

singhkay:
Organized templates into a common prefix group
Removed templates that were repetitive
Updated Readme deploy to azure buttons

More details in #1022


singhkay:
CI canceled as there are too many templates

PR #1038, opened 2015-12-28
Update domain join template to use the extension

singhkay:


singhkay:
Skipping CI since this needs an existing domain controller to be setup

PR #1053, opened 2015-12-31
VM Scale Sets as Lustre clients

arsenvlad:


intel-lustre-clients-vmss-centos/azuredeploy.json
```diff
@@ -0,0 +1,414 @@
...
+      "defaultValue": "password",
+      "allowedValues": [
+        "password",
+        "sshPublicKey"
+      ],
+      "metadata": {
+        "description": "Authentication type for the virtual machines"
+      }
+    },
+    "adminPassword": {
```
singhkay: I think one of the default specified above in authenticationType shouldn't have a default value of empty string. If this template is deployed without a parameters file in command line, PS or CLI will skip over any parameters that have a default value specified causing the deployment to fail.

intel-lustre-clients-vmss-centos/azuredeploy.json
```diff
@@ -0,0 +1,414 @@
...
+      }
+    }
+  },
+  "variables": {
+    "apiVersion": "2015-06-15",
+    "storageAccountType": "Standard_LRS",
+    "saCount": 5,
+    "newStorageAccountSuffix": "[concat(parameters('vmssName'), 'sa')]",
+    "uniqueStringArray": [
+      "[concat(uniqueString(concat(resourceGroup().id, deployment().name, variables('newStorageAccountSuffix'), '0')))]",
```
singhkay: There was a bug report a few months ago that said have a deployment().name as a seed in uniquestring was a bad idea for testing and deploying incremental resources into the same environment. Each time you specified a different deployment name while adding resources to the same template, it would result in different storage accounts to be spun up.

intel-lustre-clients-vmss-centos/azuredeploy.json
arsenvlad: For this specific template, to be able to provide "either-or" functionality, so either SSH -or- the password can be used, both parameters have to support empty default values. The user will provide the proper value for the parameter based on what they select for authenticationType (since there is no way to have a conditional required flag on parameters)

intel-lustre-clients-vmss-centos/azuredeploy.json
```diff
@@ -0,0 +1,414 @@
...
+    },
+    {
+      "type": "Microsoft.Network/publicIPAddresses",
+      "name": "[variables('publicIPAddressName')]",
+      "location": "[parameters('location')]",
+      "apiVersion": "[variables('apiVersion')]",
+      "properties": {
+        "publicIPAllocationMethod": "Dynamic",
+        "dnsSettings": {
+          "domainNameLabel": "[concat('vmss-',parameters('vmssName'),'-',uniqueString(resourceGroup().id,deployment().name))]"
```
singhkay: See comment above about deployment name seed in uniqueString

arsenvlad: I removed the deployment().name from the uniqueString for this template

singhkay:
lgtm, merged

PR #1055, opened 2015-12-31
Refresh create-hpc-cluster-custom-image

sunbinzhu:
1. Refresh the template create-hpc-cluster-custom-image, the new image shares similiar parameter sets, and shares some linked templates/scripts with the other two create-hpc-cluster templates.
2. Add post configuration custom script support on compute nodes.


sunbinzhu:
Regarding to the issues that acomghbot reported:
1. The file removed is not used anymore.
2. "uri(deployment().properties.templateLink.uri" is the way to get the current template as documented in https://azure.microsoft.com/en-us/documentation/articles/resource-group-linked-templates/ (the last example)

sunbinzhu:
It won't surprise me if the CI fails in create-hpc-cluster-custom-image, because it requires a valid URL of the custom image VHD as a parameter, but it is not possible for me to provide one in azuredeploy.parameters.json. Believe me, I have done plenty of tests.

sedouard:
@singhkay @sunbinzhu can't we just add this custom vhd to a storage account on the CI test subscription?

sunbinzhu:
@sedouard @singhkay Yes, It shall work if we can create a custom VHD with HPC Pack installed. One thing worth mentioning is that all the VMs created with this custom VHD will have their OS VHDs stored in the same storage account.i.e. They won't be cleared when you delete the new resource group created by the new deployment. You need clear them separately.

sunbinzhu:
And i just checked the error logs, in fact it failed at the syntax validation step, it complained the same issue as acomghbot did. Could you update the syntax validation rule? As i mentioned, the syntax is correct and works.

sedouard:
If you provide a link we can probably create a prerequisite VHD storage account which we can link future templates to. That is simple enough to do.

With respect to the your template syntax validation, this is a rule enforced by the official Azure ARM validation API.

sunbinzhu:
Ok, i have just fixed the syntax validation issue.
The custom image can be found at https://suzhubasesa2.blob.core.windows.net/vhds/iaasbasecn-000-osdisk.vhd, i have added you two as readers of the storage account.

sedouard:
@sunbinzhu can you please make the blob public read so we can download it.

sunbinzhu:
@sedouard  The CI check passed after the custom image is specified, can you help to merge?

sedouard:
I generally try to keep avoid merging templates. I just help out with CI issues :-). I'll defer to @gatneil to push the big green button

sunbinzhu:
@sedouard @singhkay @gatneil somebody please help to merge the commit. thanks

singhkay:
@sunbinzhu Looks like there are quite a bit of changes. Will try to review by Monday

create-hpc-cluster-custom-image/azuredeploy.json
```diff
@@ -2,23 +2,17 @@
   "$schema": "https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#",
   "contentVersion": "1.0.0.0",
   "parameters": {
-    "namePrefix": {
+    "clusterName": {
```
singhkay: Add the min and max properties enforce length on the parameter

sunbinzhu: can you give an example?

sunbinzhu: Added

sunbinzhu: yes, it has been a long time since the template create-hpc-cluster-custom-image was refreshed last time because it is low priority by comparing with other two templates (create-hpc-cluster and create-hpc-cluster-linux-cn). We finally have time to refresh it now.

singhkay:
lgtm, merging

PR #1057, opened 2015-12-31
CI part VIII

gatneil:


singhkay:
@gatneil everything passed except the express route template. Merging.

PR #1122, opened 2016-01-20
docker-swarm: separate nsg for nodes/masters

ahmetalpbalkan:
Separating firewall for nodes and masters as users need to add an
AllowAny (*->*) rule to make their containers accessible from
public Internet through Azure LB.

Also adding autoUpgradeMinorVersion to pick up DockerExtension v1.1+.

**Manually tested.**


singhkay:
lgtm, merged

PR #1131, opened 2016-01-21
Multiple fixes to SharePoint Templates

simongdavies:


singhkay:
@simongdavies Can you CI the parameters files? https://github.com/Azure/azure-quickstart-templates/tree/master/1-CONTRIBUTION-GUIDE#travis-ci Also looks like the parameters file for ha template doesn't have the schema style declaration

singhkay:
@simongdavies Looks like this template takes longer to deploy than TravisCI's limit. I'll verify by deploying locally.

"The job exceeded the maxmimum time limit for jobs, and has been terminated."

singhkay:
@simongdavies Deployed locally succesfully. However, this parameter confused me if it needed a value or not in case of new value in spWebIPNewOrExisting. Maybe this should have a default value of "" as well https://github.com/simongdavies/azure-quickstart-templates/blob/master/sharepoint-server-farm-ha/azuredeploy.json#L196

Also spWebIPNewOrExisting should probably be moved above spWebIPRGName and spWebIPAddressName

simongdavies:
@singhkay updated as suggested, deploy is running again but will probably timeout

singhkay:
lgtm, merged

PR #1136, opened 2016-01-22
Fix to RM template that creates a DevTestLab with policies

mithunshanbhag:
As per recent changes in the microsoft.DevTestLab Resource Provider, we're making a minor change to fix the RM template that creates a DevTestLab with custom policies.


singhkay:
merged

PR #1166, opened 2016-02-01
New template: Add subnet to existing vnet

singhkay:
Add subnet to existing vnet


singhkay:
tested passing locally. Because it needs an existing VNET so won't pass CI.

PR #1167, opened 2016-02-01
Deprecate old swarm-simple template

ahmetalpbalkan:
Deleting the old single-master Swarm template for the sake of a better
one that exists now.


singhkay:
@ahmetalpbalkan Merging, but can you edit the OP to put the link in to the better one

PR #1168, opened 2016-02-02
Modernize template

charlesjoy:
Remove location parameter
Update DSC Extension version and name
Add autoUpgradeMinorVersion

Tested in both Azure and Azure Stack - Reminder: this template is just for an extension, so it  cannot automatically be run against Azure without an existing VM (for BOT validation).


charlesjoy:
I see that the check has failed, this is to be expected for this Template, as I am not filling in specific string data for all parameters.

This was tested in both Azure and Azure Stack - Reminder: this template is just for an extension, so it cannot automatically be run against Azure without an existing VM (for BOT validation).

singhkay:
@charlesjoy Sure. lgtm, merging

PR #1170, opened 2016-02-03
Templates cleanup

singhkay:


singhkay:
cancelling build and merging since these are name changes.

PR #1177, opened 2016-02-04
docker-swarm: multi-host overlay networking

ahmetalpbalkan:
This updates the swarm template with Docker 1.9 multi-host networking features so that users can easily do `docker network create` and get overlay networks out of the box.

We're switching to CoreOS Beta feed because latest CoreOS Stable version as of this change is v835.12.0 and it does not have Docker 1.9 yet. We shall switch back to the stable channel once this situation changes.


ahmetalpbalkan:
FYI @anhowe @rgardler you might wanna integrate the changes to ACS.

singhkay:
lgtm, merged

PR #1188, opened 2016-02-08
Update SP HA Template

simongdavies:
improves reliability of SP HA template and also reduces deployment time


singhkay:
lgtm, merged

PR #1197, opened 2016-02-09
Secure password template

singhkay:


singhkay:
Skipping CI since this needs pre-setup

PR #1203, opened 2016-02-10
updated commands to set properties on the vault

bmoore-msft:
The EnableForDeployment property is *not* required for getting secrets during template deployment.  Only EnabledForTemplateDeployment is required.


singhkay:
@bmoore-msft thanks for catching! merged

PR #1228, opened 2016-02-15
Updated the Templates to now use VMSS

ChackDan:
Updated the Templates to now use VMSS


ChackDan:
closing this PR, since it seems to be stuck.

PR #1235, opened 2016-02-17
Master

OJDUDE:
Fixed Deploy and Visualize buttons link after renaming the folder. They still pointed to the older folder.


singhkay:
merged

PR #1244, opened 2016-02-17
Add new ARM template for SIOS DataKeeper SQL cluster

carrollh:
Initial commit to azure-quickstart-templates @azurecla


azurecla:
Hi __@carrollh__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
    </span>

TTYL, AZPRBOT;

singhkay:
@carrollh Looks like the CI failed with itemDisplayName in metadata file too long. Please rename to less than 60 chars. I'd suggest itemDisplayName : "SQL Server 2014 Failover Cluster with SIOS DataKeeper"

carrollh:
I have fixed that issue, but there seems to be another one the CI fails to recognize. I can publish the template successfully if I change out the baseURL for my fork in my azuredeploy.json file, and then swap out the GEN-UNIQUE-20 and GEN-PASSWORD entries in the azuredeploy.parameters.json file for valid values. The actual error noted in the CI details page is less than useful:

[cid:fdf72e15-90e4-4ec3-8a81-ced7635f5406]

Thanks for any insight.

________________________________

carrollh:
This version builds from powershell if you replace the GEN-UNIQUE params with valid ones. The error being reported is "Server Error". Do I just need to re-push?

carrollh:
This time it is failing to publish our image complaining about programmatic deployment not having been enabled. Which it is, and has been for months. 
image
Otherwise I wouldn't be able to deploy it with powershell, which I can. Please advise.

singhkay:
@carrollh Ah I see, I need to enable that image for the CI sub

carrollh:
@singhkay Thanks for the prompt response!
Unfortunately I fear this build will also fail if CI is doing more than simple validation checks. The custom image requires a license as one of the params. Currently that value is effectively a "CHANGE ME" in the json file. When the last stage of the deployment is reached it will eventually fail as the service requiring that license will never start. It uses DSC modules to activate the product and continue to provision everything. Is the CI expecting a fully deployable solution with dummy params? I can create a 1 day long license if need be to pass testing, but then the default value in the param field will look to the end users like it doesn't need to be changed.

carrollh:
Actually, it failed with the same error as before. 
image

singhkay:
@carrollh Ah I see! Let's go ahead and create a 1 day license and then replace with change_me once it passes

carrollh:
@singhkay This is likely still going to fail due to the purchase action failing permissions / programmable deployment hiccup. Unless you fixed that already, in which case thanks! 
Also, would you mind replacing the value for the licenseKeyFtpURL param with this instead of change_me?
"Contact support@us.sios.com for a free 14 day trial license key."

Thanks for your efforts.

singhkay:
@carrollh Sure, that string works as well. 

Checked again and programattic access is enabled. Is this the image being used? 

image

carrollh:
@singhkay I'm not sure if you can see it. I can't even see the production version. We didn't want it being randomly deployed so it is a hidden marketplace image. The title of it is "SIOS DataKeeper Cluster Edition with SQL installer". I can send you the sku and offer if you need them.

singhkay:
@carrollh That's probably it. Any plans to make it public image?

carrollh:
@singhkay To be clear, I can see it in the publisher's portal, just not in the azure portal. I can only see the staged version at portal.azure.com. I can deploy it with my personal account, but when it was first published there was a weird error that looked similar to this. It was wider spread than just our image though. 
There are no plans to make it visible at the moment. We want it deployable only via this template. The image you pointed out is what we want generally available.
Thanks.

singhkay:
@carrollh But if users can't enable programmatic access for the image in the Azure Portal because they can't find it, how will they deploy from this template? It will always fail with the error, need to enable programmatic access.

carrollh:
@singhkay It doesn't make sense that the customer would need to do this. Which is why I'm confused as to why the CI is even getting this error. This would be analogous to individual users being responsible for enabling programmatic deployment of the Windows Server 2012 R2 marketplace image for their account. Users don't have that control. The azure dev team sets that. I set it in the publishing portal for our image, but when using my personal account to deploy our template it works. I can't even see our template image in the marketplace using that account. 
This error shouldn't even be happening is what I'm getting at.

singhkay:
@carrollh Let me check with the marketplace team and get back to you

carrollh:
@singhkay Sure. Thanks!

singhkay:
Hi @carrollh, checked with the Marketplace team. What you need to do is to have this template go through the marketplace publishing process. Then the template will be available through the "Create a VM" workflow in azure.portal.com. In that workflow, customers don't have to enable programmatic access for images. The UI allows you to show T&Cs for a hidden image and allows a customer to deploy with your hidden image. 

Having a template in this repo, you'll need to point to an image that is not hidden and the customer will need to enable programmatic access from the Portal.

carrollh:
@singhkay Thanks for the guidance, but that seems odd considering the template is just a mod (for quickstart compliance) of the template located here:
https://github.com/SIOSDataKeeper/SIOSDataKeeper-SQL-Cluster

And that template is fully functional using the same image that is breaking your CI backend. Try it out. Use this for the license key url:
http://ftp.us.sios.com/pickup/EVAL_Joe_User_Azure_Testing_2016-02-24_DKCE/

Since that works with the marketplace image hidden, doesn't that prove the issue lies in the CI checks? Or is the fact it is working with standard user accounts a bug?

Thanks, again. I'll update my team with this info.

carrollh:
@singhkay We'd like to get the template on the quickstart page. Is there a way to get our image (not the tamplate) deployable without the user specifying the programmatic access? i.e. like most of the standard os images in the marketplace (Windows from Microsoft, Ubuntu from Canonical, etc)? Those don't have that option at all. I'm fine with publishing a new image, but that option isn't in the docs for marketplace image publication. 
Would publishing the template through the marketplace allow it to be placed on the quickstart page, or is this a lost cause?

Thanks.

singhkay:
@carrollh Finally figured this out. I had to deploy the template first using the custom template blade in Azure Portal and accept the terms in the last step before deploying. This will also happen once users click the "Deploy to Azure" button for your template. But they need to deploy from there first before being able to deploy through the clients. So it'd be great if you could leave this as a gotcha in the template readme. The other thing is, I saw some inconsistencies in the templates where the cases of parameters and where they are used is different (e.g. adminusername vs adminUsername). If you load the templates in Visual Studio or VS Code with Azure Resource Manager extension, you'll see the red squiggles/errors.

carrollh:
@singhkay 
That's great news. I've updated the json files that call the powershell dsc scripts to match. I built the template in VS2015 actually, and it doesn't seem to care for some reason. I fixed them anyway. I'm working on the readme now. What exactly do you mean by "deploy through the clients" above. The only way I know of deploying these things is via the "deploy to azure" buttons or using powershell. I suppose just cutting and pasting the contents of the azureDeploy file into the New Deployment Template blade in the azure portal would work too since all the links are in that file. I just want to make sure I know what I'm talking about when I add the note to the readme.
Thanks!

singhkay:
@carrollh By clients I mean anything that's not the Deploy to Azure button workflow (PS, CLI, SDKs, REST APIs etc.).

singhkay:
@carrollh Can you sign the CLA before I can merge?

carrollh:
@singhkay 
The original one expired a few days ago, so I created a new one and forwarded it to my program manager yesterday. Thanks for your patience.

singhkay:
@carrollh No worries!

singhkay:
@carrollh Any update on this?

singhkay:
@carrollh Let's close this PR in the meantime and follow up with another when you are ready. Sounds good?

carrollh:
@singhkay
Sounds good. Thanks.

carrollh:
@singhkays 
We're getting back around to this. Can you point me at the current version of the CLA so I can pass it along before opening a new pull request. Thanks!

PR #1247, opened 2016-02-17
template for mobile apps

tfitzmac:


tfitzmac:
@mattchenderson - here is the PR for the mobile apps template

singhkay:
@tfitzmac Looks like it's failing with the error that the namespace is not available. Can you take a look?

tfitzmac:
@mattchenderson - I had to make some changes to the template. It was using old api versions. When I updated them, it changed some of the properties that are required. I have tested the deployment and everything worked, but you might want to double-check the current azuredeploy.json file.

singhkay:
@tfitzmac interesting error :open_mouth: Is one week special?
`"error": "Deployment provisioning state was not successful\nThe namespace name is previously used in another project and has been deleted less than one week.\n",`

tfitzmac:
@singhkay - not sure why NotificationHub returns that error. I added uniquestring to fix the error. Everything looks good now.

singhkay:
lgtm, merged

PR #1263, opened 2016-02-19
Add a parameter to provide resource group of KeyVault

SudhakaraReddyEvuri:
These changes are to add a parameter to provide resource group of key vault while enabling encryption.


singhkay:
lgtm, merged

PR #1286, opened 2016-02-23
Update azuredeploy.json to take vmName as a param

sondresb:
Update azuredeploy.json to take vmName as a param, possibly through slightly clumsy use of github (more used to the Bitbucket UI and did not have the repo cloned locally, but I think this change is something one pretty much always want).

### DIFF
--- a/101-vm-from-user-image/azuredeploy.json
+++ b/101-vm-from-user-image/azuredeploy.json
@@ -47,14 +47,20 @@
       "metadata": {
         "description": "This is the size of your VM"
       }
+    },
+    "vmName": {
+      "type": "string",
+      "metadata": {
+        "description": "This is the name of your VM"
+      }
     }
   },
   "variables": {
     "location": "[resourceGroup().location]",
     "publicIPAddressName": "userImagePublicIP",
-    "vmName": "userImageVM",
-    "virtualNetworkName": "userImageVNET",
-    "nicName": "userImageVMNic",
+    "vmName": "[parameters('vmName')]",
+    "virtualNetworkName": "[concat(parameters('vmName'),'VNET')]",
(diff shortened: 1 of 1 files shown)


azurecla:
Hi __@sondresb__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
    </span>

TTYL, AZPRBOT;

singhkay:
@sondresb What does this accomplish? We try to reduce the number of parameters a user has to enter before deploying the template. What about using the [concat(resourceGroup().name + "vm")] for the vmName variable?

sondresb:
Well, if you think that would be better for the generic use case, absolutely - feel free to change. From my point of view what you change would not be usable (and I think everyone would want to name their instances, but doing based on resourceGroup().name might be fine for the sample template). What I want to do usually would be to use this for instance three times with names my-appserver-0[123] for instance.

But totally fine if you change (or even discard) the PR, I basically just felt it was good courtesy to share it back after getting help over in the comments section at https://azure.microsoft.com/en-us/documentation/articles/virtual-machines-linux-create-upload-vhd/#comment-2529047023

Thanks,

-S-

singhkay:
@sondresb Definitely appreciate the contribution! No worries about that however the general principal with 101 and basic templates is that since these are meant to be for the new users to ARM, we're trying to not have them enter too many params before deploying the template. I will have to close and change to resourceGroup().name in another PR but I'd much rather see another contributor to the repo :), so if you don't mind can you change to resourceGroup().name + vm suffix and do another commit?

sondresb:
Sure, I'll create another PR with a variant that uses a unique vm name
based on the rg name - even if that variant will not be useful for me I can
see your point... :)

PR #1290, opened 2016-02-24
Unlisting DTL templates.

mithunshanbhag:
The DTL RP is undergoing some significant refactoring work. Our team has made a decision to temporarily unlist all existing DTL templates (except one for lab creation).

We'll re-publish the templates post stabilization (once we GA).


singhkay:
lgtm, merged

FYI: DTL templates will be back after changes to the RP are finalized

PR #1307, opened 2016-02-26
Hazelcast cluster template

sedouard:
Deploy a hazelcast cluster using the Azure discovery plugin.

### DIFF
--- a/hazelcast-vm-cluster/metadata.json
+++ b/hazelcast-vm-cluster/metadata.json
@@ -1,0 +1,8 @@
+{
+  "itemDisplayName": "Hazelcast Cluster",
+  "description": "Hazelcast is an in-memory data platform that can be used for a variety of data applications. This template will deploy any number of Hazelcast nodes and they will automatically discover each other.",
+  "summary": "Deploy a Hazelcast cluster within a VNet.",
+  "githubUsername": "sedouard",
+  "dateUpdated": "2015-10-26"
+}
+
--- a/hazelcast-vm-cluster/README.md
+++ b/hazelcast-vm-cluster/README.md
@@ -1,0 +1,20 @@
+# Deploy a Hazelcast Cluster
+
+[![Deploy to Azure](http://azuredeploy.net/deploybutton.png)](https://portal.azure.com/#create/Microsoft.Template/uri/https%3A%2F%2Fraw.githubusercontent.com%2FAzure%2Fazure-quickstart-templates%2Fmaster%2Fhazelcast-vnet%2Fazuredeploy.json)
+<a href="http://armviz.io/#/?load=https%3A%2F%2Fraw.githubusercontent.com%2FAzure%2Fazure-quickstart-templates%2Fmaster%2Fhazelcast-vnet%2Fazuredeploy.json" target="_blank">
+    <img src="http://armviz.io/visualizebutton.png"/>
+</a>
+
+[Hazelcast](https://hazelcast.com) is an in-memory data platform which can support a variety of data applications such as data grids, nosql data stores, caching and web session clustering.
+
+This template will deploy any number of Ubuntu Hazelcast nodes in a vnet using the [official Hazlecast Azure Discovery Provider](https://github.com/hazelcast/hazelcast-azure). Every node is installed with Hazelcast as an [upstart service](http://upstart.ubuntu.com/) and will continue to run even after subsequent restarts. Each node will discover every other node on the network automatically so you can add and remove nodes as you see fit.
+
+Use the **Deploy to Azure** button above to get started.
+
+Checkout Hazelcast's [official documentation](http://hazelcast.org/documentation/) to learn more on how to use Hazelcast.
+
+## Azure Service Prinicpal
+
+This template deploys resources that need read access the mangaement api's for the resource group this template is deployed to.
+
+You'll need to setup [Azure Active Directory Service Principal credentials](https://azure.microsoft.com/en-us/documentation/articles/resource-group-create-service-principal-portal/) for your Azure Subscription for this plugin to work. With the credentials, fill in the `aadClientId`, `aadClientSecret`, and `aadTenantId` parameters.
--- a/hazelcast-vm-cluster/azuredeploy.json
+++ b/hazelcast-vm-cluster/azuredeploy.json
@@ -1,0 +1,393 @@
+{
+  "$schema": "https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#",
+  "contentVersion": "1.0.0.0",
+  "parameters": {
+    "adminUserName": {
+      "type": "string",
+      "metadata": {
+        "description": "Username for hazelcast VMs"
+      }
+    },
+    "adminPassword": {
+      "type": "secureString",
+      "metadata": {
+        "description": "Password for hazelcast VMs, if authenticationType is set to 'password'"
+      }
+    },
+    "sshPublicKey": {
+      "type": "string",
+      "metadata": {
+        "description": "SSH Public key for hazelcast VMs, if authenticationType is set to 'sshPublicKey'"
+      }
+    },
+    "authenticationType": {
+      "type": "string",
+      "metadata": {
+        "description": "Authenitcation method for logging into Hazelcast VMs."
+      },
+      "allowedValues": [
+        "password",
+        "sshPublicKey"
+      ]
+    },
+    "clusterUserName": {
+      "type": "string",
+      "metadata": {
+        "description": "Username for hazelcast grid"
+      },
+      "defaultValue": "[parameters('adminUserName')]"
+    },
+    "clusterPassword": {
+      "type": "string",
+      "metadata": {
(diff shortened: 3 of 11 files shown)


hazelcast-vnet/azuredeploy-password.json
```diff
@@ -0,0 +1,381 @@
+{
+  "$schema": "https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#",
+  "contentVersion": "1.0.0.0",
+  "parameters": {
+    "newStorageAccountName": {
```
singhkay: Eliminate this. This opens up a can of worms. See #944

hazelcast-vnet/azuredeploy-password.json
```diff
@@ -0,0 +1,381 @@
...
+      "type": "string",
+      "metadata": {
+        "description": "Unique DNS Name for the Storage Account where the Virtual Machine's disks will be placed."
+      }
+    },
+    "location": {
```
singhkay: Probably leave this off since there are so many locations now. Do a variable with resourceGroup().location

hazelcast-vnet/azuredeploy-password.json
```diff
@@ -0,0 +1,381 @@
...
+      },
+      "defaultValue": "[parameters('adminUserName')]"
+    },
+    "clusterPassword": {
+      "type": "string",
+      "metadata": {
+        "description": "Password for hazelcast grid"
+      }
+    },
+    "dnsNameForPublicIP": {
```
singhkay: Call this dnsPrefix. See #966.

hazelcast-vnet/azuredeploy-password.json
```diff
@@ -0,0 +1,381 @@
...
+      "type": "string",
+      "metadata": {
+        "description": "Unique DNS Name for the Public IP used to access the Virtual Machine."
+      }
+    },
+    "instanceCount": {
```
singhkay: does this have a min and a max?

hazelcast-vnet/azuredeploy-password.json
```diff
@@ -0,0 +1,381 @@
...
+        "description": "The tag used to identify this cluster"
+      }
+    },
+    "clusterPort": {
+      "type": "string",
+      "metadata": {
+        "description": "The cluster port to identify this cluster"
+      }
+    },
+    "minHeapSize": {
```
singhkay: is this in MB, GB? Probably good to suffix the parameter with it

hazelcast-vnet/azuredeploy-sshPublicKey.json
```diff
@@ -0,0 +1,397 @@
...
+        "description": "The minumum heap size for each hazelcast jvm"
+      }
+    },
+    "maxHeapSize": {
+      "type": "string",
+      "metadata": {
+        "description": "The maximum heap size for each hazelcast jvm"
+      }
+    },
+    "aadClientId": {
```
singhkay: Interesting, how do these get used? Didn't see anything in the readme about setting up an AD application

hazelcast-vnet/azuredeploy-sshPublicKey.json
```diff
@@ -0,0 +1,397 @@
...
+      }
+    },
+    "vmSize": {
+      "type": "string",
+      "metadata": {
+        "description": "The size of each instance VM"
+      },
+      "allowedValues": [
+        "Standard_DS1",
+        "Standard_DS2",
```
singhkay: How about non premium VM sizes or does this app need premium storage?

hazelcast-vnet/azuredeploy-sshPublicKey.json
```diff
@@ -0,0 +1,397 @@
...
+        }
+      }]
+    },
+    "copy": {
+      "name": "createNics",
+      "count": "[sub(parameters('instanceCount'), 1)]"
+    }
+  }, {
+    "apiVersion": "[variables('apiVersion')]",
+    "type": "Microsoft.Compute/virtualMachines",
```
singhkay: is this a jumpbox VM that has a public up and is used to reach other VMs?

hazelcast-vnet/README.md
```diff
@@ -0,0 +1,14 @@
+# Deploy a Hazelcast Cluster
```
singhkay: What's the significance of the "vnet" in the folder name? How about hazelcast-cluster?

hazelcast-vnet/azuredeploy.json
```diff
@@ -0,0 +1,230 @@
...
+    "apiVersion": "2015-01-01"
+  },
+  "resources": [{
+    "name": "hazelcast-vnet",
+    "type": "Microsoft.Resources/deployments",
+    "apiVersion": "[variables('apiVersion')]",
+    "properties": {
+      "mode": "[parameters('deploymentMode')]",
+      "templateLink": {
+        "uri": "[concat(parameters('templateBaseUrl'), 'azuredeploy-', parameters('vmAuthMethod'), '.json')]",
```
singhkay: Btw, you don't need multiple files to password or ssh auth. It can be done in the same template. See here https://github.com/Azure/azure-quickstart-templates/blob/master/checkpoint-single-nic/azuredeploy.json#L137 

Saves a lot of repeated code

hazelcast-vnet/azuredeploy-password.json
sedouard: It's limited by your compute allocation on your azure account.

hazelcast-vnet/azuredeploy-sshPublicKey.json
sedouard: What's the best guidance here? The most important thing for this scenario is the amount of memory available, but there isn't any reason I'd like to limit the options of the kind of VMs you can use. (Plus, enumerating all possible ones becomes a lengthy list)

hazelcast-vnet/README.md
sedouard: A cluster doesn't necessarily need to be restricted to a vnet. For example 2 vnets and a gateway between the two is also valid.

There's also talk of another hazelcast template with VM scalesets in a couple months after the javasdk and plugin supports it.

So `vnet` indicates these are VMs within the same vnet forming a hazelcast cluster rather than a scaleset.

hazelcast-vnet/azuredeploy.json
sedouard: I was given a bad example :-(. This is supposed to be a basis for a market place template and so I followed this template:

https://github.com/Azure/azure-quickstart-templates/tree/master/marketplace-samples/simple-Linux-VM

Which is terrible.

hazelcast-vnet/README.md
singhkay: Hmm, scale set VMs are also in the same VNET, in fact in the new stack you can't have a vm without a VNET. How about naming this hazelcast-vm-cluster and in a few months time hazelcast-vmss-cluster?

hazelcast-vnet/azuredeploy-sshPublicKey.json
singhkay: Add some standard ones like the D_V2s so this parameter covers a good gamut of sizes if not all

sedouard:
@singhkay -- heads up . There's a [bug in the CLI](https://github.com/Azure/azure-quickstart-templates/issues/1315) version we were provided which fails on VM creation timeout.

This is probably what's causing those intermittent `forEach` failure's we're seeing in travis.

hazelcast-vm-cluster/metadata.json
```diff
@@ -0,0 +1,8 @@
+{
+  "itemDisplayName": "Hazelcast VNet",
```
singhkay: Name change to go with the folder name change?

hazelcast-vm-cluster/metadata.json
```diff
@@ -0,0 +1,8 @@
...
+  "description": "Deploy a Hazelcast cluster within a VNet.",
```
singhkay: The summary text looks should go in here as well

sedouard: You mean the summary text === description text?

sedouard:
@singhkay a good practice is to use [concat(uniqueString(resourcegroup().id), 'somesuffix')] as a storage account names. This helps out with CI because subsequent deployments with the same storage account name hangs up. I suppose the storage account domain name doesn't immediately become available. That's what was causing my tests to hangup.

singhkay:
@sedouard Yep, that's what I have in #944 :) With a sub-id seed, you'll always have the same name for your storage account in your template. uniqueString is just a base64 hash of the seed.

hazelcast-vm-cluster/metadata.json
singhkay: Yep, switch those. Also update the date

sedouard:
@singhkay all feedback addressed. ci passed. merge at at will.

singhkay:
lgtm, merged

PR #1312, opened 2016-02-28
301 subnet driven deployment

JF6:
Template refactored : Less parameters, can define directly the number of subnets. Performance improvements.


singhkay:
lgtm, merged

PR #1317, opened 2016-03-01
Create a VM with dynamic data disks

singhkay:
This template allows the user to create a VM with dynamic selection of data disks without creating a per size template with different number of data disks. The disksSelector.json template takes in the number of disks required as input and forms the expected data disk array. It then passes the data disks array back to the main template using its outputs section.


singhkay:
CI passed, merging

PR #1319, opened 2016-03-01
Fix Visualize and Deploy button links

sedouard:


sedouard:
Fixes bad links from proposed name changes in #1307

PR #1320, opened 2016-03-01
Fix hazelcast-vm-cluster metadata date updated

sedouard:


sedouard:
Fixes the `dateUpdated` field in `metadata.json`

PR #1341, opened 2016-03-07
Minor fixes to LAP stack autoscale demo

gbowerman:
Change to latest 14.04.4-LTS version. Deploy D1 (single core) instead of
D3 by default. Be consistent about VMSS naming convention.


singhkay:
lgtm, merged

PR #1342, opened 2016-03-07
remove default dns name, which forces interaction instead of prompting

squillace:
if you call from the cli, you get prompted for everything that does not have a default value. including a dns name that is "uniquename" will always force manual intervention:


singhkay:
lgtm, merged

squillace:
@singhkay Speedy Kay, man. Whoosh.

PR #1359, opened 2016-03-08
Auto upgrade minor version

gatneil:
Add autoUpgradeMinorVersion


gatneil:
The CI is failing because the automation-dsc template requires pre-created azure automation dsc server; manually tested

PR #1372, opened 2016-03-10
new 2-tier template using an already existing vnet

MSSedusch:
### Changelog
* remove SLES 11 from 3-tier template
* add 2-tier template that supports existing vnet

### Description of the change
minor fix: removing SLES 11 from possible values of 3 tier


MSSedusch:
template requires an existing vnet. maybe that is that reason why travis fails

PR #1379, opened 2016-03-11
2-tier template: new parameter for existing vnet

MSSedusch:
### Changelog
* remove SLES 11 SP4
* add vnet parametr to 2-tier template

### Description of the change
2-tier parameter can now use an existing vnet


singhkay:
lgtm, merged

PR #1388, opened 2016-03-15
Update templates for DSCForLinux extension 2.0

lizzha:
### Description of the change
1. update parameters according to DSCForLinux extension 2.0
2. add function to register to Azure Automation


201-dsc-linux-azure-storage-on-ubuntu/install-resource.parameters.json
```diff
@@ -1,38 +1,39 @@
...
-  "storageAccountKey": {
-    "value": "yourstorageaccountkey"
-  },
-  "containerName": {
-    "value": "yourcontainername"
-  },
-  "resourceZipFileName": {
-    "value": "yourmodule_version.zip"
-  }
+    "$schema": "https://schema.management.azure.com/schemas/2015-01-01/deploymentParameters.json#",
```
singhkay: A template should have one azuredeploy.parameters.json file. Why is this needed?

201-dsc-linux-azure-storage-on-ubuntu/pull-meta-mof.parameters.json
```diff
@@ -1,38 +1,39 @@
...
-  "storageAccountKey": {
-    "value": "yourstorageaccountkey"
-  },
-  "containerName": {
-    "value": "yourcontainername"
-  },
-  "mofFileName": {
-    "value": "your.meta.mof"
-  }
+    "$schema": "https://schema.management.azure.com/schemas/2015-01-01/deploymentParameters.json#",
```
singhkay: A template should have one parameters file, why is this needed?

201-dsc-linux-public-storage-on-ubuntu/pull-meta-mof.parameters.json
```diff
@@ -1,29 +1,33 @@
...
-  "ubuntuOSVersion": {
-    "value": "14.04.3-LTS"
-  },
-  "mode": {
-    "value": "Pull"
-  },
-  "mofFileUri": {
-    "value": "http://yourplace/your.meta.mof"
-  }
+    "$schema": "https://schema.management.azure.com/schemas/2015-01-01/deploymentParameters.json#",
```
singhkay: Similar questions as above to multiple parameter files

201-dsc-linux-azure-storage-on-ubuntu/install-resource.parameters.json
lizzha: DSCForLinux extension has multiple configuration settings for different functionality, so I used mulitple parameter files to explain the different usage. 
Just found that with the refactor in 2.0 version, some parameter files can be removed, I will remove the duplicated parameter files.

singhkay: I see that only one value differs between the various parameter file depending on parameter value. Let's use the readme to explain the expected value in each case rather than creating 4 parameter files which will be confusing.

lizzha:
@singhkay I've removed the duplicated files, could you please review it? And the CI failed seems due to some internal error, as there's no change that will affect the deployment after the last successful check.

singhkay:
lgtm, merged

PR #1402, opened 2016-03-17
Updating copy index templates to best guidance

singhkay:
### Changelog
* Storage account best practices
* Can now deploy both Linux and Windows VMs
* apiVersion is a variable


singhkay:
copy index is passing, merged

PR #1427, opened 2016-03-20
Remove blacklist entry for /dev/sdb?

dendle:
Hello All,
I've been struggling with missing disks using the elasticsearch template, deploying ubuntu VMs.

This script always succeeds in mounting at least one disk, but when two disks are attached to the VM, 90% of the time, one disk would not get mounted.

It appears as though the disk that doesnt get mounted appears at /dev/sdb.

Having reviewed the script, I can see that this has been intentionally skipped, in the BLACKLIST.

I am not an unbuntu expert, but changing this script as above meant that all my unbuntu deployments succeeded with all disks attached.

This pull request contains the simple change of removing the /dev/sdb disk from the blacklist variable, therefore allowing disks appearing there to be mounted.

Cheers,
Matt
### Description of the change
Sometimes, Azure attaches disks which turn up at /dev/sdb (ubuntu), and due to the previous blacklist setting, this disk was not getting mounted.

### DIFF
--- a/shared_scripts/ubuntu/vm-disk-utils-0.1.sh
+++ b/shared_scripts/ubuntu/vm-disk-utils-0.1.sh
@@ -62,7 +62,7 @@ then
 fi
 
 #A set of disks to ignore from partitioning and formatting
-BLACKLIST="/dev/sda|/dev/sdb"
+BLACKLIST="/dev/sda"
 
 # Base path for data disk mount points
 DATA_BASE="/datadisks"


azurecla:
Hi __@dendle__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
    </span>

TTYL, AZPRBOT;

singhkay:
@trentmswanson @hglkrijger Since this affects elasticsearch, can you review this?

dendle:
$ lsblk
NAME   MAJ:MIN RM   SIZE RO TYPE MOUNTPOINT
sda      8:0    0  1023G  0 disk
sdb      8:16   0  1023G  0 disk
sdc      8:32   0  29.3G  0 disk
└─sdc1   8:33   0  29.3G  0 part /
sdd      8:48   0   200G  0 disk
└─sdd1   8:49   0   200G  0 part /mnt
sr0     11:0    1   1.1M  0 rom

Now they're appearing on sda, too!

dendle:
OK I think, if the original reason for putting the blacklist idea in doesn't apply/is not needed anymore, we need to remove it completely.

Before last night, i've only seen extra disks attach to /dev/sdb and later, never sda.

Now i've seen them attach to sda, this means that if this blacklist stays as it is, up to two disks will not get mounted when running this template.

I've removed the blacklist on my fork, and the disks get mounted consistently.

I'm going to close this PR and make a new one with the blacklist removed.

Cheers,
Matt

dendle:
This PR is obsoleted by:
https://github.com/Azure/azure-quickstart-templates/pull/1435

PR #1435, opened 2016-03-21
fix unmounted disks in Ubuntu

dendle:
Further to #1427 

This PR is to ensure all attached disks get mounted when using the elasticsearch template (and it probably helps with a few others?)

The symptom is that after deploying the template, the elasticsearch nodes dont have the free disk space that they should.  Root cause is that the extra disks that are attached during the VM provisioning are not mounted by this script, due to a blacklist of devices not to mount.  If this blacklist is removed, the script runs and results in consistent disk mounting.

Cheers,
Matt
### Description of the change
Azure can mount extra attached disks to /dev/sda and /dev/sdb.
See below - two 1TB disks have been attached to this ubuntu 14 LTS gallery image:
$ lsblk
NAME MAJ:MIN RM SIZE RO TYPE MOUNTPOINT
sda 8:0 0 1023G 0 disk
sdb 8:16 0 1023G 0 disk
sdc 8:32 0 29.3G 0 disk
└─sdc1 8:33 0 29.3G 0 part /
sdd 8:48 0 200G 0 disk
└─sdd1 8:49 0 200G 0 part /mnt
sr0 11:0 1 1.1M 0 rom

Removed the blacklist setting so that this disks are mounted.

### DIFF
--- a/shared_scripts/ubuntu/vm-disk-utils-0.1.sh
+++ b/shared_scripts/ubuntu/vm-disk-utils-0.1.sh
@@ -61,9 +61,6 @@ then
     exit 3
 fi
 
-#A set of disks to ignore from partitioning and formatting
-BLACKLIST="/dev/sda|/dev/sdb"
-
 # Base path for data disk mount points
 DATA_BASE="/datadisks"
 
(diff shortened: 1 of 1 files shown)


azurecla:
Hi __@dendle__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
    </span>

TTYL, AZPRBOT;

hglkrijger:
It looks ok to me. I also thought sda and sdb were reserved for the OS and temp drive, but that doesn't seem to be the case. Perhaps @trentmswanson should also take a look, since he put in the blacklist originally, but I would think partx then file is a pretty robust check for not wiping the other drives.

trentmswanson:
That's correct, at the time it was added to avoid wiping the OS and temp drives. It should not be necessary as we check to see if there's a partition and it's formatted before doing anything. I don't have time right now for a thorough review and testing but the change should be fine.

singhkay:
Thanks @trentmswanson @hglkrijger! merged

Issue #1398, opened 2016-03-16
elasticsearch template fails with invalid template


sandyplace:
Used the default parameters.

The template deployment 'Microsoft.Template' is not valid according to the validation procedure. The tracking id is '519958b7-ff4e-4204-92dd-d3dcd1e33f58'. See inner errors for details. Please see http://aka.ms/arm-deploy for usage details.

singhkay:
@sandyplace Is this the template in question? https://github.com/Azure/azure-quickstart-templates/tree/master/elasticsearch

singhkay:
@hglkrijger Tried deployed it and ran into an error because of the location parameter not having all locations a resource group can be deployed in. Let's remove the location parameter and create a variable with resourceGroup().location

sandyplace:
Yes. Sorry should have been more specific in my issue description.

hglkrijger:
@singhkay Sure, I will look into this.

mandarjadhav:
Facing the same issue with the latest template. 
The template deployment 'Microsoft.Template' is not valid according to the validation procedure. The tracking id is '83a46749-979e-402a-b6fa-37d7c63ae59a'. See inner errors for details. Please see http://aka.ms/arm-deploy for usage details.

hglkrijger:
@mandarjadhav can you describe how you deploy and which parameters you select?

mandarjadhav:
I am using default parameters.

singhkay:
@mandarjadhav I can't repro it with the latest update. It deploy succesfully for me

![image](https://cloud.githubusercontent.com/assets/6164178/14054365/eba02d7c-f295-11e5-8ed0-82827a4e7661.png)

PR #1439, opened 2016-03-21
afs improvements

hglkrijger:
-add caching for afs with cachefilesd (disabled for now)
-add some additional logging to scripts
-update elasticsearch and kibana versions
-remove location from the template and always use resource group [fix for #1398]
-enable AFS for master and client nodes
-separate afs and datadisks locations


singhkay:
manually deployed succesfully, merged

PR #1462, opened 2016-03-24
Elasticsearch template improvements

hglkrijger:
- use a restricted credentials file instead of storing the share password in /etc/fstab
- add the resource files contained in testpack.zip for change tracking


singhkay:
lgtm, merged

Issue #1497, opened 2016-03-29
Elasticsearch Windows cluster deployment fails


kirpasingh:
[Elasticsearch template](https://azure.microsoft.com/en-us/documentation/templates/elasticsearch/)

### Issue Details
Windows cluster deployment fails with the following error:
```
    "properties": {
      "statusCode": "Conflict",
      "statusMessage": "{\"status\":\"Failed\",\"error\":{\"code\":\"ResourceDeploymentFailure\",\"message\":\"The resource operation completed with terminal provisioning state 'Failed'.\",\"details\":[{\"code\":\"DeploymentFailed\",\"message\":\"At least one resource deployment operation failed. Please list deployment operations for details. Please see http://aka.ms/arm-debug for usage details.\",\"details\":[{\"code\":\"BadRequest\",\"message\":\"{\\r\\n  \\\"error\\\": {\\r\\n    \\\"code\\\": \\\"InvalidParameter\\\",\\r\\n    \\\"target\\\": \\\"linuxConfiguration\\\",\\r\\n    \\\"message\\\": \\\"The value of parameter linuxConfiguration is invalid.\\\"\\r\\n  }\\r\\n}\"},{\"code\":\"BadRequest\",\"message\":\"{\\r\\n  \\\"error\\\": {\\r\\n    \\\"code\\\": \\\"InvalidParameter\\\",\\r\\n    \\\"target\\\": \\\"linuxConfiguration\\\",\\r\\n    \\\"message\\\": \\\"The value of parameter linuxConfiguration is invalid.\\\"\\r\\n  }\\r\\n}\"},{\"code\":\"BadRequest\",\"message\":\"{\\r\\n  \\\"error\\\": {\\r\\n    \\\"code\\\": \\\"InvalidParameter\\\",\\r\\n    \\\"target\\\": \\\"linuxConfiguration\\\",\\r\\n    \\\"message\\\": \\\"The value of parameter linuxConfiguration is invalid.\\\"\\r\\n  }\\r\\n}\"}]}]}}"
    },
```
It's a strange message since I'm deploying Windows, not a Linux cluster. Some parameter name "linuxConfiguration" is messed up in the template json, I believe.

### Repro steps 
Deploy a Windows cluster with following parameters:
```
$parameters=@{
    adminUserName='testuser'
    adminPassword='hidden'
    virtualNetworkName='pingtestVnet'
    esClusterName='pingtest'
    loadBalancerType='external'
[...]

PR #1499, opened 2016-03-29
fix for #1497

hglkrijger:
fix for #1497 - deployment fails due to bad parameter format, this is now fixed


singhkay:
lgtm, merged

PR #1533, opened 2016-04-04
Changed container versions to stick to the latest v1.0.0 for server/n…

kvaes:
Changelog
- Base Template Fix
- Emptied default value for nodesApi (as it's unique per server deployment)
- Stickied versions for compatibility between agent/server


singhkay:
lgtm, merged

PR #1538, opened 2016-04-05
made saCount based on length of uniqueStringArray instead of hardcoded at 5

gatneil:
makes it easier to change later


singhkay:
@gatneil Mostly passing but you might want to see why the extension failure is happening  `"error": "Deployment provisioning state was not successful\nThe resource operation completed with terminal provisioning state 'Failed'.\nVM has reported a failure when processing extension 'Microsoft.Insights.VMDiagnosticsSettings'. Error message: \"Error while parsing status of Handler (name Microsoft.Azure.Diagnostics.IaaSDiagnostics, version 1.6.1.0)/0 against sequence number 0.\".\n",`

PR #1548, opened 2016-04-05
Fixing availability sets for rds roles and description for some param…

MahendraAgrawal:
Fixing availability sets for rds roles (CB, GW and RDSH) and description for some parameters to help users.


singhkay:
Expected failure since custom image required. merging.

PR #1549, opened 2016-04-05
Elasticsearch template updates

hglkrijger:
- add support for the cloud-azure plugin to Windows
- update the versions of Elasticsearch to include 2.3.1 and 2.2.2, Kibana to 4.5.0
- disable the standalone Marvel cluster by default, as this prevents the use of Kibana over the main cluster without further manual steps


singhkay:
lgtm, merged

PR #1581, opened 2016-04-14
Use swarm latest image instead of 1.0.1

dongluochen:
### Description of the change
The current swarm template fixes on swarm:1.0.1 which is a little outdated. We could either use a new version, or always use the latest version. Given swarm is relatively stable I think it ok to use latest version. And we don't need to update it frequently. 

Signed-off-by: Dong Chen <dongluo.chen@docker.com>

### DIFF
--- a/docker-swarm-cluster/azuredeploy.json
+++ b/docker-swarm-cluster/azuredeploy.json
@@ -499,7 +499,7 @@
               "restart": "always"
             },
             "swarm": {
-              "image": "swarm:1.0.1",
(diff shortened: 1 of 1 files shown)


azurecla:
Hi __@dongluochen__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
    </span>

TTYL, AZPRBOT;

singhkay:
lgtm, merged

PR #1587, opened 2016-04-15
Location fix

marcvaneijk:
Changed resource location from parameter to resourceGroup().location. Deleted location parameter from deployment template and parameter file.


singhkay:
Thanks for the fix @marcvaneijk! It looks like this template's parameters file is out of sync with the expected format. See example here https://github.com/Azure/azure-quickstart-templates/blob/master/101-acs-mesos/azuredeploy.parameters.json. Would you mind fixing that as well?

singhkay:
lgtm, merged

PR #1595, opened 2016-04-15
Location fix

marcvaneijk:


singhkay:
@marcvaneijk DataLake is currently only available in East US 2 so resourceGroup().location won't make sense here. 

You can also check for available locations for resources with this PS command

```
(Get-AzureRmResourceProvider -ProviderNamespace Microsoft.DataLakeStore).Locations
```

PR #1617, opened 2016-04-15
Location fix

marcvaneijk:


singhkay:
@marcvaneijk Error `AssertionError: 101-site-to-site-vpn-create/azuredeploy.parameters.json -  Parameter "sharedKey" is missing its value field`

PR #1624, opened 2016-04-15
Missing subscriptionId param in 101-rbac-builtinrole-virtualmachine

lorenpaulsen:
I was not able to use this ARM template as-is because of the reference to undefined parameter `subscriptionId`. Other similar samples infer the subscription ID through `subscription().subscriptionId` so I have made the same substitution here.


azurecla:
Hi __@lorenpaulsen__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
    </span>

TTYL, AZPRBOT;

singhkay:
@lorenpaulsen Thanks for fixing! Could you also update the metadata file according to convention? https://github.com/Azure/azure-quickstart-templates/tree/master/1-CONTRIBUTION-GUIDE#metadatajson. That is what is currently causing the failure.

lorenpaulsen:
@singhkay How do I fix this CI error? This template was originally intended to run against an existing VM, so I'm not sure what VM name to use here.

singhkay:
@lorenpaulsen Ah got it. In that case changing the metadata file will do. Thanks!

PR #1638, opened 2016-04-16
Auto-upgrade CustomScript for Cloudera.

boumenot:


azurecla:
Hi __@boumenot__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <p>
        It looks like you're working at Microsoft (chrboum). If you're full-time, we DON'T require a contribution license agreement.
    </p>
    <p>
    </p>

TTYL, AZPRBOT;

singhkay:
@netwmr01 As the template author, can you review this and make sure this is OK?

netwmr01:
What is this autoUpgradeMinorVersion? Base on what I read online
**The typeHandlerVersion specifies the version of the extension you would like to use. Setting autoUpgradeMinorVersion minor version to true ensures that you will get the latest Minor version of the extension that is available. It is highly recommended that you always set autoUpgradeMinorVersion to always be true.**

It seems reasonable. 
I just have a few questions:
What is the level of compatibility between minor version? If compatibility issues arise, will extension provider, Microsoft in this case, investigate first?
@singhkay

boumenot:
I have just taken over the Custom Script for Linux extension, so I'm on the hook to make sure I don't break things.  We have a great corpus of templates, and I would like to use them to validate the behavior of the Custom Scrpt extension, and ensure I don't *ever* break them.

We try to follow the guidelines outlined in Semantic Versioning when bumping the major, minor, or patch version.  If a compatibility issue arises then we broken Semantic versioning, and we need to fix it.

netwmr01:
@woodychristy What do you think?

netwmr01:
chatted with Woody offline, we are OK with this change.

singhkay:
@netwmr01 Thank you for confirming! merged

PR #1656, opened 2016-04-16
Location fix

marcvaneijk:


201-dependency-between-scripts-using-extensions/azuredeploy.json
```diff
@@ -212,7 +205,7 @@
         },
         "parameters": {
           "location": {
```
singhkay: Need to remove this "location" parameter from the nested template deployment. This is causing a failure.

PR #1664, opened 2016-04-16
Location fix

marcvaneijk:


singhkay:
closing as data lake has a dependency on the location

PR #1701, opened 2016-04-17
Location fix

marcvaneijk:


singhkay:
I think components is only available at that location

PR #1711, opened 2016-04-17
Location fix

marcvaneijk:


singhkay:
This seems to require the location in the hardcoded location

PR #1712, opened 2016-04-17
Location fix

marcvaneijk:
Deleted location variable from deployment template and nested templates.

### DIFF
--- a/active-directory-new-domain/azuredeploy.json
+++ b/active-directory-new-domain/azuredeploy.json
@@ -138,7 +138,6 @@
     }
   },
   "variables": {
-    "location": "[resourceGroup().location]",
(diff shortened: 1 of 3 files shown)


singhkay:
@marcvaneijk There might be a location parameter left over 

```
Deployment template validation failed: 'The value for the template parameter 'location' at line '5' and column '18' is not provided. Please see http://aka.ms/arm-deploy/#parameter-file for usage details.'.\nInvalidTemplate : Deployment template validation failed: 'The value for the template parameter 'location' at line '5' and column '18' is not provided. Please see http://aka.ms/arm-deploy/#parameter-file for usage details.'.\n"
```

marcvaneijk:
Interesting, CI is failing on the nested templates that I updated in the same PR. I rechecked the deployment template, the param file and the nested templates none have a location param anymore.

PR #1720, opened 2016-04-17
Location fix

marcvaneijk:


singhkay:
@ravbhatnagar I think this might be another case where CLI needs to return more details on the error

https://travis-ci.org/Azure/azure-quickstart-templates/builds/123764836#L218

PR #1724, opened 2016-04-17
Location fix

marcvaneijk:
Deleted location variable from deployment template.

### DIFF
--- a/bosh-setup/azuredeploy.json
+++ b/bosh-setup/azuredeploy.json
@@ -56,7 +56,6 @@
   },
   "variables": {
     "apiVersion": "2015-06-15",
-    "location": "[resourceGroup().location]",
 
     "defaultStorageAccountName": "[concat(uniqueString(resourceGroup().id), 'cfdefaultsa')]",
     "storageAccountType": "Standard_RAGRS",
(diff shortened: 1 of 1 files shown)


bosh-setup/azuredeploy.json
```diff
@@ -233,7 +232,7 @@
       "apiVersion": "[variables('apiVersion')]",
       "type": "Microsoft.Network/publicIPAddresses",
       "name": "[concat(parameters('vmName'),'-devbox')]",
-      "location": "[variables('location')]",
+      "location": "resourceGroup().location]",
```
singhkay: Missing '[' here

marcvaneijk:
Added missing bracket

bingosummer:
@marcvaneijk 

Could you please explain the reason the location should be fixed? Thanks.

marcvaneijk:
@bingosummer we're standardizing the templates with a base set from the recently updated guidelines https://github.com/Azure/azure-quickstart-templates/blob/master/1-CONTRIBUTION-GUIDE/best-practices.md

See comment #1727

bingosummer:
@marcvaneijk 

However, in `bosh-setup`, we set a variable to `resourceGroup().location` and use the variable to specify the location, which should follow the best practices.

It makes more sense for me because I don't need to use `resourceGroup().location` in each resource.

marcvaneijk:
@bingosummer yes, that is fine too. I'm just standardizing. You can deviate of course. As long as fixed values or params containing fixed values for the location are not used, unless there is a good reason for it (e.g. resource only available in one region). VS also sets the `resourceGroup().location` on resources added through the wizard.

bingosummer:
Got it. Thanks @marcvaneijk

PR #1727, opened 2016-04-17
Location fix

marcvaneijk:
and parameter file.

### DIFF
--- a/cloudera-on-centos/azuredeploy.json
+++ b/cloudera-on-centos/azuredeploy.json
@@ -40,28 +40,6 @@
         "description": "Unique public DNS name where the VMs will be exposed"
       }
     },
-    "location": {
-      "type": "string",
-      "defaultValue": "westeurope",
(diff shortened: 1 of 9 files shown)


singhkay:
@netwmr01 Can you review this change and see if you need to have a different location for resources than the resource group in your template?

netwmr01:
we don't need to have different resources in different location as the parent resources group. That being said, why the change and why now?

singhkay:
@netwmr01 We recently created a "Best Practices" guide for templates https://github.com/Azure/azure-quickstart-templates/blob/master/1-CONTRIBUTION-GUIDE/best-practices.md

Now we're trying to make the existing templates consistent with those so that any future templates that start off from existing templates at least have a base of "best practices" in place.

netwmr01:
@singhkay did this change pass the Travis CI build?

singhkay:
@marcvaneijk Can you test this out locally?

PR #1744, opened 2016-04-19
Location fix

marcvaneijk:
Changed resource location from parameter and variable to
resourceGroup().location. Deleted location parameter from deployment
template, nested templates and parameter file.


marcvaneijk:
@singhkay same issue again Kay. The CI fails on nested template params that I deleted in the PR, but somehow the CI deployment still seems to reference the old values.

singhkay:
@marcvaneijk merge conflicts

PR #1746, opened 2016-04-19
Location fix

marcvaneijk:
Changed resource location from parameter and variable to
resourceGroup().location. Deleted location parameter from deployment
template, nested templates and parameter file.

### DIFF
--- a/create-hpc-cluster/availabilitySets.json
+++ b/create-hpc-cluster/availabilitySets.json
@@ -8,12 +8,6 @@
         "description": "The API version"
       }
     },
-    "location": {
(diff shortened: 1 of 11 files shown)


singhkay:
@marcvaneijk Debugged this a bit and see there is a parmeter location at 
https://github.com/marcvaneijk/azure-quickstart-templates/blob/create-hpc-cluster/create-hpc-cluster/mainTemplate.json#L223

which is probably what is causing the failure

marcvaneijk:
Good catch @singhkay
Fixed!

PR #1772, opened 2016-04-21
Location fix

marcvaneijk:
Deleted location variable from deployment template and nested templates.


singhkay:
tested manually, passing

PR #1790, opened 2016-04-22
Add template for diagnostics with ELK

Skantha:
### Contributing guide

### Changelog
*
*
*
*
*

### DIFF
--- a/diagnostics-with-elk/azuredeploy.json
+++ b/diagnostics-with-elk/azuredeploy.json
@@ -123,17 +123,12 @@
             "type": "string",
             "defaultValue": "14.04.2-LTS",
             "allowedValues": [
-                "12.04.2-LTS",
-                "12.04.3-LTS",
-                "12.04.4-LTS",
-                "12.04.5-LTS",
                 "12.10",
                 "14.04.2-LTS",
-                "14.10",
-                "15.04"
+                "16.04-LTS"
             ],
             "metadata": {
-                "description": "The Ubuntu version for the VM. This will pick a fully patched image of this given Ubuntu version. Allowed values: 12.04.2-LTS, 12.04.3-LTS, 12.04.4-LTS, 12.04.5-LTS, 12.10, 14.04.2-LTS, 14.10, 15.04."
+                "description": "The Ubuntu version for the VM. This will pick a fully patched image of this given Ubuntu version."
             }
         },
         "existingDiagnosticsStorageAccountName": {
@@ -141,13 +136,6 @@
             "defaultValue": "",
             "metadata": {
                 "description": "Exiting diagnostics storage account name."
-            }
-        },
-        "existingDiagnosticsStorageAccountKey": {
-            "type": "securestring",
-            "defaultValue": "",
-            "metadata": {
-                "description": "Key for existing diagnostics storage account."
             }
         },
         "existingDiagnosticsStorageTableNames": {
@@ -169,14 +157,20 @@
         "storageAccountType": "Standard_LRS",
         "publicIPAddressType": "Dynamic",
         "vmStorageAccountContainerName": "vhds",
+        "storageAccountResourceGroup": "[resourceGroup().name]",
+        "storageAccountID": "[concat('/subscriptions/',subscription().subscriptionId,'/resourceGroups/',variables('storageAccountResourceGroup'),'/providers/','Microsoft.Storage/storageAccounts/', parameters('existingDiagnosticsStorageAccountName'))]",
+        "storageAccountKey": "[listkeys(variables('storageAccountID'), '2015-06-15').key1]",
         "vmName": "logstashvm1",
         "vmSize": "Standard_D1",
         "virtualNetworkName": "elkvnet2",
         "vnetID": "[resourceId('Microsoft.Network/virtualNetworks',variables('virtualNetworkName'))]",
         "subnetRef": "[concat(resourceId('Microsoft.Network/virtualNetworks', variables('virtualNetworkName')), '/subnets/', 'other')]",
-        "esHost":  "10.0.2.100"
+        "esHost":  "10.0.2.100",
+        "esTemplate":  "https://raw.githubusercontent.com/Azure/azure-quickstart-templates/master/elasticsearch/azuredeploy.json",
+        "esTemplateBase": "https://raw.githubusercontent.com/Azure/azure-quickstart-templates/master"
     },
     "resources": [
+
         {
           "apiVersion": "2015-01-01",
           "name": "Elasticsearch",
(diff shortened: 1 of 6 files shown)


azurecla:
Hi __@Skantha__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <p>
        It looks like you're working at Microsoft (skkandia). If you're full-time, we DON'T require a contribution license agreement.
    </p>
    <p>
    </p>

TTYL, AZPRBOT;

diagnostics-with-elk/azuredeploy.json
```diff
@@ -0,0 +1,367 @@
...
+        },
+        "adminPassword": {
+            "type": "securestring",
+            "metadata": {
+                "description": "Password for the Virtual Machine."
+            }
+        },
+        "ubuntuOSVersion": {
+            "type": "string",
+            "defaultValue": "14.04.2-LTS",
```
singhkay: Remove 12.04 LTS and add 16.04 LTS. Also remove 14.10 and 15.04 since these are removed from Azure.

diagnostics-with-elk/azuredeploy.json
```diff
@@ -0,0 +1,367 @@
...
+              "marvel": { "value": "no" },
+              "marvelCluster": { "value": "no" },
+              "vmSizeMarvelNodes": { "value": "Standard_D2_v2" },
+              "kibana": { "value": "yes" },
+              "sense": { "value": "no" },
+              "jmeterAgent": { "value": "no" },
+              "cloudAzure": { "value": "no" },
+              "cloudAzureStorageAccount": { "value": "" },
+              "cloudAzureStorageKey": { "value": "" },
+              "templateBase": { "value": "https://raw.githubusercontent.com/Azure/azure-quickstart-templates/master" }
```
singhkay: Possibly make this a variable so it's easier to change all URLs for testing from your own repo.

diagnostics-with-elk/azuredeploy.json
```diff
@@ -0,0 +1,367 @@
...
+                        }
+                    }
+                ]
+            }
+        },
+        
+        {
+            "type": "Microsoft.Storage/storageAccounts",
+            "name": "[variables('storageAccountName')]",
+            "apiVersion": "2015-05-01-preview",
```
singhkay: apiversion should be 2015-06-15

diagnostics-with-elk/azuredeploy.json
```diff
@@ -0,0 +1,367 @@
...
+            "location": "[variables('location')]",
+            "dependsOn": [
+                "[concat('Microsoft.Compute/virtualMachines/', variables('vmName'))]",
+                "[concat('Microsoft.Resources/deployments/', 'Elasticsearch')]"
+            ],
+            "properties": {
+                "publisher": "Microsoft.OSTCExtensions",
+                "type": "CustomScriptForLinux",
+                "typeHandlerVersion": "1.2",
```
singhkay: Use v1.5 and add the autoupgrademinorversion property.

diagnostics-with-elk/azuredeploy.json
```diff
@@ -0,0 +1,367 @@
...
+            }
+        },
+        "existingDiagnosticsStorageAccountName": {
+            "type": "string",
+            "defaultValue": "",
+            "metadata": {
+                "description": "Exiting diagnostics storage account name."
+            }
+        },
+        "existingDiagnosticsStorageAccountKey": {
```
singhkay: You can get the key using listKeys function which is more convenient for the user

diagnostics-with-elk/azuredeploy.json
```diff
@@ -0,0 +1,367 @@
...
+                "[concat('Microsoft.Compute/virtualMachines/', variables('vmName'))]",
+                "[concat('Microsoft.Resources/deployments/', 'Elasticsearch')]"
+            ],
+            "properties": {
+                "publisher": "Microsoft.OSTCExtensions",
+                "type": "CustomScriptForLinux",
+                "typeHandlerVersion": "1.2",
+                "settings": {
+                    "fileUris": [ "https://raw.githubusercontent.com/Azure/azure-quickstart-templates/master/diagnostics-with-elk/scripts/elk-simple-install-ubuntu.sh" ],
+                    "commandToExecute": "[concat('bash ./elk-simple-install-ubuntu.sh -e ', parameters('encodedConfigString'), ' -a ', parameters('existingDiagnosticsStorageAccountName'), ' -k ', parameters('existingDiagnosticsStorageAccountKey'), ' -t ', parameters('existingDiagnosticsStorageTableNames'), ' -i ', variables('esHost'))]"
```
singhkay: Since you're passing in the storage account key, use the protected settings.

diagnostics-with-elk/azuredeploy.json
Skantha: This existing storage account will be under a different resource group than the one used for resources in this template. So cannot use listKeys here.

singhkay: It'll work. You need to use the resourceId function and specify the resource group name. https://azure.microsoft.com/en-us/documentation/articles/resource-group-template-functions/#resourceid

Skantha: The storage account could also be under a different subscription so that may not work. We just need the storage account name/key to pull data from it. I put the key under protectedSettings abd that should prevent it from being visible to others, correct?

diagnostics-with-elk/azuredeploy.json
```diff
@@ -123,17 +123,12 @@
             "type": "string",
             "defaultValue": "14.04.2-LTS",
             "allowedValues": [
-                "12.04.2-LTS",
-                "12.04.3-LTS",
-                "12.04.4-LTS",
-                "12.04.5-LTS",
                 "12.10",
```
singhkay: Don't think the 12.10 image exists in Azure either since it isn't LTS

Skantha: Is there a way to determine what is supported? On the Azure portal I still see 12.04.5 LTS, 15.10. I just need to list the Azure supported versions there. Maybe there should be a ARM template helper function?

singhkay: Portal should be good indicator. Other way is to get through PS or CLI. 15.10 will probably get removed soon since its EOL from Canonical

Skantha: I will add 12.04.5-LTS back as its listed on the Portal, so my allowed list will look like this:
"allowedValues": [
                "12.04.5-LTS",
                "14.04.2-LTS",
                "16.04-LTS"
            ]
Let me know if that sounds good?

BTW I did a quick search on the existing templates and there are many that reference

singhkay: Sounds good. Yep, we need to update the templates to remove non LTS releases.

singhkay:
lgtm, merged

PR #1808, opened 2016-04-26
Location fix

marcvaneijk:


singhkay:
@vinhub There is an error in this template deployment. Can you take a look? 

`"error": "Deployment provisioning state was not successful\nUnable to process template language expressions for resource '/subscriptions/f42b4319-067a-4548-a650-33a1553b3a42/resourceGroups/qstci-49f5c6aa-ed64-8f9f-71a1-00b1d15116fe/providers/Microsoft.Compute/virtualMachines/ci4f8042VMdb/extensions/newuserscript' at line '1' and column '15453'. 'Template language expression property 'key1' doesn't exist, available properties are 'keys'. Please see http://aka.ms/arm-template-expressions for usage details.'\n",`

PR #1836, opened 2016-04-27
Location fix

marcvaneijk:


singhkay:
@marcvaneijk error `AssertionError: openjdk-tomcat-ubuntu-vm/metadata.json - Contains possible HTML elements which are not allowed`

marcvaneijk:
@singhkay removed HTML code

PR #1853, opened 2016-04-28
Location fix

marcvaneijk:
and parameter file.


singhkay:
@marcvaneijk Can you check the error? 

`AssertionError: redis-high-availability/azuredeploy.parameters.json -  Parameter "adminUsername" is missing its value field`

marcvaneijk:
@singhkay fixed param file

PR #1868, opened 2016-04-28
Location fix

marcvaneijk:
and parameter file.


singhkay:
@marcvaneijk  error ` AssertionError: spark-on-ubuntu/azuredeploy.parameters.json -  Parameter "storageAccountName" is missing its value field`

marcvaneijk:
@singhkay 
- moved storageAccountName to var
- replaced hardcoded endpoint with dynamic
- deleted storageAccountName param from param file
- Replaced fixed password with gen in param file

PR #1871, opened 2016-04-29
Location fix

marcvaneijk:
and parameter file.


singhkay:
@rarsan - We're making changes across the templates according to the new guidelines and eliminating the location parameter in favor of resource group location. 
https://github.com/Azure/azure-quickstart-templates/blob/master/1-CONTRIBUTION-GUIDE/best-practices.md

Can you review?

rarsan:
Given that eventually we'll be adding multi-site clustering which includes multiple regions, this change is not future-compatible: we would still need an explicit location parameter with one or multiple values, and cannot rely on resource group location.

singhkay:
@rarsan Makes sense in this case, thanks. I'll close this one out.

PR #1895, opened 2016-05-01
Add maxlength to app prefix

sedouard:


sedouard:
My editor seems to remove the extra spaces at the end. But that doesn't seem like a bad change.

Adding a maxLength of 11 on this property since its is concatenated with the result of a `uniqueString` call and used as a storage account name. This call returns string of length 13 and 24 characters is the max length of a storage account name.

singhkay:
@sedouard Since I discovered `?w=1` on github URLs whitespace changes on PRs don't scare me anymore :)  merged

PR #1905, opened 2016-05-03
Added template guacamole-existing-vnet

EgoAleSum:
Added a new ARM template to deploy a VM with Guacamole, the open source RDP/VNC proxy. This is useful in corporate environments where firewalls restrict connections to ports other than 80/443.

The deployment uses Docker containers running on a CoreOS host. It's expected that the VM is placed inside an existing VNet and existing subnet.


azurecla:
Hi __@EgoAleSum__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <p>
        It looks like you're working at Microsoft (asegala). If you're full-time, we DON'T require a contribution license agreement.
    </p>
    <p>
    </p>

TTYL, AZPRBOT;

singhkay:
@EgoAleSum Instead of requiring an existing a VNET, what if this template offered the choice to create a new VNET or use an existing VNET based on the new or existing VNET pattern? https://github.com/rjmax/ArmExamples/blob/master/NewOrExistingTemplate.json

EgoAleSum:
@singhkay I'll investigate that, thanks

EgoAleSum:
@singhkay I've been thinking about that, but I don't think it makes sense to even allow the option of deploying a new VNet.

I made this template for a customer of mine who is running an Azure PoC without VPN yet, and they can't connect to RDP because the corporate firewall is blocking it. As Guacamole is a web-based RDP/VNC proxy, it makes sense only if you deploy it in a VNet in which other VMs exist... My 2 cents.

I'm not sure how to fix the issues with the CI, however.

singhkay:
@EgoAleSum Makes sense. Let's rename the folder to be "guacamole-rdp-vnc-gateway" or something else that describes the scenario a bit more

guacamole-existing-vnet/azuredeploy.json
```diff
@@ -0,0 +1,236 @@
+{
+  "$schema": "https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#",
+  "contentVersion": "1.0.0.0",
+  "parameters": {
+    "location": {
```
singhkay: Use resourceGroup().location instead

guacamole-existing-vnet/azuredeploy.json
```diff
@@ -0,0 +1,236 @@
...
+      }
+    },
+    "vmSize": {
+      "type": "string",
+      "defaultValue": "Standard_A1",
+      "metadata": {
+        "description": "Instance size for the Guacamole VM"
+      }
+    },
+    "storageAccountNamePrefix": {
```
singhkay: See #944 for storage account name guidance

guacamole-existing-vnet/azuredeploy.json
EgoAleSum: For the location, I'd prefer the user to specify it because it has to be the same location of the existing VNet. However, if you still think resourceGroup().location is a better approach, i'll change it.

EgoAleSum:
@singhkay Thanks for the feedback.

I've renamed the folder and I've fixed the storage account name (hope it's fine now). As I've written in the response to the inline comment, for the location I'd prefer the user to specify it because it has to be the same location of the existing VNet. However, if you still think resourceGroup().location is a better approach, i'll change it.

guacamole-existing-vnet/azuredeploy.json
singhkay: Makes sense in this case. Btw will you be extending this template to make it existing or new VNET?

guacamole-rdp-vnc-gateway-existing-vnet/README.md
```diff
@@ -0,0 +1,58 @@
+# Guacamole VM in existing VNet
+
+This template deploys a VM with [Guacamole](http://guac-dev.org), the open source HTML5 RDP/VNC proxy.
+
+You will need an existing Virtual Network, and you will need the name of the VNet and a subnet in that VNet. This template deploys Guacamole and MariaDB using Docker containers, and it's based on CoreOS (channel "stable").
+
+## Template parameters
+
+| Parameter | Description | Default value |
```
singhkay: The table of parameters is not needed since it is rendered on Azure.com

guacamole-rdp-vnc-gateway-existing-vnet/azuredeploy.json
```diff
@@ -0,0 +1,229 @@
+{
+  "$schema": "https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#",
+  "contentVersion": "1.0.0.0",
+  "parameters": {
+    "location": {
```
singhkay: Since this is the location of existing VNET, then it's best to call this parameter "existingVNETLocation" to set the right expectation

guacamole-existing-vnet/azuredeploy.json
EgoAleSum: As per my comment above (https://github.com/Azure/azure-quickstart-templates/pull/1905#issuecomment-216701584) I'm not planning to, as it's a lot of work for something that's more like an edge case IMHO. But I'll gladly accept input from you and the community.

singhkay:
@EgoAleSum Left few more comments

EgoAleSum:
@singhkay Applied the changes, thanks

singhkay:
lgtm, merged

PR #1949, opened 2016-05-13
Update two DSC Module, now the CreateVirtualDisk will have a paramete…

weidaiMS:
Update two DSC Module, now the CreateVirtualDisk will have a paramete…


singhkay:
lgtm, merged

PR #1967, opened 2016-05-17
Add ports for ASCS and SCS

MSSedusch:
### Changelog
* add additional ports for ASCS/SCS
* increase timeout of load balancer to 30 minutes

### Description of the change
Added additional ports to support deployment for ASCS + SCS use case. Also increased the timeout of the ASCS/SCS load balancer to 30 minutes


MSSedusch:
@singhkay Can you please check Travis? Not sure what the issue is. I was able to validate and deploy the template.

singhkay:
merged

PR #1978, opened 2016-05-19
make mdadm raiding more resillient

Mpdreamz:
This is a PR with two changes that made our disk provisioning more resilient over at https://github.com/elastic/azure-marketplace

The first problem was that quite often `mdadm` was not being installed properly and the command was never available on the machine. We solved this by retrying `apt-get update` and installing mdadm under sudo.

The other thing we saw was that quite often we'd get `resource busy` messages from `mdadm` which caused `mkfs.ext4` to fail and later on in the script the call to `blkid` to not return a UUID for `/dev/md0` which in turn made the call to `add_to_fstab` to be called with no UUID.

In short this caused `/datadisks/disk1` to point to the local `sda` partition `30gb` instead of the attached raided disks (2TB in this particular case).

The cause for this seems to be a race condition and the fix documented here:

http://dev.bizo.com/2012/07/mdadm-device-or-resource-busy.html

Has worked very well on our template.

### DIFF
--- a/shared_scripts/ubuntu/vm-disk-utils-0.1.sh
+++ b/shared_scripts/ubuntu/vm-disk-utils-0.1.sh
@@ -273,9 +273,10 @@ create_striped_volume()
 	done
 
     MDDEVICE=$(get_next_md_device)    
-    
+	sudo udevadm control --stop-exec-queue
 	mdadm --create ${MDDEVICE} --level 0 --raid-devices ${#PARTITIONS[@]} ${PARTITIONS[*]}
-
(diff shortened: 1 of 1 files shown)


azurecla:
Hi __@Mpdreamz__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
    </span>

TTYL, AZPRBOT;

singhkay:
@Mpdreamz Thank you for the fix! merged

PR #2052, opened 2016-06-03
fix max agents limit for ACS

### LINKED ISSUE #2048: 101-acs-dcos: incorrect value for agentCount.maxValue
[101-acs-dcos](https://github.com/Azure/azure-quickstart-templates/tree/master/101-acs-dcos)

### Issue Details

All three `101-acs-*` templates specify `agentCount.maxValue`. Two of them (`101-acs-mesos` and `101-acs-swarm`) have 100. The `101-acs-dcos` has 40. The documentation in both `101-acs-dcos` and `101-acs-mesos` states `This value can be from 1 to 40` which is also inconsistent with the actual value in `101-acs-mesos`.

I suspect the correct value in all cases should be 100.

rgardler:
### Changelog
* Fix max agent limit in mesos and dcos templates (fixes #2048 )

### DIFF
--- a/101-acs-dcos/azuredeploy.json
+++ b/101-acs-dcos/azuredeploy.json
@@ -12,10 +12,10 @@
       "type": "int",
       "defaultValue": 1,
       "metadata": {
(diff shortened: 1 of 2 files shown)


singhkay:
lgtm, merged

PR #2055, opened 2016-06-04
strong versioning for endpoint

marcvaneijk:
Removed "latest" from reference examples for public endpoint and replaced with strong versioning examples


singhkay:
lgtm, merged

PR #2104, opened 2016-06-14
"osdisk" VHD being overwritten.

t-anjan:
When creating multiple VMs using "copy()", the osdisk name should also be indexed.

If this is not done, then the same OS disk is overwritten for subsequent VMs.


azurecla:
Hi __@t-anjan__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
    </span>

TTYL, AZPRBOT;

singhkay:
lgtm, merged

PR #2112, opened 2016-06-16
This template is to provide HA for RDS gateway servers. So changing the name accordingly.

navesh123:
This template is to provide HA for RDS gateway servers. So changing the
name accordingly.


singhkay:
lgtm, merged

PR #2116, opened 2016-06-16
Remove sql-server-2014-alwayson-dsc solution

simongdavies:
### Changelog

* Remove sql-server-2014-alwayson-dsc folder and solution

### Description of the change

This change removes the sql-server-2014-alwayson-dsc solution as there is now an official solution that is created and maintained by the SQL Server PG that supports SQL Server 2012,2014 and 2016 and other capabilities not supported by this solution


singhkay:
merged

PR #2125, opened 2016-06-17
Fix the args passed to resourceId()

philon-msft:
Some resource name parts weren't being concatenated together before being passed to resourceId()


singhkay:
@philon-msft lgtm, thanks for fixing!

PR #2150, opened 2016-06-21
Adjust defaults for _artifactsLocation and _artifactsLocationSasToken

philon-msft:
Minor fixes for publishing the template via the Portal, where the default values for _artifactsLocation* are used to retrieve scripts and nested templates directly from https://raw.githubusercontent.com rather than staging them in a storage account.

### DIFF
--- a/elasticsearch/azuredeploy.json
+++ b/elasticsearch/azuredeploy.json
@@ -295,14 +295,14 @@
     },
     "_artifactsLocation": {
       "type": "string",
-      "defaultValue": "https://raw.githubusercontent.com/Azure/azure-quickstart-templates/master/elasticsearch/",
+      "defaultValue": "https://raw.githubusercontent.com/Azure/azure-quickstart-templates/master/elasticsearch",
       "metadata": {
         "description": "Change this value to your repo name if deploying from a fork"
       }
     },
     "_artifactsLocationSasToken": {
       "type": "securestring",
-      "defaultValue": "?",
+      "defaultValue": "",
       "metadata": {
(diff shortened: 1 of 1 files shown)


elasticsearch/azuredeploy.json
```diff
@@ -295,23 +295,22 @@
...
-      "defaultValue": "https://raw.githubusercontent.com/Azure/azure-quickstart-templates/master/elasticsearch/",
+      "defaultValue": "https://raw.githubusercontent.com/Azure/azure-quickstart-templates/master/elasticsearch",
       "metadata": {
         "description": "Change this value to your repo name if deploying from a fork"
       }
     },
     "_artifactsLocationSasToken": {
       "type": "securestring",
-      "defaultValue": "?",
+      "defaultValue": "",
```
singhkay: @philon-msft I thought we're doing "?" as default for sastoken parameter?

philon-msft: @singhkay, unfortunately the question mark causes problems. For example: Windows CSE tries to download the target to a local file named "filename.ps1?" and the question mark is an invalid filename character. 
Was there a reason to use "?" rather than ""?

singhkay: I asked the same thing by was told that ? should be ok since there are no query parameters after it. If this is causing a problem then we need to change all existing templates where this change was made and also the guidance docs.

cc: @bmoore-msft

philon-msft: It looks like 201-vm-custom-script-windows is the only other template using "?"

PR #2234, opened 2016-07-11
Open edX scalable (multi-VM) template

chenriksson:
Contributing guide

Description of the change

Open edX Dogwood scalable (multi-VM) ARM template. Originally created by Appsembler for Microsoft, now migrated to Dogwood release.

### DIFF
--- a/openedx-scalable-ubuntu/azuredeploy.json
+++ b/openedx-scalable-ubuntu/azuredeploy.json
@@ -148,7 +148,7 @@
   "variables": {
     "apiVersion": "2015-06-15",
     "openedxVersion": "named-release/dogwood.3",
-    "scriptDownloadUri": "https://raw.githubusercontent.com/azure/azure-quickstart-templates/master/openedx-scalable-ubuntu/",
+    "scriptDownloadUri": "https://raw.githubusercontent.com/Azure/azure-quickstart-templates/master/openedx-scalable-ubuntu/",
     "installScript": "install-openedx.sh",
     "installCommand": "[concat('bash -c ''nohup ./', variables('installScript'), ' ', variables('openedxVersion'), ' ', parameters('appVmCount'), ' ', parameters('adminUsername'), ' ', parameters('adminPassword'), ' </dev/null &>/var/log/azure/openedx-install.log &''')]",
     "availabilitySetName": "avail-set",
--- a/openedx-scalable-ubuntu/azuredeploy.json
+++ b/openedx-scalable-ubuntu/azuredeploy.json
@@ -1,0 +1,694 @@
+{
+  "$schema": "https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#",
+  "contentVersion": "1.0.0.0",
+  "parameters": {
+    "adminUsername": {
+      "type": "string",
+      "defaultValue": "openedxuser",
+      "metadata": {
+        "description": "Administrator username."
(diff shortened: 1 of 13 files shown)


azurecla:
Hi __@chenriksson__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
    </span>

TTYL, AZPRBOT;

openedx-scalable-ubuntu/azuredeploy.json
```diff
@@ -0,0 +1,694 @@
...
+        ]
+      }
+    },
+    {
+      "apiVersion": "[variables('apiVersion')]",
+      "type": "Microsoft.Network/networkSecurityGroups",
+      "name": "[variables('nsgName')]",
+      "location": "[resourceGroup().location]",
+      "properties": {
+        "securityRules": [
```
gatneil: If there are no deny rules, every port is open, so these allow rules aren't actually changing anything; I recommend either removing them to simplify the template or adding a deny all else rule so they actually do something :)

openedx-scalable-ubuntu/azuredeploy.json
```diff
@@ -0,0 +1,694 @@
...
+        "[concat('Microsoft.Network/loadBalancers/',variables('lbName'))]",
+        "[concat('Microsoft.Network/loadBalancers/',variables('lbName'), '/inboundNatRules/SSH-VM', copyIndex())]"
+      ],
+      "properties": {
+        "ipConfigurations": [
+          {
+            "name": "ipConfigNode",
+            "properties": {
+              "privateIPAllocationMethod": "Static",
+              "privateIPAddress": "[concat('10.0.0.1',copyIndex())]",
```
gatneil: 2 things:

1) I believe that 10.0.0.0-10.0.0-3 (inclusive) get reserved by the vnet, so I doubt you can use them as intended here; if you still want to use static IPs, you should use a different range.

2) Do you really need static IPs here? Could you use dynamic IPs with an internal/external load balancer with NAT rules instead? Or perhaps dynamic IPs and use the within-vnet DNS to reference VMs by their name? This would remove the limit of 9 VMs and perhaps be a bit less confusing.

chenriksson:
Sorry for the long delay on this PR!!!! Installation for these templates broke a couple weeks ago due to an update from a python module dependency. This should be fixed in my next push.

openedx-scalable-ubuntu/azuredeploy.json
chenriksson: I've found the ports are blocked if we don't explicitly open them. In the portal under the NSG inbound security rules, if I toggle 'Default rules' I can see a default 'DenyAllInBound' rule was created.

This seems consistent with this article: https://azure.microsoft.com/en-us/documentation/articles/best-practices-resource-manager-security

openedx-scalable-ubuntu/azuredeploy.json
chenriksson: The static IP range is 10.0.0.10-10.0.0.19, and 10.0.0.20/10.0.0.30 for MySQL/Mongo VMs.

I'll look into whether I can easily use dynamic IPs instead.

chenriksson:
Retracting for now. Will issue PR for edx fullstack fix shortly, and will resubmit scalable separately.

PR #2264, opened 2016-07-21
Rename template to prevent disappearance from Azure index

faiz-bloq:
After a few merges it seems our template has disappeared from the index again, so to correct that I have tried changing the name to be at least as long as some of the other templates, in case that was triggering its disappearance.


azurecla:
Hi __@faiz-bloq__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!

TTYL, AZPRBOT;

singhkay:
lgtm, merged

PR #2284, opened 2016-07-28
updated to SQL Server 2014 SP1 Image

simongdavies:
### Changelog
* Replaced SQL Server 2014 Image with SP1 Version
* Moved SQL Server DSC extensions to WMF 5.0

### Description of the change

This change updated the version of SQL Server 2014 used by the templates to 2014 SP1 and also uses WMF5 rather than WMF4 for DSC extension execution on the SQL Server VMs


singhkay:
lgtm, merged

Issue #934, opened 2015-12-09
[HELP WANTED] Make apiVersion properties a variable across templates


singhkays:
Currently we have most templates that hardcode an `apiVersion` for the resource. For example 

    "apiVersion": "2015-08-01",

You can help us by finding templates that do this and modifying it to use a variable. Here's how to do this

* Define a `variable apiVersion` in the variables section. Potentially different resources could use different apiVersions. In that case, you'll need to define multiple apiVersion variables.

          "variables": {   
             "apiVersion": "2015-08-01"
          },

* Then use this variable for the `apiVersion` property of the resource

        "apiVersion": "[variables('apiVersion')]",

When submitting the PR, please deploy the template to make sure there are no errors. If you run into errors, please file a Github issue against the template and we can investigate.

boutquin:
Happy to work on this; started with a single template #1087 . Validation still failing though the deployment using the template succeeds. Will also incorporate #944 and #966 when touching a template.

bmoore-msft:
We are going to move to using literal string values for apiVersion so that JSON schema validation can be performed on the templates.

Issue #2296, opened 2016-07-28
concourse-ci: invalid values in CustomScriptForLinux section


ahmetalpbalkan:
`concourse-ci/azuredeploy.json` has invalid keys under its CustomScriptForLinux's `
"properties"` section:

```json
    {
      "type": "Microsoft.Compute/virtualMachines/extensions",
      "name": "[concat(parameters('vmName'),'/initdevbox')]",
      "apiVersion": "[variables('apiVersion')]",
      "location": "[resourceGroup().location]",
      "dependsOn": [
        "[concat('Microsoft.Compute/virtualMachines/', parameters('vmName'))]"
      ],
      "properties": {
        "publisher": "Microsoft.OSTCExtensions",
        "type": "CustomScriptForLinux",
        "typeHandlerVersion": "1.2",
        "settings": {
          "fileUris": [
            "[concat(variables('githubPath'),'launch_script.sh')]"
          ],
          "commandToExecute": "[concat('sh launch_script.sh',' ',variables('githubPath'))]",
          "VNET-NAME": "[variables('virtualNetworkName')]",
          "SUBNET-NAME": "[variables('subnetNameForBosh')]",
          "SUBNET-NAME-FOR-CONCOURSE": "[variables('subnetNameForConcourse')]",
          "SUBSCRIPTION-ID": "[subscription().subscriptionId]",
          "STORAGE-ACCOUNT-NAME": "[variables('newStorageAccountName')]",
          "STORAGE-ACCESS-KEY": "[listKeys(variables('storageid'),variables('apiVersion')).key1]",
          "RESOURCE-GROUP-NAME": "[resourceGroup().name]",
          "TENANT-ID": "[parameters('tenantID')]",
          "CLIENT-ID": "[parameters('clientID')]",
          "CLIENT-SECRET": "[parameters('clientSecret')]",
          "CONCOURSE-IP": "[reference('concourse').ipAddress]",
          "username": "[parameters('adminUserName')]"
        }
      }
    }

```

These unfamiliar keys (everything except `commandToExecute` and `fileUris`)  will break with the new version of CustomScriptForLinux extension which does strict json-schema validation. 

cc: @bingosummer please fix

ahmetalpbalkan:
same for `bosh-setup/azuredeploy.json`.

bingosummer:
Yes. WIll fix.
I will fix `bosh-setup`, and @halimacc will fix `concource-ci`.

[...]

PR #2327, opened 2016-08-08
Using CustomData to pass information to CustomScript

bingosummer:
### Contributing guide

### Changelog
* Use CustomData to pass information to CustomScript

### Description of the change

https://github.com/Azure/azure-quickstart-templates/issues/2296#issuecomment-236046332

### DIFF
--- a/bosh-setup/azuredeploy.json
+++ b/bosh-setup/azuredeploy.json
@@ -65,9 +65,6 @@
     }
   },
   "variables": {
-    "apiVersion": "2015-06-15",
-    "apiVersionForDeployment": "2015-01-01",
-
     "defaultStorageAccountName": "[concat(uniqueString(resourceGroup().id), 'cfdefaultsa')]",
     "storageAccountType": "Standard_RAGRS",
     "vmStorageAccountContainerName": "vhds",
@@ -168,7 +165,7 @@
   },
   "resources": [
     {
-      "apiVersion": "[variables('apiVersion')]",
+      "apiVersion": "2015-06-15",
       "type": "Microsoft.Network/networkSecurityGroups",
       "name": "[variables('boshNetworkSecurityGroup')]",
       "location": "[resourceGroup().location]",
@@ -234,7 +231,7 @@
       }
     },
     {
-      "apiVersion": "[variables('apiVersion')]",
+      "apiVersion": "2015-06-15",
       "type": "Microsoft.Network/networkSecurityGroups",
       "name": "[variables('devboxNetworkSecurityGroup')]",
       "location": "[resourceGroup().location]",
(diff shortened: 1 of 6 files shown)


bosh-setup/azuredeploy.json
```diff
@@ -66,6 +66,7 @@
   },
   "variables": {
     "apiVersion": "2015-06-15",
```
bmoore-msft: remove these and hard-code the apiVersion in the resources - otherwise json editors cannot provide intellisense.

bosh-setup/azuredeploy.json
```diff
@@ -457,7 +515,7 @@
           "osDisk": {
             "name": "osdisk",
             "vhd": {
-              "uri": "[concat('http://',variables('defaultStorageAccountName'),'.blob.',variables('environment').serviceHostBase,'/',variables('vmStorageAccountContainerName'),'/',parameters('vmName'),'-osdisk.vhd')]"
+              "uri": "[concat('http://', variables('defaultStorageAccountName'), '.blob.', variables('environment').serviceHostBase, '/', variables('vmStorageAccountContainerName'), '/', parameters('vmName'), '-osdisk.vhd')]"
```
bmoore-msft: Any reason we can't follow the best practices for uri's here: https://github.com/Azure/azure-quickstart-templates/blob/master/1-CONTRIBUTION-GUIDE/best-practices.md  (see #9)

bingosummer: Thanks @bmoore-msft .
I didn't know the best practice for the vhd uri. I think, we can follow the best practice.

bosh-setup/azuredeploy.json
bingosummer: I think it's a good way to define a variable `apiVersion`, and use it everywhere.
What does the intellisense of json editors mean? Which json editor?
Which one is the best practice? Defining a variable `apiVersion` or hard-coding the apiVersion in the resources?

bmoore-msft: BP for the repo is to use literal string value for the apiVersion.

It is a good practice to use a var in cases like this but unfortunately the way the json schemas are designed the document can't be evaluated if an expression is used.  It affects all JSON schema validations (editors, rest api's, etc).

bingosummer: I am following https://github.com/Azure/azure-quickstart-templates/issues/934, and it works well. The template can pass the template validation, and can be deployed successfully. So, I prefer to keep apiVersion as a var.

bmoore-msft: Thanks for catching that - I've closed #934 as we are going to use literal apiVersions for the samples in this repo.  So you can update as such in accordance with the BP.

bingosummer: Got it. Thanks.
Will remove the var `apiVersion`, and specify `apiVersion` in each resource.

bosh-setup/azuredeploy.json
```diff
@@ -66,6 +66,7 @@
   },
   "variables": {
     "apiVersion": "2015-06-15",
+    "apiVersionForDeployment": "2015-01-01",
```
bingosummer: I believe that I should also remove `apiVersionForDeployment`. Right?

bmoore-msft: Yes, please...

PR #2332, opened 2016-08-10
Updated Storage Account

pcgeek86:
### Contributing guide

### Changelog
* Updated Storage Account definition for idempotent use

### Description of the change
This fixes a problem where Storage Accounts were not treated in an idempotent fashion. Updating to 2016-01-01 for the API version, just for the Storage Account, will resolve this problem. However, the schema for the Storage Account changed, so we must add the `sku` and `kind` properties at the root of the Storage Account definition.

### DIFF
--- a/101-vm-simple-linux/azuredeploy.json
+++ b/101-vm-simple-linux/azuredeploy.json
@@ -61,7 +61,9 @@
       "name": "[variables('storageAccountName')]",
       "apiVersion": "2016-01-01",
       "location": "[resourceGroup().location]",
-      "sku": "[variables('storageAccountType')]",
+      "sku": {
+        "name": "[variables('storageAccountType')]"
+      },
       "kind": "Storage",
       "properties": {
       }
--- a/101-vm-simple-linux/azuredeploy.json
+++ b/101-vm-simple-linux/azuredeploy.json
@@ -59,10 +59,11 @@
     {
       "type": "Microsoft.Storage/storageAccounts",
       "name": "[variables('storageAccountName')]",
-      "apiVersion": "[variables('apiVersion')]",
+      "apiVersion": "2016-01-01",
       "location": "[resourceGroup().location]",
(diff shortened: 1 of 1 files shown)


azurecla:
Hi __@pcgeek86__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
        <p>The agreement was validated by Microsoft and real humans are currently evaluating your PR.</p>

TTYL, AZPRBOT;

singhkays:
> This fixes a problem where Storage Accounts were not treated in an idempotent fashion. Updating to 2016-01-01 for the API version, just for the Storage Account, will resolve this problem.

@pcgeek86 Can you elaborate on this? My understading is that all templates in ARM should be idempotent by default. 

FYI folks who might be interested - @rjmax @gatneil @bmoore-msft 

#### Review comments
The CI is failing with `"error": "error: InvalidTemplate : Deployment template parse failed: 'Error converting value \"Standard_LRS\" to type 'Microsoft.WindowsAzure.ResourceStack.Frontdoor.Data.Entities.Resources.ResourceSku'. Path ''.'.\n"`. This is because the SKU property needs to look like below
```
    "sku": {
        "name": "Standard_LRS|Standard_ZRS|Standard_GRS|Standard_RAGRS|Premium_LRS"
    }, 
```

pcgeek86:
@singhkays Sure, the issue is that the `Microsoft.Storage` Resource Provider had an issue where the older API version of `2015-06-15` would complain if a Storage Account already existed. This issue broke the idempotence of all ARM templates that declare a Storage Account resource. The newer API version `2016-01-01` for the `Microsoft.Storage` Resource Provider fixes this issue. 

There was a pretty lengthy conversation about this on the Azure MVP mailing list.

pcgeek86:
**Update**: Fixed the issue with the `sku` property.

gbowerman:
@pcgeek86 merged. Thanks.

Issue #2340, opened 2016-08-11
Flighting the new CustomScript extension for Linux


ahmetalpbalkan:
We're rolling out a rewrite of the `Microsoft.OSTCExtensions.CustomScriptForLinux` extension. Pertinent to internal issue, opening an issue to track progress of the rollout in this repository.

# Summary

Notable changes in the new extension
-         Written in Go. Has no dependencies to the waagent or Python.
   - means increased success rate on distros which does not have Python in PATH
-         Strict configuration validation (fails on unrecognized, mistyped extension settings etc)
    - so if the extension config was invalid before (or had extra keys etc) it will fail (I did a manual verification of templates using the extension and opened couple of issues about this)
-         Better retries around file downloads
-         Better error messages for many cases

# Migration

The migration is as simple as changing the:

- publisher to `Microsoft.Azure.Extensions`
- type to `CustomScript`
- version to `2.0`
- and we should add `"autoUpgradeMinorVersion": true` to the extension resources.

# Rollout status

I will be updating & shuffling the list below frequently to determine the slices to rollout and track progress etc. Please subscribe from the right sidebar. (list might not be complete)

### 🍏   Stage 1 (PR: #2344) (2016-08-12)

> **Status:** Complete
> 
> We did not hear any issues from these templates over the past 7 days of flighting
> ![image](https://cloud.githubusercontent.com/assets/159209/17866258/b4e49896-685a-11e6-898c-e7aafdf7bdb2.png)

- [x] [elasticsearch](https://github.com/Azure/azure-quickstart-templates/tree/master/elasticsearch)
- [x] [elasticsearch-centos-3node](https://github.com/Azure/azure-quickstart-templates/tree/master/elasticsearch-centos-3node)
- [x] [elasticsearch-jmeter](https://github.com/Azure/azure-quickstart-templates/tree/master/elasticsearch-jmeter)
- [x] [moodle-cluster-centos](https://github.com/Azure/azure-quickstart-templates/tree/master/moodle-cluster-centos) (template owner reported issues, possibly unrelated)
- [x] [moodle-cluster-ubuntu](https://github.com/Azure/azure-quickstart-templates/tree/master/moodle-cluster-ubuntu)
- [x] [moodle-singlevm-centos](https://github.com/Azure/azure-quickstart-templates/tree/master/moodle-singlevm-centos)
[...]

PR #2344, opened 2016-08-12
CustomScript rollout - stage #1

ahmetalpbalkan:
cc:  @bmoore-msft @gatneil PTAL
See #2340.

### DIFF
--- a/apache2-on-ubuntu-vm/azuredeploy.json
+++ b/apache2-on-ubuntu-vm/azuredeploy.json
@@ -178,9 +178,10 @@
         "[concat('Microsoft.Compute/virtualMachines/', variables('vmName'))]"
       ],
       "properties": {
-        "publisher": "Microsoft.OSTCExtensions",
-        "type": "CustomScriptForLinux",
-        "typeHandlerVersion": "1.2",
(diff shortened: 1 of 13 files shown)


ahmetalpbalkan:
@gatneil can you please take a look at test failures, they don't seem pertinent to the changes.

gatneil:
It looks like in some cases we have publisher "Microsoft.Azure.Extension" with type "CustomScript" and in other cases we have publisher "Microsoft.Compute" with type "CustomScriptExtension". Is this on purpose?

ahmetalpbalkan:
@gatneil Microsoft.Compute one is the Windows CustomScript extension. We don't touch that just yet.

We have plans to merge them all in this new extension. Unfortunately the organization names (such as Microsoft.Compute or Microsoft.OSTC) ended up in extension names, the new extension fixes that (see: [Conway's Law](https://en.wikipedia.org/wiki/Conway%27s_law)).

gatneil:
Cool. Otherwise, the PR looks good. You're correct that the CI errors are unrelated to your changes. The azuredeploy.parameters.json files should be updated on 2 of them so they utilize the CI properly, but I can do that in a follow-up PR :)

PR #2345, opened 2016-08-12
Added template for OMS

marcvaneijk:


marcvaneijk:
@bmoore-msft @gatneil, can you have a look at these two templates if they align with the best practices and maybe understand why they fail in the CI. Thanks, Marc //@singhkays

PR #2347, opened 2016-08-15
use lower case resource group name as container

MSSedusch:
### Changelog
* fix issue for user image templates when resource group name has upper case characters

### Description of the change
The deployment of the user image templates (2-tier and 3-tier) would fail prior to this fix if the resource group name contains an upper case character

### DIFF
--- a/sap-2-tier-marketplace-image/shared/os-disk-parts.json
+++ b/sap-2-tier-marketplace-image/shared/os-disk-parts.json
@@ -74,6 +74,7 @@
   "variables": {
     "osDiskName": "[concat(parameters('sidlower'),'-os.vhd')]",
     "osDiskCaching": "ReadWrite",
+    "resourceGroupLower":"[toLower(resourceGroup().name)]",
     "osDiskParts": {
       "image": {
         "imageReference": {
@@ -100,7 +101,7 @@
             "uri": "[parameters('userImageVhdUri')]"
           },
           "vhd": {
-            "uri": "[concat('http://',parameters('userImageStorageAccount'),'.blob.core.windows.net/', resourceGroup().name, '/', parameters('vmName'), '-', variables('osDiskName'))]"
+            "uri": "[concat('http://',parameters('userImageStorageAccount'),'.blob.core.windows.net/', variables('resourceGroupLower'), '/', parameters('vmName'), '-', variables('osDiskName'))]"
           },
           "caching": "[variables('osDiskCaching')]",
           "createOption": "FromImage"
--- a/sap-2-tier-marketplace-image/shared/os-disk-parts.json
+++ b/sap-2-tier-marketplace-image/shared/os-disk-parts.json
@@ -86,7 +86,7 @@
         "osDisk": {
           "name": "osdisk",
           "vhd": {
-            "uri": "[concat('http://', parameters('storageAccountName'), '.blob.core.windows.net/', parameters('vmName'), '/', variables('osDiskName'))]"
+            "uri": "[concat('http://', parameters('storageAccountName'), '.blob.core.windows.net/', concat('vhds-', uniqueString(parameters('vmName'))), '/', variables('osDiskName'))]"
           },
           "caching": "[variables('osDiskCaching')]",
           "createOption": "FromImage"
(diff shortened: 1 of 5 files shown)


sap-2-tier-marketplace-image/shared/os-disk-parts.json
```diff
@@ -100,7 +101,7 @@
             "uri": "[parameters('userImageVhdUri')]"
           },
           "vhd": {
-            "uri": "[concat('http://',parameters('userImageStorageAccount'),'.blob.core.windows.net/', resourceGroup().name, '/', parameters('vmName'), '-', variables('osDiskName'))]"
+            "uri": "[concat('http://',parameters('userImageStorageAccount'),'.blob.core.windows.net/', variables('resourceGroupLower'), '/', parameters('vmName'), '-', variables('osDiskName'))]"
```
bmoore-msft: resourceGroup names can contain other chars/sequences not allowed in storage container names - can we just remove this and hardcode a container name (e.g. /vhds/...) - seems like that will work fine with this sample.

MSSedusch: The storage account does not belong to the resource group so there could potentially already be a vhd container containing the same blobs. I would rather stick with the resource group name.

bmoore-msft: The diskname already contains a few components that are likely to be unique (e.g. SAP SID) so that should be sufficient.  If a container is really needed, use concat('vhds-', uniqueString(resourceGroup().id)) which will ensure naming requirements are met.

MSSedusch: ok, I changed the name as you suggested. Thanks

bmoore-msft: Need to make sure there are no underscores either - see: https://azure.microsoft.com/en-us/documentation/articles/guidance-naming-conventions/#naming-rules-and-restrictions

Which is why you don't see this pattern used for storage resources...

MSSedusch: not sure to which part of the template you are referring to. The storage account container should not contain underscores. The blog can be anything and the storage account name is the name of an actual storage account so the name should also be correct.

bmoore-msft: You're using toLower(resourceGroup().name) for the name of the storage container.  The resource group name can contain an underscore.  The container name cannot.  This template will fail to deploy when the resource group name contains an underscore.

MSSedusch: Can you please have a look at the file itself? I changed it like you suggested and the toLower resource group name is no longer used.
Thanks

bmoore-msft: Strange I was not seeing the latest commit - yes it does look good now.

sap-2-tier-marketplace-image/shared/os-disk-parts.json
```diff
@@ -85,7 +86,7 @@
         "osDisk": {
           "name": "osdisk",
           "vhd": {
-            "uri": "[concat('http://', parameters('storageAccountName'), '.blob.core.windows.net/', parameters('vmName'), '/', variables('osDiskName'))]"
+            "uri": "[concat('http://', parameters('storageAccountName'), '.blob.core.windows.net/', concat('vhds-', uniqueString(parameters('vmName'))), '/', variables('osDiskName'))]"
```
bmoore-msft: Also see #9 in the BP guide for using uri's

sap-2-tier-marketplace-image/shared/os-disk-parts.json
```diff
@@ -100,7 +101,7 @@
             "uri": "[parameters('userImageVhdUri')]"
           },
           "vhd": {
-            "uri": "[concat('http://',parameters('userImageStorageAccount'),'.blob.core.windows.net/', resourceGroup().name, '/', parameters('vmName'), '-', variables('osDiskName'))]"
+            "uri": "[concat('http://',parameters('userImageStorageAccount'),'.blob.core.windows.net/', concat('vhds-', uniqueString(resourceGroup().id)), '/', parameters('vmName'), '-', variables('osDiskName'))]"
```
bmoore-msft: Also see #9 in the BP guide for using uri's

MSSedusch: I will make a bigger reorg of all sap templates soon. Can you in the meantime merge this PR?

Thanks

PR #2355, opened 2016-08-15
Ci

gatneil:
Update a couple parameter files to utilize Travis CI properly


gatneil:
just changed parameters file to use CI properly; CI passed; merging

PR #2363, opened 2016-08-17
Resolve location error

markscholman:
### Description of the change

When using the [resourceGroup().location] I got an error about that the region is not known. When hardcoding or parameterize these it worked.

### DIFF
--- a/app-service-environment-serviceplan-logicapp/azuredeploy.json
+++ b/app-service-environment-serviceplan-logicapp/azuredeploy.json
@@ -8,13 +8,12 @@
         "description": "The name to use for the new AppService Environment"
       }
     },
-    {
-      "location": {
+    "location": {
         "type": "string",
-        "description": "The geo region to deploy your ASE" in"
-      }
+        "metadata": {
+          "description": "The geo region to deploy your ASE in"
+        }
     },
-    }
     "ipSslAddressCount": {
       "type": "int",
       "defaultValue": 1,
--- a/app-service-environment-serviceplan-logicapp/azuredeploy.json
+++ b/app-service-environment-serviceplan-logicapp/azuredeploy.json
@@ -8,6 +8,13 @@
         "description": "The name to use for the new AppService Environment"
       }
     },
+    {
+      "location": {
+        "type": "string",
+        "description": "The geo region to deploy your ASE" in"
+      }
+    },
+    }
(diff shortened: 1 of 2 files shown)


azurecla:
Hi __@markscholman__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
    </span>

TTYL, AZPRBOT;

krnese:
@markscholman there's another "location" within the properties envelope that refers to resourceGroup().location. can you add the parameter to this too.

app-service-environment-serviceplan-logicapp/azuredeploy.json
```diff
@@ -8,6 +8,12 @@
         "description": "The name to use for the new AppService Environment"
       }
     },
+    "location": {
```
bmoore-msft: using resourceGroup().location is a better practice here, is there something specific we need to break it out into a param for?

markscholman: When doing so deployment fails. That's why I changed it back to a parameter and commit it. Maybe a fix is needed in the parsing of the template in order to support this.

bmoore-msft: I'm not able to get a successful deployment with this template regardless... it looks like the RP does have a bug in handling locations, but even when I get around that the deployment fails on finding the vnet.  How do we fix that?

markscholman: If i recall correctly the network should be already in place and is used as a resource reference. (What can be outside the RG you deploy this template to). The subnet should be a dedicated worker subnet only no other vm's should live there.

bmoore-msft: Yeah I tried a vnet in the RG and outside the RG (in the same region) but never got a successful deployment of this template.

bmoore-msft:
@yossidahan - I've been unable to get any variation of this sample to successfully deploy, even with a classic vnet (deployment times out).  Can you update the sample and provide instructions to make it work?

azurecla:
__@markscholman__, Thanks for signing the contribution license agreement so quickly! Actual humans will now validate the agreement and then evaluate the PR.
<br />Thanks, AZPRBOT;

PR #2391, opened 2016-08-23
Moved most service names to parameters to keep things consistent & si…

Preskton:
Moved parameters that named services down to variables so that we generate a consistent set of predictably named services instead of relying on the user to create them.  Avoids conflicts with existing resources.


azurecla:
Hi __@Preskton__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
    </span>

TTYL, AZPRBOT;

PR #2424, opened 2016-08-30
updated solution

krnese:
updated solution. redesigned template into nested templates


marcvaneijk:
@krnese can you re-add the _assetlocation parameter again and replace the omsNestedTemplateRepo variable with the _assetlocation parameter. In a forked repo, the only requirements is to change the param file for edit, allowing to reference the items in the fork instead of the origin.

krnese:
Go go go! :-)

marcvaneijk:
@krnese the variable name is renamed to _assetlocation, but it should be a param, this allows a different param file to be used in a fork without changing azuredeploy.json. If you have trouble replacing the variable throughout the template, you can create _assetlocation as a param and use the existing variable to reference that.

marcvaneijk:
@krnese the omsAutomation.json contains an _assetlocation variable that points to your own repo

oms-all-deploy/azuredeploy.json
```diff
@@ -80,1129 +80,114 @@
...
-      },
-      "azureRmCompute": {
-        "name": "AzureRm.Compute",
-        "url": "https://devopsgallerystorage.blob.core.windows.net/packages/azurerm.compute.1.3.3.nupkg"
-      },
-      "azureRmAutomation": {
-        "name": "AzureRm.Automation",
-        "url": "https://devopsgallerystorage.blob.core.windows.net/packages/azurerm.automation.1.0.11.nupkg"
+    },
+    "_assetLocation": {
```
bmoore-msft: We want to standardize our approach for using artifacts in samples - see: https://github.com/Azure/azure-quickstart-templates/blob/master/1-CONTRIBUTION-GUIDE/best-practices.md#samples-that-contain-extra-artifacts-custom-scripts-nested-templates-etc

this allows for a standard way to deploy samples that will work in a customer environment (not just a public repo)

krnese: That I get. In this case we were aiming to expose as few paramaters as possible to the user to simplify the onboarding process.

krnese:
Please merge. Also verified locally
image

oms-all-deploy/nestedtemplates/omsAutomation.json
```diff
@@ -0,0 +1,433 @@
...
+    },
+    "azureAdminPwd": {
+      "type": "securestring",
+      "metadata": {
+        "description": "Enter the pwd for the service admin user. The pwd is enrypted during runtime and in the Automation assets"
+      }
+    },
+    "_artifactsLocation": {
+      "type": "string",
+      "defaultValue": "https://raw.githubusercontent.com/Azure/azure-quickstart-templates/master/oms-all-deploy",
```
bmoore-msft: Since this is a nested template I think you want to remove the default value to ensure this is passed in an matches the global setting for artifacts

oms-all-deploy/azuredeploy.parameters.json
```diff
@@ -19,6 +19,15 @@
...
       "value": "West Europe"
+    },
+    "azureAdminPwd": {
+      "value": "GEN-PASSWORD"
+    },
+    "azureAdmin": {
+      "value": "azureadmin"
+    },
+    "_artifactsLocation": {
+      "value": "https://raw.githubusercontent.com/krnese/AzureDeploy/master/OMS/MSOMS/OMSnested"
```
bmoore-msft: remove this now that we've done a merge...

PR #2437, opened 2016-09-02
Fix the name of the runbook to schedule

MCKLMT:
### Description of the change
The name of the runbook is not correct. The script targets the OMS SQL ingestion runbook instead of the web app ingestion runbook.


azurecla:
Hi __@MCKLMT__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
    </span>

TTYL, AZPRBOT;

PR #2450, opened 2016-09-08
CustomScript rollout - stage #4

ahmetalpbalkan:
See #2340 for more context.
cc: @gatneil to be merged preferably before Monday.
cc: @bingosummer bosh-setup template is now being migrated to the new extension FYI


gatneil:
Looks good to me. Merging

Issue #2459, opened 2016-09-09
Deployment using "Deploy to Azure" not working. Not using "Storage Account Name" parameter.


aceptra:
Deployment using "Deploy to Azure" not working. Assuming it is using latest deployment scripts.
### Repro steps (_if necessary, delete otherwise_)
1. "Deploy to Azure"
2.  Modify parameters appropriately including default storage name of "cbstorage". I changed to "somelongnamestorage".
3. Preflight error check throws validation.
## 

**Preflight validation failed**
- The storage account named cbstoragejb is already taken. (Code: StorageAccountAlreadyTaken)
- The storage account named cbstorage0 is already taken. (Code: StorageAccountAlreadyTaken)

Obviously it isn't using the parameter as it should.

MCKLMT:
The pull request has been accepted and merged. Could you close the issue?

aceptra:
Closing.

PR #2461, opened 2016-09-10
Fix Parameterize 'cbstorage'

aceptra:
See Issue #2459

azuredeploy.json and jumpbox-resources-jumpbox-windows did not completely use "storageAccountNamePrefix" parameter.

Bug:
Preflight validation failed

    The storage account named cbstoragejb is already taken. (Code: StorageAccountAlreadyTaken)

    The storage account named cbstorage0 is already taken. (Code: StorageAccountAlreadyTaken)

Fix:
* azuredeploy.json resource creation statments to use storageAccountNamePrefix instead of hard coded 'cbstorage'
* add storageAccountNamePrefix parameter to jumpbox-resources-jumpbox-windows.json and its vm storage resource creation. This parameter is passed from "jumpbox-resources" creation in azuredeploy.json

NOTE:    jumpbox-resources-jumpbox-ubuntu.json and others may also need this fix. I only tested with the Windows jumpbox.


azurecla:
Hi __@aceptra__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
    </span>

TTYL, AZPRBOT;

azurecla:
__@aceptra__, Thanks for signing the contribution license agreement so quickly! Actual humans will now validate the agreement and then evaluate the PR.
<br />Thanks, AZPRBOT;

bmoore-msft:
parameterizing the storage account name will make the sample more likely to fail deployment than the previous version - is there a reason that's needed?

aceptra:
Static "cbstorage" doesn't not work out of the box with my testing. Appends creating "cbstorage0" and "cbstoragejb" doesn't create an Azure-wide unique name. See the error:

bmoore-msft:
likely because it's the default value - can we use uniqueString instead?  e.g. concat('cbstorage', uniqueString(resourceGroup().id)?

aceptra:
Likely.. yes that should be fine. Will you make the change in the Main trunk?

But the bigger picture might desire control of that a bit, for naming conventions. 

For mine I updated only jumpbox-resources-jumpbox-windows.json to also use this parameter and  azuredeploy.json passes it in "jumpbox-resources". Perhaps others would need this also.

bmoore-msft:
you're right that is a curious commit - would most definitely lead to name collisions, it's likely that was just missed in the review...

For some bg, we'd like to simply use of these and so are trying to adhere to some BPs that have less margin for error (like supplying storage account names).  We're doing it as changes come in until we can get some automation in place.

That said, that BP would be a bigger commit so we'll put this through for now.

PR #2465, opened 2016-09-12
Fixed  template with latest RP API support

nilaydshah:
Added required field: sourceResourceId for configure protection.
Updated template with latest RP api version.


singhkays:
merged

PR #2469, opened 2016-09-12
Update existing storage account

krnese:


krnese:
Updated readme.md so that Azure Deploy button is working. However, the deployment using Travis will fail since the template will let a user specify an existing storage account that should be added.

PR #2477, opened 2016-09-13
mysql-standalone-server-ubuntu: remove wrong setting key

ahmetalpbalkan:
@marcvaneijk you're not supposed to add `autoUpgradeMinorVersion` to `settings`. it should be one level above.

### DIFF
--- a/mysql-standalone-server-ubuntu/azuredeploy.json
+++ b/mysql-standalone-server-ubuntu/azuredeploy.json
@@ -198,7 +198,6 @@
         "type": "CustomScriptForLinux",
         "typeHandlerVersion": "1.4",
         "settings": {
-          "autoUpgradeMinorVersion": true,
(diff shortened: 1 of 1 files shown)


ahmetalpbalkan:
merging as trivial.

ahmetalpbalkan:
To give more context on why I submitted the patch: Apparently a customer has tried to deploy the template and it failed with error:

> VM has reported a failure when processing extension 'newuserscript'. Error message: \"Enable failed: failed to get configuration: json validation error: invalid public settings JSON: autoUpgradeMinorVersion: Additional property autoUpgradeMinorVersion is not allowed\".

Why it worked before and why it started failing now: I recently migrated this template to use the brand new CustomScript extension v2.0, which does strict validation of keys in `"settings"` section and extraneous keys will cause errors.

PR #2500, opened 2016-09-19
tempalte and script

neilpeterson:
### Contributing guide

### Description of the change

Template and script to deploy .Net Core application across Linus VM and Azure SQL DB.


dotnet-core-music-linux/README.md
```diff
@@ -0,0 +1,14 @@
+# Deploy Two Tier Application on Linux and Azure SQL DB
+
+This template deploys a two tier demo application and all necessary infrastructural components. When deployed, the resulting resource group will contain a fully working, highly available, and secure application. Deploy the Azure resource group and application using the below button. Once completed, the application can be access on the Azure public IP address assigned in the resource group.
```
bmoore-msft: Can also be accessed by the dns, correct?

dotnet-core-music-linux/azuredeploy.json
```diff
@@ -0,0 +1,435 @@
+{
+  "$schema": "https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#",
+  "contentVersion": "1.0.0.0",
+  "parameters": {
+    "vmName": {
```
bmoore-msft: move this to a variable

dotnet-core-music-linux/azuredeploy.json
```diff
@@ -0,0 +1,435 @@
...
+    "virtualNetworkName": "VirtualNetwork",
+    "vnetId": "[resourceId('Microsoft.Network/virtualNetworks', variables('virtualNetworkName'))]",
+    "networkSecurityGroup": "NetworkSecurityGroup",
+    "musicStoreSqlName": "[concat('sql', uniqueString(resourceGroup().id))]",
+    "sshKeyPath": "[concat('/home/',parameters('adminUsername'),'/.ssh/authorized_keys')]"
+  },
+
+  "resources": [
+    {
+      "apiVersion": "2015-05-01-preview",
```
bmoore-msft: use latest apiVersion

dotnet-core-music-linux/azuredeploy.json
```diff
@@ -0,0 +1,435 @@
...
+              "destinationAddressPrefix": "*",
+              "access": "Allow",
+              "priority": 100,
+              "direction": "Inbound"
+            }
+          },
+          {
+            "name": "ssh",
+            "properties": {
+              "description": "http endpoint",
```
bmoore-msft: ssh endpoint?

dotnet-core-music-linux/azuredeploy.json
```diff
@@ -0,0 +1,435 @@
...
+          "imageReference": {
+            "publisher": "[variables('imagePublisher')]",
+            "offer": "[variables('imageOffer')]",
+            "sku": "[variables('ubuntuOSVersion')]",
+            "version": "latest"
+          },
+          "osDisk": {
+            "name": "osdisk",
+            "vhd": {
+              "uri": "[concat('http://', variables('vhdStorageName'), '.blob.core.windows.net/vhds/', 'osdisk', copyindex(), '.vhd')]"
```
bmoore-msft: See #9 in https://github.com/Azure/azure-quickstart-templates/blob/master/1-CONTRIBUTION-GUIDE/best-practices.md

dotnet-core-music-linux/azuredeploy.json
```diff
@@ -0,0 +1,435 @@
...
+            "displayName": "config-app"
+          },
+          "properties": {
+            "publisher": "Microsoft.Azure.Extensions",
+            "type": "CustomScript",
+            "typeHandlerVersion": "2.0",
+            "autoUpgradeMinorVersion": true,
+            "settings": {
+              "fileUris": [
+                "https://raw.githubusercontent.com/Azure/azure-quickstart-templates/master/dotnet-core-music-linux/scripts/config-music.sh"
```
bmoore-msft: See: https://github.com/Azure/azure-quickstart-templates/blob/master/1-CONTRIBUTION-GUIDE/best-practices.md#samples-that-contain-extra-artifacts-custom-scripts-nested-templates-etc for handling artifacts

dotnet-core-music-linux/azuredeploy.parameters.json
```diff
@@ -0,0 +1,21 @@
...
+            "value": "GEN-SSH-PUB-KEY"
+        },
+        "adminUsername": {
+            "value": "adminuser"
+        },
+        "publicipaddressDnsName": {
+            "value": "GEN-UNIQUE-13"
+        },
+        "sqlAdminPassword": {
+            "value": "Password16"
```
bmoore-msft: GEN-PASSWORD

dotnet-core-music-linux/azuredeploy.json
neilpeterson: @bmoore-msft for the notes here. For this one specifically, what are you recommending I move to a variable?

bmoore-msft: yes - the scenario you want to demonstrate doesn't really benefit from a user defined resource name...  one less question to ask, one less thing to fail...

dotnet-core-music-linux/azuredeploy.json
neilpeterson: Thanks @bmoore-msft  , appreciate the response. I totally get that the solution here is not dependent on the VM name, however this template will be used in a training exercise / as a training mechanism. I think that ‘connection’ made when having control over this value will be worthwhile as participants make the mental transition (datacenter – Azure). Can I get a pass on this one?

The rest of this is good stuff, I’ve learned some new things – Thanks.

dotnet-core-music-linux/azuredeploy.json
neilpeterson: On this one, I think that I get this conceptually, but not sure that I think this fits my scenario either. The location should not change, nor need to be modified. I do not want to add the extra parameters here. In this case it would be unnecessary cruft for the training participant to digest. If I cannot get a pass with the template as is, what if I were to place the script location value in a variable?

Again, thanks for the help.

dotnet-core-music-linux/azuredeploy.json
neilpeterson: Thanks @bmoore-msft , appreciate the response. I totally get that the solution here is not dependent on the VM name, however this template will be used in a training exercise / as a training mechanism. I think that ‘connection’ made when having control over this value will be worthwhile as participants make the mental transition (datacenter – Azure). Can I get a pass on this one?

The rest of this is good stuff, I’ve learned some new things – Thanks.

dotnet-core-music-linux/azuredeploy.json
neilpeterson: Thanks - good stuff.

dotnet-core-music-linux/azuredeploy.json
neilpeterson: Thanks for the catch.

dotnet-core-music-linux/azuredeploy.json
neilpeterson: Thanks for the catch.

dotnet-core-music-linux/README.md
neilpeterson: Thanks for the catch.

dotnet-core-music-linux/azuredeploy.parameters.json
neilpeterson: Thanks.

dotnet-core-music-linux/azuredeploy.json
bmoore-msft: Ok, if your purpose for the template is a structured training exercise, then this probably isn't the proper repo for the template (a gist may be a better choice).  We want to use consistent patterns and practices for handling these scenarios - something that will be useful in a real working environment.  For example, in this particular case, we wouldn't tell users to stage artifacts for their deployments (which may contain secrets) in a public location.  It's fine to default for the sample, but shouldn't require changes to the template itself...

dotnet-core-music-linux/azuredeploy.json
neilpeterson: @bmoore-msft - ok after some thought, I agree with you. This has been updated.

dotnet-core-music-linux/azuredeploy.json
neilpeterson: Totally get it, and after some additional thought, agree with the pattern.  

After some additional thinking on this one, I also agree. Can you check out the changes here, and comment back if any further changes are required.

Thanks

neilpeterson:
@bmoore-msft thanks for the help here. I've completed all suggested updates.

dotnet-core-music-linux/azuredeploy.json
```diff
@@ -42,9 +35,17 @@
       "metadata": {
         "description": "Globally unique DNS Name for the Public IP used to access the Virtual Machine."
       }
+    },
+    "fileUris": {
```
bmoore-msft: See: https://github.com/Azure/azure-quickstart-templates/blob/master/1-CONTRIBUTION-GUIDE/best-practices.md#samples-that-contain-extra-artifacts-custom-scripts-nested-templates-etc
for how to handle artifacts

dotnet-core-music-linux/azuredeploy.json
```diff
@@ -0,0 +1,446 @@
...
+      "name": "[concat(variables('vmName'),copyindex())]",
+      "copy": {
+        "name": "virtualMachineLoop",
+        "count": "[parameters('numberOfInstances')]"
+      },
+      "location": "[resourceGroup().location]",
+      "tags": {
+        "displayName": "virtual-machine"
+      },
+      "dependsOn": [
```
bmoore-msft: I think we need a dependency on the NIC?

dotnet-core-music-linux/azuredeploy.json
```diff
@@ -0,0 +1,446 @@
...
+      }
+    },
+    "publicipaddressDnsName": {
+      "type": "string",
+      "minLength": 1,
+      "metadata": {
+        "description": "Globally unique DNS Name for the Public IP used to access the Virtual Machine."
+      }
+    },
+    "fileUris": {
```
bmoore-msft: See https://github.com/Azure/azure-quickstart-templates/blob/master/1-CONTRIBUTION-GUIDE/best-practices.md#samples-that-contain-extra-artifacts-custom-scripts-nested-templates-etc for using artifacts in the samples

PR #2501, opened 2016-09-19
AzureGermany Update

gitralf:
updated vm sizes and storage location to work with Azure Germany etc.

### DIFF
--- a/101-vm-simple-linux/azuredeploy.json
+++ b/101-vm-simple-linux/azuredeploy.json
@@ -14,6 +14,13 @@
         "description": "Password for the Virtual Machine."
       }
     },
+    "vmSize": {
+      "type": "string",
+      "defaultValue": "Standard_D1_v2",
+      "metadata": {
+         "description": "The size of the VM"
+      }
+    },  
     "dnsLabelPrefix": {
       "type": "string",
       "metadata": {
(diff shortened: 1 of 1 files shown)


azurecla:
Hi __@gitralf__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
    </span>

TTYL, AZPRBOT;

azurecla:
__@gitralf__, Thanks for signing the contribution license agreement so quickly! Actual humans will now validate the agreement and then evaluate the PR.
<br />Thanks, AZPRBOT;

go-ethereum-on-ubuntu/azuredeploy.json
```diff
@@ -70,11 +71,13 @@
     {
       "type": "Microsoft.Storage/storageAccounts",
       "name": "[variables('storageAccountName')]",
-      "apiVersion": "[variables('apiVersion')]",
+      "apiVersion": "[variables('storageApiVersion')]",
```
bmoore-msft: Use literal values for apiVersion - otherwise schema validation cannot be done

101-vm-simple-linux/azuredeploy.json
```diff
@@ -14,6 +14,13 @@
         "description": "Password for the Virtual Machine."
       }
     },
+    "vmSize": {
```
bmoore-msft: We want to avoid this pattern of putting VM sizes in parameters - it creates a maintenance problem.  Let's pick a size that works for the scenario and leave it as a var.

go-ethereum-on-ubuntu/azuredeploy.json
```diff
@@ -36,10 +36,10 @@
         "Standard_A2",
         "Standard_A3",
         "Standard_A4",
-        "Standard_D1",
-        "Standard_D2",
-        "Standard_D3",
-        "Standard_D4"
+        "Standard_D1_v2",
```
bmoore-msft: We want to avoid this pattern of putting VM sizes in parameters - it creates a maintenance problem.  Let's pick a size that works for the scenario and leave it as a var.

PR #2537, opened 2016-09-23
Update api mgmt template to use new API version and RG location

jonsabo:
### Changelog
- Update API management template to use API version 2016-07-07 instead of 2014-02-14
- Change location to use the location of the resource group rather than a hard coded location
- Change capacity parameter to an int rather than a string. Previous template also restricted capacity to only 1 and 2, though API mgmt services can have potentially unlimited. Removed restriction.
- Change default tier to Standard as the old default Developer has no SLA and is not recommended for production
- Clarify some parameter descriptions

### DIFF
--- a/101-azure-api-management-create/azuredeploy.json
+++ b/101-azure-api-management-create/azuredeploy.json
@@ -1,80 +1,60 @@
 ﻿{
-    "$schema": "http://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#",
-    "contentVersion": "1.0.0.0",
-    "parameters":
-    {
-        "publisherEmail":
-        {
-            "type": "string",
-            "minLength": 1,
-	    "metadata": {
-		"description": "The email address of the owner of the service"
-	    }
-        }
-        ,
-        "publisherName":
-        {
-            "type": "string",
-            "minLength": 1,
-	    "metadata": {
-		"description": "The name of the owner of the service"
-	    }	
-        }
-        ,
-        "sku":
-        {
-            "type": "string",
-            "allowedValues": [
-            "Developer",
-            "Standard",
-            "Premium"
-            ],
-            "defaultValue": "Developer",
-	    "metadata": {
-		"description": "The pricing tier of this API Management service"
-	    }	
-        }
-        ,
-        "skuCount":
-        {
-            "type": "string",
-            "allowedValues": [
-            "1",
-            "2"
-            ],
-            "defaultValue": "1",
-	    "metadata": {
-		"description": "The instance size of this API Management service."
-	    }	
-        }
+  "$schema": "http://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#",
+  "contentVersion": "1.0.0.0",
+  "parameters": {
+    "publisherEmail": {
+      "type": "string",
+      "minLength": 1,
+      "metadata": {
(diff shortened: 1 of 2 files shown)


azurecla:
Hi __@jonsabo__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
    </span>

TTYL, AZPRBOT;

101-azure-api-management-create/azuredeploy.json
```diff
@@ -1,80 +1,60 @@
...
-            "sku":
-            {
-                "name": "[parameters('sku')]",
-                "capacity": "[parameters('skuCount')]"
-            }
-            ,                
-            "publisherEmail": "[parameters('publisherEmail')]",
-            "publisherName": "[parameters('publisherName')]"
-        }
+      "apiVersion": "2016-07-07",
```
daschult: This apiVersion is not currently supported by the Azure Resource Schemas. Have you added a schema to the Azure/azure-resource-manager-schemas repository?

jonsabo: Hey @daschult I have not, I essentially used this method:
https://azure.microsoft.com/en-us/documentation/articles/resource-manager-supported-services/#supported-api-versions
to find the latest API version supported by the resource then discovered the changes required in the template via some trial and error.

I just took a look at the schemas repo, but there doesn't appear to be too much info in the way of how to contribute new schema. I could take a crack at it going by what I figured out, but I am not sure what the 'correct' way to do it is. If you could provide some advice on that (perhaps out of this thread, or point me to some resources, a mailing list etc.) I'd be happy to try though.

jeffpatton1971: Can we get some clarification on your statement @daschult it sounds like you're saying we need to write a schema for a quickstart template? Wouldn't this be something that Microsoft would own?

singhkays: @miaojiang Since you contributed this template originally, could you shed some light on this converstaion?

jeffpatton1971: Hi @singhkays thanks for chiming in :)

miaojiang: @daschult The APIM team will add the schema to ARM schemes next week. In other words, it's fine to update the apiVersion to 2016-07-07.

@jonsabo I would recommend keeping the default sku to be Developer given the cost of Standard tier instance. All other changes look good. Thanks!

jonsabo: thanks @miaojiang, I'll go ahead and change it back to developer edition when I'm back in the office. I think with the extra description it'll be fine, I just didn't want people to be surprised when they built one but without an SLA.

jonsabo:
Changed tier back to developer, so going by the above thread it looks like this is good to merge unless there are any other objections.

bmoore-msft:
LGTM - just need the CLA bit flipped...

azurecla:
__@jonsabo__, Thanks for signing the contribution license agreement so quickly! Actual humans will now validate the agreement and then evaluate the PR.
<br />Thanks, AZPRBOT;

jonsabo:
Looks like there is also a PR to add this API version to the schemas repo now as well
https://github.com/Azure/azure-resource-manager-schemas/pull/245

PR #2548, opened 2016-09-26
Adding an automatic VM-to-VM bandwidth meter template

AlekseiPolkovnikov:
### Contributing guide

### Changelog
*
*
*
*
*

### Description of the change


vm-to-vm-bandwidth-meter/azuredeploy.json
```diff
@@ -0,0 +1,611 @@
+{
+  "$schema": "https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#",
+  "contentVersion": "1.0.0.0",
+  "parameters": {
+    "probeVmSize": {
```
bmoore-msft: Can we make this a variable with a default that is globally available?  Otherwise I think we'll see too many failures trying to match region with VM Size availability...  and we don't have a maintenance task keeping this up to date.  The readme can underscored the role size plays here...

vm-to-vm-bandwidth-meter/azuredeploy.json
```diff
@@ -0,0 +1,611 @@
...
+        "Brazil South",
+        "Canada Central",
+        "Canada East"
+      ],
+      "defaultValue": "East US",
+      "metadata": {
+        "description": "Location of the VM that runs the test."
+      }
+    },
+    "targetVmSize": {
```
bmoore-msft: same here...

vm-to-vm-bandwidth-meter/azuredeploy.json
```diff
@@ -0,0 +1,611 @@
...
+      "targetVM"
+    ],
+    "subnetRef": [
+      "[concat(variables('vnetID'),'1','/subnets/',variables('subnetName'))]",
+      "[concat(variables('vnetID'),'2','/subnets/',variables('subnetName'))]"
+    ]
+  },
+  "resources": [
+    {
+      "type": "Microsoft.Storage/storageAccounts",
```
bmoore-msft: can we update this to the new schema: https://github.com/Azure/azure-quickstart-templates/blob/master/101-storage-account-create/azuredeploy.json

vm-to-vm-bandwidth-meter/azuredeploy.json
```diff
@@ -0,0 +1,611 @@
...
+          "networkInterfaces": [
+            {
+              "id": "[resourceId('Microsoft.Network/networkInterfaces', concat(variables('nicName'), '1'))]"
+            }
+          ]
+        }
+      },
+      "resources": [
+        {
+          "type": "extensions",
```
bmoore-msft: Can we update these to the new extension publisher/type... e.g.
        "publisher": "Microsoft.Azure.Extensions",
        "type": "CustomScript",
        "typeHandlerVersion": "2.0",

Same as the nix one below...

vm-to-vm-bandwidth-meter/azuredeploy.parameters.json
```diff
@@ -0,0 +1,12 @@
+{
+  "$schema": "http://schema.management.azure.com/schemas/2015-01-01/deploymentParameters.json#",
+  "contentVersion": "1.0.0.0",
+  "parameters": {
+    "adminUsername": {
+      "value": "serveradmin"
```
bmoore-msft: GEN-USER

vm-to-vm-bandwidth-meter/azuredeploy.json
AlekseiPolkovnikov: It is Standard_A2 by default, but to make the tool/template really usable the intention was to expose it to be able to measure bandwidth any-size-to-any-size (all possible regions). There is a note in the readme however, that the size should be available for the region. Would appreciate if we can leave this easily changeable.

vm-to-vm-bandwidth-meter/azuredeploy.json
AlekseiPolkovnikov: Same thing..

vm-to-vm-bandwidth-meter/azuredeploy.json
AlekseiPolkovnikov: Getting: "Handler 'Microsoft.Azure.Extensions.CustomScript' has reported failure for VM Extension 'CustomScriptExtension' with terminal error code '1007' and error message: 'Install failed for plugin (name: Microsoft.Azure.Extensions.CustomScript, version 2.0.1) with exception The specified executable is not a valid application for this OS platform.'"
Can you share an example for Windows?

vm-to-vm-bandwidth-meter/azuredeploy.json
```diff
@@ -0,0 +1,611 @@
...
+        "metadata": {
+          "description": "Password for the probe and target VMs."
+        }
+      },
+      "_artifactsLocation": {
+        "type": "string",
+        "metadata": {
+          "description": "The base URI where artifacts required by this template are located. When the template is deployed using the accompanying scripts, a private location in the subscription will be used and this value will be automatically generated."
+        },
+        "defaultValue": "https://raw.githubusercontent.com/AlekseiPolkovnikov/azure-quickstart-templates/bw/vm-to-vm-bandwidth-meter"
```
singhkays: URL needs to be this repo

vm-to-vm-bandwidth-meter/azuredeploy.json
```diff
@@ -0,0 +1,611 @@
...
+        "metadata": {
+          "description": "The sasToken required to access _artifactsLocation.  When the template is deployed using the accompanying scripts, a sasToken will be automatically generated."
+        },
+        "defaultValue": ""
+      }
+    },
+  "variables": {
+    "windowsOSVersion": [
+      "2012-R2-Datacenter",
+      "14.04.5-LTS"
```
singhkays: variable name should be osVersion since it's linux as well

vm-to-vm-bandwidth-meter/azuredeploy.json
bmoore-msft: Ok, but then let's reduce the list to say one per series.  We're adding a 150 lines to the template for this and users are more likely to fail matching the region/series than they are to struggle to figure out how to change a variable value...

vm-to-vm-bandwidth-meter/azuredeploy.json
bmoore-msft: ok, nm - they haven't updated windows yet...

AlekseiPolkovnikov:
@bmoore-msft, @singhkays thanks. Partially fixed, please see above.

vm-to-vm-bandwidth-meter/azuredeploy.json
AlekseiPolkovnikov: done

vm-to-vm-bandwidth-meter/azuredeploy.json
AlekseiPolkovnikov: done

vm-to-vm-bandwidth-meter/azuredeploy.parameters.json
AlekseiPolkovnikov: done

vm-to-vm-bandwidth-meter/azuredeploy.json
AlekseiPolkovnikov: OK, leaving as is.

vm-to-vm-bandwidth-meter/azuredeploy.json
AlekseiPolkovnikov: But what if the user has a specific instance type in mind and wants to validate real quick? For (A-series networking is different enough, for example) The idea was to give a tool not just the example.. I understand this makes more lines, however. Failing for reason template is also educating a user, if the link in README is ignored.

bmoore-msft: We're just trying to provide good dev samples, if someone wants to try A8 instead of A2 it's pretty easy to update the list.  It might be worth adding a link to this: https://azure.microsoft.com/en-us/documentation/articles/virtual-machines-windows-sizes/ so folks can understand the differences

bmoore-msft: Thanks - much more manageable... I would also add a link to the series details in the readme - you could even give a couple of use cases you think are interesting...

AlekseiPolkovnikov: @bmoore-msft thanks for the idea, added.

PR #2549, opened 2016-09-26
Update README.md

bmoore-msft:
### Description of the change
made the readme a complete sentence


singhkays:
@bmoore-msft Why don't we make this the starter template?

bmoore-msft:
Do you mean just remove the starter template?  Rename this one? and/or add a little bit more to this one?  I don't want to rename this since it's unambiguous what it contains...  But with all the other 100-level templates, do we really need a starter template (in addition to the blank one)?  Probably not, IMO...

PR #2554, opened 2016-09-28
Added summary property to metadata.md

bmoore-msft:


singhkays:
let's lower case the folder name 100-blank-template for consistency with other names. Also let's remove the started template in favor of this.

bmoore-msft:
ok, will send a separate PR for removal...

PR #2556, opened 2016-09-28
remove 100 starter

bmoore-msft:
removed the starter template - replaced with blank


singhkays:
There might be links in contribution guide which you might want to remove.

bmoore-msft:
you're right - another PR for that...

https://github.com/Azure/azure-quickstart-templates/pull/2561

PR #2570, opened 2016-09-30
201-web-app-sql updated for BPs

bmoore-msft:
### Description of the change
reduced params for SQL config
updated api versions


daschult:
Looks good to me.

PR #2586, opened 2016-10-05
Blockchainmerger

gatneil:
Many of the blockchain templates in the repo are the exact same template just with a different script. This PR merges most of these into one template with the option to select the desired blockchain software as a parameter. For future blockchain templates that are the same, they should simply add their software to this template to reduce clutter and make it easier to find these block chain templates.


gatneil:
@singhkays @bmoore-msft to review the blockchain merger

blockchain/README.md
```diff
@@ -0,0 +1,27 @@
+# Blockchain Template
+
+This Microsoft Azure Resource Manager template deploys a single VM and offers a variety of blockchain software to run on that VM.
```
bmoore-msft: reword?
This Microsoft Azure Resource Manager template deploys a single VM and allows you to select from a number of blockchain software options to run on the VM.

blockchain/metadata.json
```diff
@@ -0,0 +1,7 @@
+{
+  "itemDisplayName": "Blockchain Template",
+    "description": "This template deploys a variety of blockchain services.",
+  "summary": "This template deploys a variety of blockchain services.",
```
bmoore-msft: reword?
This Microsoft Azure Resource Manager template deploys a single VM and allows you to select from a number of blockchain software options to run on the VM.

PR #2603, opened 2016-10-07
Adding sysbench CPU performance meter template

AlekseiPolkovnikov:
### Contributing guide

### Changelog
*
*
*
*
*

### Description of the change


bmoore-msft:
What do you think about refactoring your samples into 1 base VM and then other templates that apply the extension?

AlekseiPolkovnikov:
@bmoore-msft yes for the part that uses Linux VM doable, not sure about cleanliness/simplicity/coupling however (parameters/outputs in and out, nested templates, limitation on sizes intact, etc.)

PR #2616, opened 2016-10-13
updating to use windows server 2016 image and feedback from Ignite

MahendraAgrawal:
Fixed formatting feedback from Neil
updating templates to use windows server 2016 GA image
updated strings after Ignite feedback.


MahendraAgrawal:
Fixed max length issues on metadata.json files.

MahendraAgrawal:
This is because of custom image. this always happens and Kay/Neil know about it. Please help merge it asap as team is dependent on this pull. thanks a lot for help.

gatneil:
Still doesn't follow the best practice around _artifactsLocation. Merging because it's hi-pri, but please address in a future PR

PR #2640, opened 2016-10-18
Migrate templates to new CustomScript

ahmetalpbalkan:
Apparently some templates are getting still merged to use the
old CustomScriptForLinux extension, for which we are trying to drive
the usage down. (#2340) Replacing those templates to use the new extension.

CC: @gatneil @bmoore-msft Can you please help making sure templates using
the old `CustomScriptForLinux` do not get merged anymore?

Thanks.


ahmetalpbalkan:
@bmoore-msft reverted changes on `vm-{vm-cpu-sysbench,disk-performance}-meter` templates per offline discussion. PTAL again.

PR #2647, opened 2016-10-19
update storage and add validate option

bmoore-msft:
### Description of the change
updated the script to use same storage account as PS1
added support for ValidateOnly

### DIFF
--- a/azure-group-deploy.sh
+++ b/azure-group-deploy.sh
@@ -1,5 +1,5 @@
 #!/bin/bash -e
-while getopts "a:l:g:s:f:e:u" opt; do
+while getopts "a:l:g:s:f:e:u:v" opt; do
     case $opt in
         a)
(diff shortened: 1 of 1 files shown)


azure-group-deploy.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/bash -e
-while getopts "a:l:g:s:f:e:u" opt; do
+while getopts "a:l:g:s:f:e:u:v" opt; do
```
philon-msft: 'u' should not have a colon after it, since it does not require an option argument

bmoore-msft: Ok, thanks - you mean like this: https://github.com/Azure/azure-quickstart-templates/pull/2650

philon-msft: Yep, Dan fixed it :)

PR #2652, opened 2016-10-20
New version 2.1.1 for the bosh-setup template

bingosummer:
### Changelog
- Add a retry logic for uploading stemcell and release
- Upgrade to CustomScript 2.0
- Some other minor improvements


bosh-setup/azuredeploy.json
```diff
@@ -561,9 +561,9 @@
...
       "properties": {
-        "publisher": "Microsoft.OSTCExtensions",
-        "type": "CustomScriptForLinux",
-        "typeHandlerVersion": "1.4",
+        "publisher": "Microsoft.Azure.Extensions",
+        "type": "CustomScript",
+        "typeHandlerVersion": "2.0",
         "settings": {
           "fileUris": "[variables('filesToDownload')]"
         },
```
bingosummer: @ahmetalpbalkan 
CustomScript is upgraded to 2.0 in this commit.

bosh-setup/scripts/utils.sh
```diff
@@ -0,0 +1,23 @@
...
+  max_retries=10
+  interval=30
+  while [ ${retry} -lt ${max_retries} ]; do
+    echo "Operation: $1, Retry #${retry}"
+    eval $1
+    if [ $? -eq 0 ]; then
+      echo "Successfully."
+      break
+    else
+      let retry=retry+1
```
ahmetalpbalkan: maybe add a log here as well `echo Retrying`?

bosh-setup/scripts/utils.sh
```diff
@@ -0,0 +1,23 @@
...
+function retryop()
+{
+  retry=0
+  max_retries=10
+  interval=30
+  while [ ${retry} -lt ${max_retries} ]; do
+    echo "Operation: $1, Retry #${retry}"
+    eval $1
+    if [ $? -eq 0 ]; then
+      echo "Successfully."
```
ahmetalpbalkan: `Successful` sounds a bit better?

bosh-setup/update-china-storage-account.sh
```diff
@@ -3,13 +3,13 @@
...
-azure storage blob upload scripts/deploy_bosh.sh ${container_name} ${template_version}/scripts/deploy_bosh.sh
-azure storage blob upload scripts/deploy_cloudfoundry.sh ${container_name} ${template_version}/scripts/deploy_cloudfoundry.sh
-azure storage blob upload scripts/init.sh ${container_name} ${template_version}/scripts/init.sh
-azure storage blob upload scripts/setup_env ${container_name} ${template_version}/scripts/setup_env
-azure storage blob upload scripts/setup_env.py ${container_name} ${template_version}/scripts/setup_env.py
-azure storage blob upload manifests/bosh.yml ${container_name} ${template_version}/manifests/bosh.yml
-azure storage blob upload manifests/single-vm-cf.yml ${container_name} ${template_version}/manifests/single-vm-cf.yml
-azure storage blob upload manifests/multiple-vm-cf.yml ${container_name} ${template_version}/manifests/multiple-vm-cf.yml
+directories="scripts manifests"
+for directory in $directories
```
ahmetalpbalkan: these two scripts look like they're written by different people :) I guess mostly because `do` statements are on a separate line, unlike retry.sh.

ahmetalpbalkan:
Thanks!

bingosummer:
@ahmetalpbalkan 
This PR is closed by myself for test. Now reopen it.
It has not been merged.

bingosummer:
@bmoore-msft 
Could you please have a review on this PR? Thanks in advance.

bosh-setup/azuredeploy.json
```diff
@@ -113,10 +113,10 @@
 
     "baseUriAzureCloud": "https://raw.githubusercontent.com/Azure/azure-quickstart-templates/master/bosh-setup/",
```
bmoore-msft: Why are we hardcoding this instead of using the best practice from here: https://github.com/Azure/azure-quickstart-templates/blob/master/1-CONTRIBUTION-GUIDE/best-practices.md#samples-that-contain-extra-artifacts-custom-scripts-nested-templates-etc

bingosummer: @bmoore-msft 
From my understanding, the BP for the artifacts are mainly for the template developers. Because for end users, mostly they will use the artifacts we provide, and won't change them. Specially, for the bosh-setup template, the purpose is POC of Cloud Foundry deployment on Azure. And too many parameters may confuse users. So we try to make it simple.

When our team are developing and testing the template, we just change the hardcoding URL to the private location. I think, this way works well.

bingosummer: @bmoore-msft 
Do you have any updates? Does my understanding make sense? Thanks for your kindly reviewing.

bingosummer:
@ahmetalpbalkan Any concerns or updates?

bosh-setup/azuredeploy.json
bmoore-msft: We want to have templates that work well for developers as well as end-users, showing how templates can be flexible and used across clouds with "standard" tools and scripts.  You can default the parameter values as described in the bp guide and portal users won't need to worry about it.

bingosummer: I understand I can give the default value to the parameter. But from our point of view, we don't want to expose the `baseUriAzureCloud` as a parameter to users because most end-users don't need to change the url.
If this best practice is not a **MUST**, could you please merge the PR at this point? And we can investigate whether we should expose it or not. If users ask for it, we can expose it in future.

bmoore-msft: It is a *must* we want a consistent staging experience - we've updated BP and PR templates to help with messaging.

bingosummer: You mentioned the templates are used across clouds with "standard" tools and scripts. What are the tools? From our understanding, we only need to make sure Azure Portal, CLI and Powershell can deploy the template successfully.
We still think this bp is not applicable to the bosh-setup template because end users never need to change the `baseUriAzureCloud` as a parameter.

bmoore-msft: We agree there are a set of users that don't need to change the parameter value - that's why the BP details an approach that provides a default value for those users.  That same approach also works for other consumers of the repo...

PR #2657, opened 2016-10-20
updates to cisco CSR templates

bmoore-msft:
### Description of the change
some simplification and clean up of the Cisco CSR template for Azure Gov
Also added an "existing vnet" option


daschult:
Looks fine to me. I'll work on updating the validator service so that it stops failing all the time.

PR #2673, opened 2016-10-25
Changed default VM size to D1_v2

sidkri:
### Changelog
* Changed default VM size to D1_v2

### Description of the change
This size is available in all regions including both Azure Gov regions.  D1 is not available in USGov Iowa so is not a good user experience for Azure Gov customers


azurecla:
Hi __@sidkri__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
    </span>

TTYL, AZPRBOT;

PR #2727, opened 2016-11-05
Adding new quickstart template

Teodelas:
Added a new Quickstart template to deploy a standalone Remote Desktop Gateway Server (RDGW).   This can be useful to RDP and SSH into Azure environments when you can't use ports 3389 (RDP) or 22 (SSH).  RDGW uses RDP over HTTPS


bmoore-msft:
Not sure how, but this PR is showing a change in 2K files, can you try this again with a clean branch?

PR #2733, opened 2016-11-07
renamed rhel template

gatneil:


gatneil:
@bmoore-msft @singhkays just FYI this is the RHEL rename

gatneil:
Simple rename; merging

PR #2778, opened 2016-11-17
Adding a NSG template with dedicated SF NodeType

ChackDan:


service-fabric-secure-nsg-cluster-65-node-3-nodetype/azuredeploy.json
```diff
@@ -0,0 +1,1849 @@
...
+        "description": "Remote desktop user Id"
+      }
+    },
+    "adminPassword": {
+      "type": "securestring",
+      "metadata": {
+        "description": "Remote desktop user password. Must be a strong password"
+      }
+    },
+    "vmImagePublisher": {
```
singhkays: Hardcode these image values unless there is another choice of OS

ChackDan:
The customers can choose to deploy on WS2016. So we do need these to be parameterized

service-fabric-secure-nsg-cluster-65-node-3-nodetype/azuredeploy.json
ChackDan: yes, customer can choose other OS. also bring in their own

PR #2779, opened 2016-11-17
Vmss existing vnet

gatneil:


gatneil:
Requires existing resources, so CI wil fail; tested manually; merging

PR #2782, opened 2016-11-19
a commandtoexecute value does not need the scriptfolder name in its p…

lukka:
The value of the commandToExecute attribute of an 'extension' type resource does not need any subdirectory value, as erroneously the example shows in the bp doc file.


azurecla:
Hi __@lukka__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <p>
        It looks like you're working at Microsoft (lucappa). If you're full-time, we DON'T require a contribution license agreement.
    </p>
    <p>
    </p>

TTYL, AZPRBOT;

bmoore-msft:
This is actually a "bug" in the Custom Script extension... when artifacts are downloaded from Azure storage, the folder is included in the local destination on the VM.  When they are downloaded from somewhere other than Azure storage, the folder is not included locally.  So it changes depending on where the artifacts is stored and staged.  We need to figure out a common pattern for this - which may be moving all script artifacts to the root of the sample.  So let's hold off on this for now...

PR #2783, opened 2016-11-21
Aqt/existingdb

robe070:
### Best Practice Checklist

### Changelog
Add a template to Autoscale a LANSA Windows VM Scale Set using an **existing** Database

### DIFF
--- a/lansa-vmss-windows-autoscale-existing-db/README.md
+++ b/lansa-vmss-windows-autoscale-existing-db/README.md
@@ -1,9 +1,9 @@
 # Autoscale a LANSA Windows VM Scale Set using an existing database server
 
-<a href="https://portal.azure.com/#create/Microsoft.Template/uri/https%3A%2F%2Fraw.githubusercontent.com%2Frobe070%2Fazure-quickstart-templates%2Ffeature%2Fexistingdb%2Flansa-vmss-windows-autoscale-existing-db%2Fazuredeploy.json" target="_blank">
+<a href="https://portal.azure.com/#create/Microsoft.Template/uri/https%3A%2F%2Fraw.githubusercontent.com%2FAzure%2Fazure-quickstart-templates%2Fmaster%2Flansa-vmss-windows-autoscale-existing-db%2Fazuredeploy.json" target="_blank">
     <img src="http://azuredeploy.net/deploybutton.png"/>
 </a>
-<a href="http://armviz.io/#/?load=https%3A%2F%2Fraw.githubusercontent.com%2Frobe070%2Fazure-quickstart-templates%2Ffeature%2Fexistingdb%2Flansa-vmss-windows-autoscale-existing-db%2Fazuredeploy.json" target="_blank">
+<a href="http://armviz.io/#/?load=https%3A%2F%2Fraw.githubusercontent.com%2FAzure%2Fazure-quickstart-templates%2Fmaster%2Flansa-vmss-windows-autoscale-existing-db%2Fazuredeploy.json" target="_blank">
     <img src="http://armviz.io/visualizebutton.png"/>
 </a>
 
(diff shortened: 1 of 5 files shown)


azurecla:
Hi __@robe070__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!

TTYL, AZPRBOT;

robe070:
@singhkays This is a very similar template to the last one, except it requires the database to already exist. Apart from that its identical. Might be efficient for @gatneil to review it?

lansa-vmss-windows-autoscale-existing-db/README.md
```diff
@@ -0,0 +1,66 @@
...
+
+| Endpoint        | Version           | Validated  |
+| ------------- |:-------------:| -----:|
+| Microsoft Azure      | - | yes |
+| Microsoft Azure Stack      | TP2      |  incompatible |
+
+## Solution overview and deployed resources
+
+The Autoscale rules are configured as follows
+- sample for CPU (\\Processor\\PercentProcessorTime) in each VM every 1 Minute
```
gatneil: isn't it using Percentage CPU, not \Processor\PercentProcessorTime?

lansa-vmss-windows-autoscale-existing-db/README.md
```diff
@@ -0,0 +1,66 @@
...
+- Its also highly recommended to follow the usage instructions below :)
+
+## Usage Instructions
+
+For full instructions for using this template go to [Azure Deployment Tutorial](http://docs.lansa.com/14/en/lansa022/index.htm#lansa/vldtoolct_0250.htm#_Toc461606162%3FTocPath%3DLANSA%2520Application%2520Deployment%2520Tool|Cloud%2520Tutorials|Microsoft%2520Azure%2520Tutorial|_____0)
+
+
+## Notes
+
+1. Two VMSS. One to install the database; 1 to run the web site. OverProvision = false. This is so that extra VMs are not created which would put more load on the database which slows down provisioning. Failure to provision has not been seen. If it occurs, the VMSS will ensure another VM is created to establish the correct InstanceCount. A second reason is that the database installer VMSS MUST NEVER have more than 1 instance installing at a time. Errors occur when publishing the weblets. As well as the database state not being matched to what the VM thinks the state of the database is in terms of table creation, etc.
```
gatneil: 2 things:

1) I'd be interested to hear why overprovisioning doesn't meet your needs. Does it break the application?
2) If you turn overprovisioning off and there's a failure to provision, scale sets won't ensure another VM is created to establish the correct InstanceCount. The user will need to re-image the offending VM, or delete it then re-deploy the scale set with the same instance count (which will only deploy as many VMs as needed to get to that count)

gatneil:
Please address the above comments :)

lansa-vmss-windows-autoscale-existing-db/README.md
robe070: The reasons for not using over provisioning are provided in the paragraph above. Essentially, it breaks it and slows it down.

If the Image is the problem, then surely no instances will be created, so overprovisioning makes no difference.

If an instance gets provisioned but doesn't come up and respond to the probe, I see it get terminated and another instance is created. Quite annoying when debugging the probe to have the instance terminated whilst you are investigating the cause of the failed probe :) Of course I learnt to re-configure the probe so it appeared to be working.

So that seems to imply that a failed probe is not classified the same as a failed provision. Really? Lets say the instance count is 2. One succeeds the other fails. Obviously they are identical configurations and so there has been a failure of the infrastructure. Surely the non provisioned (and non-responsive probe) should be terminated?

gatneil: How did you see that behavior of the instance being terminated and re-created? That shouldn't happen. Yes, a failed probe is a separate type of problem from a failed provision. In the case of a successful provision but a failed probe, the VM won't receive traffic from the LB but will still run whatever code is on the VM (via extension, custom image, etc.). In the case of failed provision, the VM won't run code because it didn't come up successfully (and it won't receive traffic from the LB). In both cases, though, the VM won't be terminated. If overprovisioning was on, then it's possible that multiple VMs were added to the LB (even though you only asked for one), then one of them was terminated. But it shouldn't be the case that 1 VM fails provisioning then gets replaced. (if, however, one were to delete the failed VM, it's possible that autoscale would automatically bring it back up to 1 if the min was 1)

gatneil: Failed probe causing a VM to be terminated is a bug. That is a useful feature, but it's not one that scale sets support today. The VM is no longer sent traffic by the load balancer, but the VM doesn't get terminated.

Basically I'd like to see this sentence removed: "If it occurs, the VMSS will ensure another VM is created to establish the correct InstanceCount."

gatneil:
Looks good to me; merging

PR #2816, opened 2016-12-03
Vmss windows custom script

gatneil:


gatneil:
Requires existing resources, so CI fails; tested manually; merging

PR #2846, opened 2016-12-08
Added provider tag to all resources

sebastus:
added unique provider tag as guid to all resources. this allows all chef-related resources to be queried by this tag/guid. also fixed a bug with vm size default value.


sebastus:
#sign-off 

I think the travis check failed because this is a two step deployment. Some assets are created by prereq.json and then others are created using azureDeploy.json.

PR #2850, opened 2016-12-08
add provider tag to all resources

sebastus:
added unique provider tag to all resources so that all Chef resources on the platform can be found by searching for this tag value.


sebastus:
#sign-off 
@singhkays

PR #2874, opened 2016-12-12
Minor init.sh fixes for bigchaindb-on-ubuntu

ttmc:
There were some minor problems with the `init.sh` script for `bigchaindb-on-ubuntu`. This pull request fixes them:

1. The first line was `#!/bin/bash` but it seems the Azure Linux Agent runs `sh` regardless, so I changed that first line to `#!/bin/sh`
2. The line `source /etc/lsb-release && ...` wasn't working because `source` doesn't work with `sh`. The word `source` was replaced with just `.`
3. The last line was changed from `sudo pip install bigchaindb` to `sudo pip3 install bigchaindb`. This may not have been necessary but it seemed like a good idea to be explicit about which version of pip we want to use.

### DIFF
--- a/bigchaindb-on-ubuntu/scripts/init.sh
+++ b/bigchaindb-on-ubuntu/scripts/init.sh
@@ -1,4 +1,4 @@
-#!/bin/sh 
+#!/bin/sh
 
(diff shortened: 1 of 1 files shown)


azurecla:
Hi __@ttmc__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!

TTYL, AZPRBOT;

PR #2884, opened 2016-12-14
Fix pip3 encoding issues

tomconte:
pip3 sometimes fails with encoding errors b/c the LANG environment variable is set to "POSIX" when the init.sh is executed, at VM provisioning time. This change forces LANG to UTF-8 to fix the issue.

### DIFF
--- a/bigchaindb-on-ubuntu/scripts/init.sh
+++ b/bigchaindb-on-ubuntu/scripts/init.sh
@@ -9,8 +9,10 @@ rethinkdb --bind all --daemon
 
 sudo apt-get -y install make
 sudo apt-get -y install g++ python3-dev libffi-dev
-
 sudo apt-get -y install python3-setuptools
+
+export LANG="en_US.UTF-8"
+
 sudo easy_install3 pip
 sudo pip3 install --upgrade pip wheel setuptools
 


azurecla:
Hi __@tomconte__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <p>
        It looks like you're working at Microsoft (tconte). If you're full-time, we DON'T require a contribution license agreement.
    </p>
    <p>
    </p>

TTYL, AZPRBOT;

tomconte:
This is for the BigChainDB template BTW.

ttmc:
Thanks @tomconte — Hopefully this will resolve all the remaining issues with the `init.sh` script.

:+1:

gatneil:
looks good to me; merging

Jiycefer:
With this line export LANG="en_US.UTF-8" installation of  python-rapidjson fails:
    Complete output from command python setup.py egg_info:
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/tmp/pip-build-jindavfo/python-rapidjson/setup.py", line 43, in <module>
        with open('CHANGES.rst', encoding='utf-8') as f:
    FileNotFoundError: [Errno 2] No such file or directory: 'CHANGES.rst'
    
    ----------------------------------------
Command "python setup.py egg_info" failed with error code 1 in /tmp/pip-build-jindavfo/python-rapidjson/

Without this line - everything passing

tomconte:
@Jiycefer how did you run the script? We found it necessary when running in the context of the CustomScript extension.

Jiycefer:
@tomconte First one I tried to use the template – first deployment was good, without errors, after I tried to deploy one more instance (in another RG) and I was get the error like this: “VM Extension ProvisioningError” – error was in init.sh(the same error), after I’ve decided to create the clean VM and deploy manually your script and received the error concerning python-rapidjson/setup.py.

CustomScript Extension context - if i understand correctly with CustomScript - we have 2 context: protected launch and "standart launch" - i thinx it's a standart lunch because in init.sh we have sudo commands which by default not pretected commands....

ttmc:
I think I know what's going wrong: rapidjson updated recently, and that update broke things, since we weren't pinning the version of rapidjson.

Today or tomorrow, we will be releasing a new version (0.9.3) of BigchainDB Server, so that will become the latest version on PyPI (which is what `pip3 install bigchaindb` installs). That should fix this problem. You can follow our progress on GitHub. The bug/issue we're fixing is this one:

https://github.com/bigchaindb/bigchaindb/issues/1231

PR #2914, opened 2016-12-22
remove function app dependency

vinaysingh123:
Removed function app dependency for veter pipeline template.

PR #2915, opened 2016-12-22
Removed function app dependency

vinaysingh123:
Also changed api version from preview to GA.


vinaysingh123:
@gatneil this is similar to #2914 for a different template.

gatneil:
Looks good to me; merging

PR #2923, opened 2016-12-25
Update description for the RD Gateway high availability template

mmarch:


gatneil:
readme change only; merging

PR #3019, opened 2017-01-31
MD2

gatneil:
DON'T MERGE YET


gatneil:
only 1 template failed, and that's due to pre-reqs; ready to merge

PR #3028, opened 2017-01-31
MD6

gatneil:
DON'T MERGE YET


gatneil:
only 1 failure and that's on the extension, so the MD parts are fine

PR #3033, opened 2017-02-01
MD8

gatneil:
DON'T MERGE YET


gatneil:
there is only 1 failure, and that's an internal server error on the app gw, not the MD (the VMs with MD are provisioning properly)

PR #3041, opened 2017-02-01
MD11

gatneil:
DON'T MERGE YET


gatneil:
only 1 failure, and that's expected because it requires pre-created resources; tested manually; merging

PR #3075, opened 2017-02-10
MD list

gatneil:


gatneil:
@chagarw @bmoore-msft @singhkays I'd like to add this file temporarily while we get the templates updated to managed disk so customers can tell what is using managed disk or not without having to closely inspect the templates; once we have all the numbered ones using managed disk (except for ones with 'unmanaged' in the folder name), I'll remove this file. Sound good? :)

bmoore-msft:
Looks good - I might rename it to something more descriptive (e.g. managed-disk-support-list.md) since md generally means markdown around gh

gatneil:
Good call; renamed; merging

PR #3126, opened 2017-02-23
Small improvements to 201-vmss-ubuntu-web-ssl/keyvault.sh

ne-msft:
* Use a single sed call for all replacements.
* Add azuredeploy.parameters.json to .gitignore, as it should always be
  generated from azuredeploy.parameters.json.template
* Make base64 calls OSX compatible
* Show usage and exit whenever less than 6 arguments are passed
* Create a random 33-byte password for the certificate

### DIFF
--- a/201-vmss-ubuntu-web-ssl/keyvault.sh
+++ b/201-vmss-ubuntu-web-ssl/keyvault.sh
@@ -80,7 +80,7 @@ storesecret()
 {
     local secretfile=$1
     local name=$2
-    filecontentencoded=$( cat $secretfile | base64 -w 0 )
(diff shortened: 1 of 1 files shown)


azurecla:
Hi __@ne-msft__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <p>
        It looks like you're working at Microsoft (nierfurt). If you're full-time, we DON'T require a contribution license agreement.
    </p>
    <p>
    </p>

TTYL, AZPRBOT;

ne-msft:
This template fails validation because the parameters.json has placeholders which are replaced when calling a certain script (keyvault.sh) before deployment.

Is there a way to still pass validation in such a case?

PR #3274, opened 2017-03-30
made most vmss templates have ReadWrite caching on OS Disk for perf i…

gatneil:
…mprovmeents


gatneil:
@gbowerman as an FYI: this is the PR for the caching of OS disks on VMSS templates

gatneil:
simple change; tested manually, and the few CI tests that ran before timeout were successful; merging

gbowerman:
@gatneil thanks for fixing those.

PR #3284, opened 2017-04-03
Update azuredeploy.json

suvadip30:
### Best Practice Checklist

### Changelog

*
*
*

### DIFF
--- a/multi-vm-chef-template-ubuntu-vm/azuredeploy.json
+++ b/multi-vm-chef-template-ubuntu-vm/azuredeploy.json
@@ -1,291 +1,312 @@
 {
-  "$schema": "https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#",
-  "contentVersion": "1.0.0.0",
-  "parameters": {
-    "newStorageAccountName": {
-      "type": "string",
-      "metadata": {
-        "description": "Unique  Name for the Storage Account where the Virtual Machine's disks will be placed"
-      }
-    },
-    "vmDnsName": {
-      "type": "string",
-      "metadata": {
-        "description": "DNS  name for the VM"
-      }
-    },
-    "numberOfInstances": {
-      "type": "int",
-      "defaultValue": 2,
-      "metadata": {
-        "description": "Number of VM instances to create, default is 2"
-      }
-    },
-    "availabilitySetName": {
-      "type": "string",
-      "metadata": {
-        "description": "Availability Set name for the VMs"
-      }
-    },
-    "adminUserName": {
-      "type": "string",
-      "metadata": {
-        "description": "Admin user name for the Virtual Machines"
-      }
-    },
-    "adminPassword": {
-      "type": "securestring",
-      "metadata": {
-        "description": "Admin password for the Virtual Machines"
-      }
-    },
-    "imagePublisher": {
-      "type": "string",
-      "defaultValue": "Canonical",
-      "metadata": {
-        "description": " Publisher for the OS image, the default is Canonical"
-      }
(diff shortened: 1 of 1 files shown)


suvadip30:
Can you please give me the update.

multi-vm-chef-template-ubuntu-vm/azuredeploy.json
```diff
@@ -1,291 +1,312 @@
...
-      }
-    },
-    "runlist": {
-      "type": "string",
-      "defaultValue": "recipe[getting-started]",
-      "metadata": {
-        "description": "Optional Run List to Execute"
-      }
-    },
-    "validation_key": {
```
NimishaS: `validation_key` should not removed as it needs to be passed to the chef extension: https://github.com/Azure/azure-quickstart-templates/pull/3284/files?diff=split#diff-17812ea6afe06b6ea8eabb2b908d497fL286

The extension will fail without this.

suvadip30: It need to pass and hard-coded in virtual-extensions, so no need to pass as a parameter, you can check the line number 304 in edited script. When we are passing from parameter it is ending up in failing in registering with Chef server as it is not able to read the format of validation key.

I have tested many times and it is working good.

multi-vm-chef-template-ubuntu-vm/azuredeploy.json
```diff
@@ -1,291 +1,312 @@
...
+                }
```
NimishaS: @suvadip30 , private keys should not be exposed like this. I hope this is not your actual validation key.

multi-vm-chef-template-ubuntu-vm/azuredeploy.json
NimishaS: `validation_key` is going to differ for every user and hence it makes sense to pass it as a parameter.

Beauty of these JSON templates is that, all the parameters and variables can be replaced by the hardcoded values, and the template will still work. But that's not a good practice.

multi-vm-chef-template-ubuntu-vm/azuredeploy.json
suvadip30: Hi Nimisha

From portal I tried with the reference still it is not working, that is the reason I tried in hard-coded the validation-key and it worked well.

Yes, I can understand that every validation-key are different for different users, but every users need to go and hard-code the validation key, it is not working from passing to the parameters.

I tried #102 issue, the pointed reference link, but still facing the same issue.
https://github.com/Azure/azure-quickstart-templates/blob/master/chef-json-parameters-linux-vm/azuredeploy.json#L69

Please refer me any solution to pass the validation-key from parameter.

** No this is not my validation_key, this is a dummy one as it is refered by Chef https://docs.chef.io/azure_portal.html

Thanks
-Suvadip

NimishaS: @suvadip30 , I got your issue. When `validation_key` is being passed from the parameters in `plaintext` format, all the new lines need to be replaced by `\n` and a single line string needs to be passed. This was not very user friendly and that's why we had added support to pass `validation_key` in `base64encoded` format as you pointed  https://github.com/Azure/azure-quickstart-templates/blob/master/chef-json-parameters-linux-vm/azuredeploy.json#L69.

Please try out replacing new lines with `\n`. It will work.
Hardcoding is not recommended and the quickstart templates shouldn't have it since it's referred by several people.

bmoore-msft:
Closing due to no activity

PR #3304, opened 2017-04-07
Added bash script that uses CLI 2.0 to deploy ARM templates.

rickrain:


singhkays:
@bmoore-msft You'll be interested in reviewing this. CLI equivalent to your PS script :)

@rickrain neat change!

bmoore-msft:
@rickrain - I'll catch up with you offline and talk through this...

bmoore-msft:
Will continue conversation offline

PR #3305, opened 2017-04-09
Removing utf8 bom

Nepomuceno:
### Changelog

* Removed the uft8 BOM from the current files.


Nepomuceno:
The build it is failing because of the time it takes. to test every deployment

Nepomuceno:
Closed by mistake :(

PR #3349, opened 2017-04-18
Update to GA ACR for devops templates

EricJizbaMSFT:
Original PR by @mekenthompson here: https://github.com/Azure/azure-quickstart-templates/pull/3316, but he had some issues with rebasing/formatting. I've applied the original commit to all three of our templates that use ACR. Thanks for the contribution Ken! (I've listed you as the author for this commit)


mekenthompson:
Thanks for fixing up @EricJizbaMSFT! - FYI I had a rogue extension that was reformatting on save in VS Code, have removed it.

PR #3351, opened 2017-04-19
azure-jenkins and readme updates

jonbgallant:
### Best Practice Checklist

### Changelog

* Updated readme to be ubuntu/windows friendly
* Update azure-jenkins to include 8080/http and unique names
* Updated azure-group-deploy to respect -v


README.md
```diff
@@ -14,27 +14,74 @@ You can deploy these samples directly through the Azure Portal or by using the s
...
 ```
+
+### Bash
 ```bash
 azure-group-deploy.sh -a [foldername] -l eastus
 ```
+
+### Bash on Ubuntu on Windows
+
+Run the following the first time to get all prerequisites installed:
```
bmoore-msft: shouldn't the prereq install be doc'd on ms.com?  (we don't list the PowerShell prereqs)

README.md
```diff
@@ -14,27 +14,74 @@ You can deploy these samples directly through the Azure Portal or by using the s
...
+sudo apt install jq
+dos2unix azure-group-deploy.sh
+chmod +x azure-group-deploy.sh
+azure login
+azure account set [subscriptionNameOrId]
+```
+
+Run the following to execute the script:
+```bash
+./azure-group-deploy.sh -a [foldername] -l [location] -g [resourcegroup] -e azuredeploy.parameters.json
```
bmoore-msft: remove this - it's covered below in a simpler/default format

README.md
```diff
@@ -14,27 +14,74 @@ You can deploy these samples directly through the Azure Portal or by using the s
...
+### Bash
+```bash
+azure-group-deploy.sh -a [foldername] -l eastus -u
+```
+
+### Bash on Ubuntu on Windows
+
 ```bash
-azure-group-deploy.sh -a '201-vm-custom-script-windows' -l eastus -u
+./azure-group-deploy.sh -a [foldername] -l [location] -g [resourcegroup] -e azuredeploy.parameters.json -u
```
bmoore-msft: why a different use case for windows?

azure-group-deploy.sh
```diff
@@ -106,7 +106,7 @@ azure group create "$resourceGroupName" "$location"
 # Remove line endings from parameter JSON so it can be passed in to the CLI as a single line
 parameterJson=$( echo "$parameterJson" | jq -c '.' )
 
-if [[ -z $validateOnly ]]
+if [[ $validateOnly ]]
```
bmoore-msft: Let's make this consistent with the upload if statement (simpler is fine, just make them the same)

azure-jenkins/azuredeploy.json
```diff
@@ -18,13 +18,14 @@
...
       "minLength": 1,
+      "maxLength": 11,
       "metadata": {
         "description": "Globally unique DNS Name for the Public IP used to access the Virtual Machine."
       }
     }
   },
   "variables": {
-    "resourcePrefix": "jenkins",
+    "resourcePrefix": "[parameters('jenkinsDnsPrefix')]",
```
bmoore-msft: if we're just assigning the param value to the var - can we just remove the var?  Seems unnecessary then...

README.md
jonbgallant: These are pre-reqs for azure-group-deploy.sh in this repo.

README.md
jonbgallant: I do not see where all that is listed below. Please point me to what you are talking about.

README.md
bmoore-msft: Other than jq I think everything is covered here https://docs.microsoft.com/en-us/azure/cli-install-nodejs - we don't list these for PowerShell... 
Also, not sure why dos2unix is in there, I've never seen that needed (though happy to be educated).

README.md
bmoore-msft: example of how to run the script is given on lines 74 and 80

README.md
jonbgallant: I'm looking for lowest dev friction here as possible.

1. That link does not show how to install npm or node, which are not installed in Linux subsystem by default. There's a link on that page, but now we are 3 links deep into node install and their install page scrolls forever.  It doesn't hurt to include the exact steps here. Without the instructions inline you will increase the developer start time by at least 5 minutes probably more.

2. I had never heard of dos2unix before either, but without running it I get the following error (on Linux Subsystem). 

If you see the following error: "syntax error near unexpected token `$'in\r''", make sure you run the dos2unix command above.

README.md
jonbgallant: I see, you are talking about the azure-group-deploy call. I prefer the format I'm using because it does a better job explaining the options that a developer will likely use when running the script locally.  They will want to set the resource group (-g) and they will likely want to override the default parameters (-e)

README.md
jonbgallant: I'd rather see us change the PowerShell and Bash versions to include the -g and -e parameters.  See reasoning in previous comment.

azure-group-deploy.sh
jonbgallant: Got it.  Will be updated in a min.

azure-jenkins/azuredeploy.json
jonbgallant: Great point.  Fixing...

README.md
bmoore-msft: Understood - but the short default is intentional based on early feedback...  So we want the minimal command to be that default - the readme should be terse.  We can expand and explain but that should be somewhere other than the readme (which is fine).

README.md
bmoore-msft: the other versions of the script do include those options, but we use the defaults (per above)

PR #3352, opened 2017-04-19
Standardize devops template ssh commands

EricJizbaMSFT:
Use one-liner instead of ssh-tunneling
Make the output variable the same name so we can use it for CI


clguimanMSFT:
Looks good, but you should also add a Windows option:
 putty.exe -ssh -L 8080:localhost:8080 <username>@<domain name>

PR #3367, opened 2017-04-21
Md13

gatneil:
Update for managed disk; fixed formatting as well


gatneil:
CI will fail because requires custom image; tested manually

PR #3369, opened 2017-04-21
update to use MD

gatneil:


gatneil:
will fail CI because requires existing resources; tested manually :)

PR #3372, opened 2017-04-22
Revert changes to azure-jenkins template

EricJizbaMSFT:
From commit 3a47045d5cb212907fdbba8314949e85448fc55a and 67b9bca3f2292ac3d310ce3e0235841afd48f5d1

These changes broke the template for a few reasons:
1. We don't want port 8080 exposed in the NSG because Jenkins uses "http" by default. If someone logs in to Jenkins over their public IP address, their password and other information could be stolen through a man-in-the-middle attack. The only secure way to log in is with ssh tunneling (until the Jenkins admin sets up https with their own certificate)
2. If the error says `bind:address is already in use`, that means something else is using that port on your local machine. The fix is to close whatever is using port 8080 on your localhost (maybe you have another ssh tunnel going in the background?) or to use a different port: `ssh -L 8082:localhost:8080 <User name>@<Public DNS name of instance you just created>`
3. The jenkinsDnsPrefix should not be used as the resource prefix for all resources because it has different naming constraints. For example a dns prefix can be up to 63 characters long and can contain hyphens, but a storage account name can only be up to 24 characters long and _cannot_ contain hyphens

### DIFF
--- a/azure-jenkins/README.md
+++ b/azure-jenkins/README.md
@@ -38,13 +38,6 @@ Run this command:
 ssh -L 8080:localhost:8080 <User name>@<Public DNS name of instance you just created>
 ```
 
-If you see the following error: "bind:address is already in use", then you need to include the IP Address as follows:
-
-```
-ssh -L localhost:8080:<Public IP>:8080 <User name@<Public IP>
-```
-
-
 ## C. Configure Sample Jobs and Azure Active Directory configuration
 1. Once you are logged into the VM, run /opt/azure_jenkins_config/config_azure.sh and pick option 1 - "All of the below". This script will guide you to set up and configure the Azure Storage plugin to be used in the sample jobs to upload and download to Storage.
 It will also provide a Service Principal to access Azure resources from Jenkins.
--- a/azure-jenkins/azuredeploy.json
+++ b/azure-jenkins/azuredeploy.json
@@ -18,23 +18,22 @@
     "jenkinsDnsPrefix": {
       "type": "string",
       "minLength": 1,
-      "maxLength": 11,
       "metadata": {
         "description": "Globally unique DNS Name for the Public IP used to access the Virtual Machine."
       }
     }
   },
   "variables": {
-    "resourcePrefix": "[parameters('jenkinsDnsPrefix')]",
-    "storageAccountName": "[concat(parameters('jenkinsDnsPrefix'), uniquestring(resourceGroup().id))]",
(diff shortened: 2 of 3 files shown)


EricJizbaMSFT:
@jonbgallant FYI

jonbgallant:
1. The NSG still needs 8080 open, even with a SSH port forwarding.  Try deleting 8080 in azure, open ssh with port forward and then hit 127.0.0.1:8080 and see if it works for you. Make sure you clear all port forward caches locally, like you are starting from a fresh machine.

2. This is a Windows Bash known issue, that has been fixed in Creators https://superuser.com/questions/1119836/ssh-tunnel-via-native-windows-bash-address-already-in-use

3. The current deployment uses a hardcoded "jenkins" as a prefix, which prevents people from deploying additional azure-jenkins deployments to the same resource group. Used the jenkinsDnxPrefix to solve this. Let me know if you have other suggestions.

EricJizbaMSFT:
Yes, I can connect even without the 8080 rule in the NSG. I think you need 8080 because you're running this command: `ssh -L localhost:8080:<Public IP>:8080 <User name@<Public IP>`

But if you're ssh-ing into the VM, you can just use localhost. If I read the bug right, the workaround is to specify the full bind address, which means this might work for you: `ssh -L 127.0.0.1:8080:127.0.0.1:8080 <User name@<Public IP>`

If that didn't get around the bash on ubuntu bug let me know and I can help find a workaround that doesn't require exposing 8080 in the NSG.

In terms of the naming - what is your use case for multiple deployments in the same resource group? We don't see that scenario very often and would be interested to understand why you're doing that. However, we _do_ see people name resources incorrectly and then their deployment fails, which is why we use a resource prefix. Some excerpts from the [best practices doc](https://github.com/Azure/azure-quickstart-templates/blob/master/1-CONTRIBUTION-GUIDE/best-practices.md) on this topic:
>Name variables using this scheme templateScenarioResourceName (e.g. simpleLinuxVMVNET, userRoutesNSG, elasticsearchPublicIP etc.) that describe the scenario

>Do not create a parameter for a storage account name. Storage account names need to be lower case and can't contain hyphens (-) in addition to other domain name restrictions.

If you want to go against the best practices because it works better in your specific scenario that is totally fine, but then we recommend forking the template yourself. We want this template to be consistent with all other templates in the repo.

arroyc:
@jonbgallant  .. can you please explain your use case? why do you want to use the same resource groups? Usually these quickstart templates are designed for deployment per resourcegroup. If you have a special use case, let us know, we can either change this one or can introduce a new one with custom fit.

PR #3398, opened 2017-04-27
Fix deployment of 201-encrypt-create-new-vm-gallery-image - need to u…

StephenWeatherford:
Doesn't find storage account if HTTP is used

- [X ] - Please check this box once you've submitted the PR if you've read through the Contribution Guide and best practices checklist.

### Changelog

* HTTP -> HTTPS (doesn't find storage account if HTTP)
* newStorageAccount -> storageAccount (since it requires an existing one)

### DIFF
--- a/201-encrypt-create-new-vm-gallery-image/azuredeploy.json
+++ b/201-encrypt-create-new-vm-gallery-image/azuredeploy.json
@@ -175,7 +175,7 @@
           "osDisk": {
             "name": "[variables('osDiskName')]",
             "vhd": {
-              "uri": "[concat('http://',parameters('newStorageAccountName'),'.blob.core.windows.net/',parameters('vmStorageContainerName'),'/',variables('osDiskName'),'.vhd')]"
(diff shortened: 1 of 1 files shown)


201-encrypt-create-new-vm-gallery-image/azuredeploy.json
```diff
@@ -175,7 +175,7 @@
           "osDisk": {
             "name": "[variables('osDiskName')]",
             "vhd": {
-              "uri": "[concat('http://',parameters('newStorageAccountName'),'.blob.core.windows.net/',parameters('vmStorageContainerName'),'/',variables('osDiskName'),'.vhd')]"
+              "uri": "[concat('https://',parameters('newStorageAccountName'),'.blob.core.windows.net/',parameters('vmStorageContainerName'),'/',variables('osDiskName'),'.vhd')]"
```
EricJizbaMSFT: Best practice is to dynamically retrieve the endpoint rather than hardcode any part of the url. Something like this:

[concat(reference(resourceId('Microsoft.Storage/storageAccounts', variables('newStorageAccountName')), '2016-01-01').primaryEndpoints.blob, variables('vmStorageContainerName'), '/', variables('osDiskName'), '.vhd')]

PR #3427, opened 2017-05-06
Multivmssbump

gatneil:


gatneil:
CI seems to be down (subscription disabled); tested manually; merging

PR #3453, opened 2017-05-12
Upgrading all Jenkins scripts to azure-devops-utils v0.14.0

clguimanMSFT:
v0.14.0 also installs Git and the Azure CLI on the Jenkins VM


clguimanMSFT:
@ArchangelSDY FYI

clguimanMSFT:
I've updated the commit to include the metadata files

PR #3457, opened 2017-05-14
Create Storage Template

hemantpro:
Using this template user can create different storage skus


bmoore-msft:
I think we have storage covered in another of other samples

PR #3531, opened 2017-06-07
Updated to use new VPN GW SKUs

bhummerstone:
Updated the 201-vnet-to-vnet-bgp template to include new VPN SKUs (VpnGw1/2/3)


bhummerstone:
Not sure why the Travis check failed: I've deployed this, and Test-AzureRMResourceGroupDeployment reports that the template is valid

PR #3559, opened 2017-06-15
Update Spinnaker templates

EricJizbaMSFT:
* Deprecate 201-spinnaker-acr-k8s
* Update templates to leverage Halyard
* Reduce size of VMs so that default cost is lower


101-spinnaker/azuredeploy.json
```diff
@@ -32,7 +32,8 @@
     "vmName": "[concat(variables('resourcePrefix'), 'VM')]",
     "vmExtensionName": "[concat(variables('resourcePrefix'), 'Init')]",
     "virtualNetworkName": "[concat(variables('resourcePrefix'), 'VNET')]",
-    "_artifactsLocation": "https://raw.githubusercontent.com/Azure/azure-devops-utils/v0.12.0/",
+    "_artifactsLocation": "https://raw.githubusercontent.com/Azure/azure-devops-utils/hal/",
```
EricJizbaMSFT: TODO: change 'hal' to the new version of the utils repo once released

PR #3572, opened 2017-06-21
Updating Chef quickstart-templates

stuartpreston:
Signed-off-by: Stuart Preston <stuart@chef.io>

### Changelog

* Adding providerGuid as a tag on all resources
* Removing Chef Automate solution that took a dependency on a decomissioned Chef Server


chef-automate-ha/azuredeploy.json
```diff
@@ -252,7 +252,8 @@
         "felbPublicIPAddressID": "[resourceId('Microsoft.Network/publicIPAddresses',variables('networkSettings').felbPublicIPAddressName)]",
         "chefAutoPublicIPAddressID": "[resourceId('Microsoft.Network/publicIPAddresses',variables('networkSettings').chefAutoPublicIPAddressName)]",
         "deploymentApiVersion": "2015-01-01",
-        "prefix": "[uniqueString(resourceGroup().id)]"
+        "prefix": "[uniqueString(resourceGroup().id)]",
+        "provider": "[toUpper('33194f91-eb5f-4110-827a-e95f640a9e46')]"
```
bmoore-msft: Just to confirm - this a universal guid we want all users to use?

chef-automate-ha/nested/availabilitySetsResource.json
```diff
@@ -6,7 +6,9 @@
             "type": "object"
         }
     },
-    "variables": {},
+    "variables": {
+        "provider": "[toUpper('33194f91-eb5f-4110-827a-e95f640a9e46')]"
```
bmoore-msft: if this is the same guid as used in the main template, why wouldn't we pass it through as a parameter instead of re-defining in each template?

chef-automate-ha/nested/be-networkInterfacesResource.json
```diff
@@ -9,7 +9,9 @@
             "type": "string"
         }
     },
-    "variables": {},
+    "variables": {
+        "provider": "[toUpper('33194f91-eb5f-4110-827a-e95f640a9e46')]"
```
bmoore-msft: ditto for all the other files we define it in...

chef-automate/README.md
```diff
@@ -1,260 +0,0 @@
-# Chef - Cloud Devops Solution Template
```
bmoore-msft: Ok from the PR comments looks like we just want to remove this sample - but the diff looks like we're just emptying the files (not deleting them)???

chef-automate-ha/azuredeploy.json
stuartpreston: Yes, this is a tracking device that our DX Liaison @sebastus has asked us to put in all our templates.

chef-automate-ha/nested/availabilitySetsResource.json
stuartpreston: Simply it shouldn't be able to be parameterized and there shouldn't be any customization of it.

chef-automate/README.md
stuartpreston: I think that's just how it looks in the diff?  The files are deleted if you navigate to the branch.

PR #3582, opened 2017-06-23
Added Service Bus Topic Subscription w/ SQL Filter template

crpietschmann:
This ARM Template is a further addition to a previous one I submitted that extends it to add the ability to set up a SQL Filter on the Azure Service Bus Topic Subscription that gets created. This is something that isn't documented ANYWHERE, but I was able to figure it out, so I thought I'd share it in the quick-start-templates project so it can help others too. :)

### DIFF
--- a/101-servicebus-topic-subscription-sqlfilter/azuredeploy.json
+++ b/101-servicebus-topic-subscription-sqlfilter/azuredeploy.json
@@ -1,0 +1,83 @@
+{
+  "$schema": "https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#",
+  "contentVersion": "1.0.0.0",
+  "parameters": {
+    "serviceBusNamespaceName": {
+      "type": "string",
+      "metadata": {
+        "description": "Name of the Service Bus Namespace"
+      }
+    },
+    "serviceBusTopicName": {
+      "type": "string",
+      "metadata": {
+        "description": "Name of the Service Bus Topic"
+      }
+    },
+    "serviceBusTopicSubscriptionName": {
+      "type": "string",
+      "metadata": {
+        "description": "Name of the Service Bus Topic Subscription"
+      }
+    },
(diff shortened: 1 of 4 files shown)


101-servicebus-topic-subscription-sqlfilter/azuredeploy.json
```diff
@@ -0,0 +1,83 @@
+{
+  "$schema": "https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#",
+  "contentVersion": "1.0.0.0",
+  "parameters": {
+    "serviceBusNamespaceName": {
```
bmoore-msft: are these all globally unique or can we add default values?

101-servicebus-topic-subscription-sqlfilter/azuredeploy.json
```diff
@@ -0,0 +1,83 @@
...
+    },
+	"serviceBusTopicSubscriptionSqlFilter": {
+      "type": "string",
+      "metadata": {
+        "description": "The SQL Filter of the Service Bus Topic Subscription"
+      }
+    }
+  },
+  "variables": {
+    "sbVersion": "2015-08-01"
```
bmoore-msft: use a literal apiVersion so schema validation can be done

101-servicebus-topic-subscription-sqlfilter/azuredeploy.json
```diff
@@ -0,0 +1,83 @@
...
+                {
+                    "apiVersion": "[variables('sbVersion')]",
+                    "name": "[parameters('serviceBusTopicSubscriptionName')]",
+                    "type": "Subscriptions",
+                    "dependsOn": [
+                        "[parameters('serviceBusTopicName')]"
+                    ],
+                    "properties": {
+                    },
+                    "resources": [{
```
bmoore-msft: Can you format the document to clean up the json?

PR #3611, opened 2017-07-03
 201-key-vault-secret-create | propsoing secureObject wrapper for array 

danielstocker:
adding secureObject wrapper to align more with ARM templating recommendations: https://docs.microsoft.com/en-us/azure/azure-resource-manager/resource-group-authoring-templates#parameters

### Best Practice Checklist

[x] Please check this box once you've submitted the PR if you've read through the Contribution Guide and best practices checklist.

### Changelog

* changed variable input to secureObject rather than array
* added object wrapper around array
* changed reference expressions in template to support wrapper
* using a secureObject prevents other users from seeing the secrets in the portal after the deployment

### Screenshots of the difference

Before: (everyone can see the secrets)
secret-value-before-wrapper

After: (secrets hidden)
using-secret-wrapper

### DIFF
--- a/201-key-vault-secret-create/azuredeploy.json
+++ b/201-key-vault-secret-create/azuredeploy.json
@@ -53,7 +53,7 @@
                 "description": "Specifies if the vault is enabled for volume encryption"
             }
         },
-        "secureObject": {
+        "secretsObject": {
             "type": "secureObject",
             "defaultValue": "{}",
             "metadata": {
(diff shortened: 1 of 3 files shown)


201-key-vault-secret-create/azuredeploy.json
```diff
@@ -54,10 +54,10 @@
             }
         },
         "secrets": {
-            "type": "array",
+            "type": "secureObject",
```
bmoore-msft: I like this but it took me a while to wrap my head around what you're doing given the naming convention...  Can we use "secretObject" for the parameter name and "secrets" for the objectName?

danielstocker: Hi Brian, yeah I agree it makes more sense if we call the wrapper "secureObject" and the array in it "secrets". I have made these adjustments now.

201-key-vault-secret-create/azuredeploy.json
```diff
@@ -53,11 +53,11 @@
                 "description": "Specifies if the vault is enabled for volume encryption"
             }
         },
-        "secrets": {
-            "type": "array",
+        "secureObject": {
```
bmoore-msft: let's just call this "secretsObject" to avoid confusion with the typeName...

danielstocker: okay yeah that's a fair point

PR #3620, opened 2017-07-05
Add templates: autoscale alerts

asheniam:
Adding new ARM templates for activity log alerts for autoscale record


gatneil:
Looks good to me; merging :)

PR #3695, opened 2017-07-16
CDN API version up.

SychevIgor:
CDN API version up.  CDN sku listed

### DIFF
--- a/101-cdn-with-custom-origin/azuredeploy.json
+++ b/101-cdn-with-custom-origin/azuredeploy.json
@@ -19,10 +19,22 @@
       "metadata": {
         "description": "Url of the origin"
       }
+    },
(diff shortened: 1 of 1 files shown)


101-cdn-with-custom-origin/azuredeploy.json
```diff
@@ -19,53 +19,66 @@
...
-            "properties": {
-              "hostName": "[parameters('originUrl')]"
-            }
+      "sku": {
+        "name": "[parameters('CDNSku')]"
+      },
+      "resources":
+      [
+        {
+          "apiVersion": "[variables('cdnApiVersion')]",
```
bmoore-msft: use a literal value for apiVersion so schema validation can be done

PR #3826, opened 2017-08-25
Removed unsupported template

yossidahan:
Removed app-service-environment-serviceplan-logicapp as per note from
Kevin Lam thi sis no longer supported


bmoore-msft:
opened another PR and removed it

PR #3953, opened 2017-09-28
updated readme

spelluru:
Added some details to the readme.

### DIFF
--- a/101-data-factory-v2-blob-to-blob-copy/azuredeploy.parameters.json
+++ b/101-data-factory-v2-blob-to-blob-copy/azuredeploy.parameters.json
@@ -3,7 +3,7 @@
   "contentVersion": "1.0.0.0",
   "parameters": {
     "dataFactoryName": {
-      "value": "ADFTutorialFactory0928"
+      "value": "GEN-UNIQUE"
     },    
     "dataFactoryLocation": {
       "value": "East US"
--- a/101-data-factory-v2-blob-to-blob-copy/azuredeploy.parameters.json
+++ b/101-data-factory-v2-blob-to-blob-copy/azuredeploy.parameters.json
@@ -2,6 +2,9 @@
   "$schema": "https://schema.management.azure.com/schemas/2015-01-01/deploymentParameters.json#",
   "contentVersion": "1.0.0.0",
   "parameters": {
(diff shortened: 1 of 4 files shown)


spelluru:
@bmoore-msft - I am back with some changes:). 

1. Added a script in the scripts folder that customers can use to run and monitor the pipeline created by the template. 
2. Updated the readme document to include instructions for downloading and running the script. 
3. azuredeploy.json - Removed the "dataFactoryName" variable. Created the "dataFactoryName" as a parameter. Customer enters this name while deploying the template. 
4. azuredeploy-parameters.json - Updated the parameter file to pass value to the dataFactoryName parameter. 

This reduces a step for customers to get the data factory name from the portal blades.

101-data-factory-v2-blob-to-blob-copy/azuredeploy.parameters.json
```diff
@@ -2,6 +2,9 @@
   "$schema": "https://schema.management.azure.com/schemas/2015-01-01/deploymentParameters.json#",
   "contentVersion": "1.0.0.0",
   "parameters": {
+    "dataFactoryName": {
+      "value": "ADFTutorialFactory0928"
```
bmoore-msft: use GEN-UNIQUE

spelluru: @bmoore-msft - I thought about using it, but then customer will have to go through the following steps to find the name of the data factory: 

1. Click the successful deployment message. 
2. Launch resource group.
3. Search for the new data factory from the list of data factories. 

If this is must, I will go back to the previous where the data factory will have a prefix: ADFTutorialFactory. 

Please let me know. Thanks.

spelluru: How can I make customers enter a name for it so that they don't need to go through the blades to find the name of the data factory? Thanks.

spelluru: Thanks. Updated the file. Uploading it now.

spelluru:
@bmoore-msft - Done. Thanks.

PR #3993, opened 2017-10-12
MS Dynamics NAV

tfenster:
With this quickstart template you can deploy a Dynamics NAV container

### Best Practice Checklist

### Changelog

*
*
*

### DIFF
--- a/101-aci-dynamicsnav/README.md
+++ b/101-aci-dynamicsnav/README.md
@@ -1,0 +1,9 @@
+# Azure Container Instances
+
+This template demonstrates how you can run MS Dynnamics NAV in [Azure Container Instances](https://docs.microsoft.com/en-us/azure/container-instances/). To find out more about Dynamics NAV inside a Windows Container visit [Github](https://github.com/microsoft/nav-docker)
+
+To start the instance, you currently need to provide the password for the private Docker registry where the Dynamics NAV images are stored. That should change to Docker Hub soon. You also need to accept the end user license agreement by setting the param acceptEula to Y
+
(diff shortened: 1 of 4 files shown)


bmoore-msft:
@tfenster can you check the CI failures?

tfenster:
The first failure seems to be because some name ist too long. But as there are things like "101-application-gateway-public-ip-ssl-offload" I am not sure which name is the problem here?

The second failure is because I am using a private repo set up by Freddy Kristiansen from Microsoft. I am not the owner and therefore won't share it as I didn't find a way to encrypt it without tying it to my registry. Or did I misunderstand how the key vault works? Or I could ask him if he would want to share it with you but I am not sure if you would be willing to support that kind of scenario where automated CI will always fail?

bmoore-msft:
The first is referring to metadata.json - the properties in that file have maxlengths (you can read in that schema or the readme for the repo).

re: the second, the image needs to be public (or accessible) no point in having a sample no one can deploy...

tfenster:
Ok, I've fixed the maxlength problem, thanks for pointing me in the right direction. 

For the second problem, I would think that everyone interested in this specific sample should have the password, but I get your point. The images should become publicly available in the next weeks and I'll then get back and update the sample. Should I close the PR until then or can I leave it open?

bmoore-msft:
Let's close it for now and the reopen when the images are available to the public

tfenster:
Good timing, the images were made available publicly today and as you can see, build succeeds now as well :)

tfenster:
Hi @bmoore-msft, is there anything I need to change in addition to the build passing?

PR #4008, opened 2017-10-18
New template for SQL Server 2016 AOAG across AZ's using ILB for AG listener

robotechredmond:
### Best Practice Checklist

### Changelog

* New template for SQL Server 2016 AlwaysOn AG deployed across Azure Availability Zones with new cross-zone Azure ILB for AG listener - this is a different scenario from the existing template which deploys across Availability Zones using a multi-subnet listener - both scenarios are valid for different customer use case scenarios.
*
*


sql-alwayson-md-ilb-zones/.ci_skip
```diff
@@ -0,0 +1,2 @@
+Skip CI due to existing resource requirements
```
bmoore-msft: See the readme in the repo for how to handle existing resources/prereqs...  I'd like to see if we can get this validated in CI...

robotechredmond: This approach will be challenging for this template, because an existing Active Directory domain environment is also required - the prereq's folder would need to include the equivalent of an entire quickstart template for deploying Windows Server AD domain controllers.  We've already successfully tested this SQL template with several deployments in multiple regions, so the template itself is solid. Can we approve as-is?

bmoore-msft: Let's provide some instructions in the readme on how to get a successful deployment...

robotechredmond: I've updated the README with the instructions needed to deploy the prerequisites.  Is this sufficient?

bmoore-msft: Let's add the prereq templates - the az zones templates are simple enough that it should work fine (unless the deployment takes too long) - we're going to start automating deployment and reporting status and ci_skip will make it show as "not tested"

robotechredmond: Would it be acceptable to have the prereq template call the azuredeploy.json for the existing AD-zones Quick Start template, rather than copying the entire AD-zones template into the prereq folder of this template?

robotechredmond: @bmoore-msft Could you comment on the approach of linking to the existing AD zones template in the Quick Start gallery as a prereq vs copying the entire AD zones template into the prereq folder of this template?  Is linking an acceptable approach?  We just trying to avoid create multiple copies of what is essentially the same template code that then increases maintenance over time.

bmoore-msft: The engine doesn't support that right now, and I'm not sure we will since a change in one of those templates could break dependencies... since the prereqs are static, copying should be fine.

bmoore-msft:

PR #4038, opened 2017-10-25
Fix location casing

pplu:
These templates are using "Location" instead of "location", which is the documented key.

Is ARM templates supposed to accept "Location", or does it fail?


bmoore-msft:
camelCase is the JSON norm, so good PR

PR #4040, opened 2017-10-25
delete undefined, undocumented scale property

pplu:
Deleting the undefined, undocumented scale property


bmoore-msft:
@pplu - can you check the CI failures, looks like the JSON is not valid

pplu:
Hi,

  I don't understand why the tests are failing... The changes I've done to the fails shouldn't be causing syntax errors... Can you help me out?

PR #4047, opened 2017-10-27
Adding AKS ARM template example

lenadroid:
Adding AKS ARM template example


lenadroid:
Closing, noticed there's already an AKS pull request pending

PR #4106, opened 2017-11-30
Add the ability to create multiple access policies with the template.

sandshadow:
### Best Practice Checklist

### Changelog

* Add the ability to create multiple access policies with the template.
* Return the DataAccessFQDN instead of DataAccessId as output property, as it's more useful.

### DIFF
--- a/201-timeseriesinsights-environment-with-eventhub/azuredeploy.json
+++ b/201-timeseriesinsights-environment-with-eventhub/azuredeploy.json
@@ -91,6 +91,20 @@
       "metadata": {
         "description": "The name of the shared access key that the Time Series Insights service will use to connect to the event hub."
       }
+    },
+    "accessPolicyReaderObjectIds": {
+      "type": "array",
+      "defaultValue": [],
+      "metadata": {
+        "description": "A list of object ids of the users or applications in AAD that should have Reader access to the environment. The service principal objectId can be obtained by calling the Get-AzureRMADUser or the Get-AzureRMADServicePrincipal cmdlets. Creating an access policy for AAD groups is not yet supported."
+      }
+    },
+    "accessPolicyContributorObjectIds": {
+      "type": "array",
+      "defaultValue": [],
+      "metadata": {
+        "description": "A list of object ids of the users or applications in AAD that should have Contributor access to the environment. The service principal objectId can be obtained by calling the Get-AzureRMADUser or the Get-AzureRMADServicePrincipal cmdlets. Creating an access policy for AAD groups is not yet supported."
+      }
     }
   },
   "variables": {
@@ -194,16 +208,48 @@
           },
           "tags": "[variables('eventSourceTagsValue')]",
           "dependsOn": [
-            "[concat('Microsoft.TimeSeriesInsights/environments/',parameters('environmentName'))]",
+            "[concat('Microsoft.TimeSeriesInsights/environments/', parameters('environmentName'))]",
             "[concat('Microsoft.EventHub/namespaces/', parameters('eventHubNamespaceName'), '/eventHubs/', parameters('eventHubName'), '/consumerGroups/', parameters('consumerGroupname'))]"
           ]
         }
+      ]
+    },
+    {
+      "type": "Microsoft.TimeSeriesInsights/environments/accesspolicies",
+      "name": "[concat(parameters('environmentName'), '/', 'readerAccessPolicy', copyIndex())]",
+      "copy": { 
+        "name": "accessPolicyReaderCopy", 
+        "count": "[length(parameters('accessPolicyReaderObjectIds'))]"
+      },
+      "apiVersion": "2017-11-15",
+      "properties": {
+        "principalObjectId": "[parameters('accessPolicyReaderObjectIds')[copyIndex()]]",
+        "roles": ["Reader"]
+      },
+      "dependsOn": [
+        "[concat('Microsoft.TimeSeriesInsights/environments/', parameters('environmentName'))]"
+      ]
+    },
+    {
+      "type": "Microsoft.TimeSeriesInsights/environments/accesspolicies",
+      "name": "[concat(parameters('environmentName'), '/', 'contributorAccessPolicy', copyIndex())]",
+      "copy": { 
+        "name": "accessPolicyReaderCopy", 
+        "count": "[length(parameters('accessPolicyContributorObjectIds'))]"
+      },
+      "apiVersion": "2017-11-15",
+      "properties": {
+        "principalObjectId": "[parameters('accessPolicyContributorObjectIds')[copyIndex()]]",
+        "roles": ["Contributor"]
+      },
+      "dependsOn": [
+        "[concat('Microsoft.TimeSeriesInsights/environments/', parameters('environmentName'))]"
       ]
     }
   ],
   "outputs": {
     "dataAccessId": {
-      "value": "[reference(resourceId('Microsoft.TimeSeriesInsights/environments', parameters('environmentName'))).dataAccessId]",
+      "value": "[reference(resourceId('Microsoft.TimeSeriesInsights/environments', parameters('environmentName'))).dataAccessFQDN]",
       "type": "string"
     }
   }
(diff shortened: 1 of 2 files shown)


201-timeseriesinsights-environment-with-eventhub/azuredeploy.json
```diff
@@ -91,6 +91,20 @@
       "metadata": {
         "description": "The name of the shared access key that the Time Series Insights service will use to connect to the event hub."
       }
+    },
+    "accessPolicyReaderObjectIds": {
+      "type": "array",
+      "defaultValue": ["thisArrayIsReallyEmpty"],
```
bmoore-msft: just use an empty array then... e.g.

"defaultValue": [ ],

201-timeseriesinsights-environment-with-eventhub/azuredeploy.json
```diff
@@ -91,6 +91,20 @@
...
+      "metadata": {
+        "description": "A list of object ids of the users or applications in AAD that should have Reader access to the environment. The service principal objectId can be obtained by calling the Get-AzureRMADUser or the Get-AzureRMADServicePrincipal cmdlets. Creating an access policy for AAD groups is not yet supported."
+      }
+    },
+    "accessPolicyContributorObjectIds": {
+      "type": "array",
+      "defaultValue": ["thisArrayIsReallyEmpty"],
```
bmoore-msft: just use an empty array then... e.g.

"defaultValue": [ ],

201-timeseriesinsights-environment-with-eventhub/azuredeploy.json
```diff
@@ -194,16 +208,50 @@
...
           "dependsOn": [
-            "[concat('Microsoft.TimeSeriesInsights/environments/',parameters('environmentName'))]",
+            "[concat('Microsoft.TimeSeriesInsights/environments/', parameters('environmentName'))]",
             "[concat('Microsoft.EventHub/namespaces/', parameters('eventHubNamespaceName'), '/eventHubs/', parameters('eventHubName'), '/consumerGroups/', parameters('consumerGroupname'))]"
           ]
         }
       ]
+    },
+    {
+      "condition": "[if(equals(first(parameters('accessPolicyReaderObjectIds')), 'thisArrayIsReallyEmpty'), bool('false'), bool('true'))]",
```
bmoore-msft: simplify:

"condition": "[not(empty(parameters('accessPolicyReaderObjectIds')))]"

201-timeseriesinsights-environment-with-eventhub/azuredeploy.json
```diff
@@ -194,16 +208,50 @@
...
+      "properties": {
+        "principalObjectId": "[parameters('accessPolicyReaderObjectIds')[copyIndex()]]",
+        "roles": ["Reader"]
+      },
+      "dependsOn": [
+        "[concat('Microsoft.TimeSeriesInsights/environments/', parameters('environmentName'))]"
+      ]
+    },
+    {
+      "condition": "[if(equals(first(parameters('accessPolicyContributorObjectIds')), 'thisArrayIsReallyEmpty'), bool('false'), bool('true'))]",
```
bmoore-msft: same as above

201-timeseriesinsights-environment-with-eventhub/azuredeploy.json
```diff
@@ -194,16 +208,50 @@
...
         }
       ]
+    },
+    {
+      "condition": "[if(equals(first(parameters('accessPolicyReaderObjectIds')), 'thisArrayIsReallyEmpty'), bool('false'), bool('true'))]",
+      "type": "Microsoft.TimeSeriesInsights/environments/accesspolicies",
+      "name": "[concat(parameters('environmentName'), '/', 'readerAccessPolicy', copyIndex())]",
+      "copy": { 
+        "name": "accessPolicyReaderCopy", 
+        "count": "[if(equals(parameters('accessPolicyReaderObjectIds'), 0), 1, length(parameters('accessPolicyReaderObjectIds')))]"
```
bmoore-msft: simplify:

length(parameters('accessPolicyReaderObjectIds'))

201-timeseriesinsights-environment-with-eventhub/azuredeploy.json
```diff
@@ -194,16 +208,50 @@
...
+        "[concat('Microsoft.TimeSeriesInsights/environments/', parameters('environmentName'))]"
+      ]
+    },
+    {
+      "condition": "[if(equals(first(parameters('accessPolicyContributorObjectIds')), 'thisArrayIsReallyEmpty'), bool('false'), bool('true'))]",
+      "type": "Microsoft.TimeSeriesInsights/environments/accesspolicies",
+      "name": "[concat(parameters('environmentName'), '/', 'contributorAccessPolicy', copyIndex())]",
+      "copy": { 
+        "name": "accessPolicyReaderCopy", 
+        "count": "[if(equals(parameters('accessPolicyContributorObjectIds'), 0), 1, length(parameters('accessPolicyContributorObjectIds')))]"
```
bmoore-msft: as above

201-timeseriesinsights-environment-with-eventhub/azuredeploy.json
sandshadow: I started with that, but the copy seems unable to enumerate over an empty collection:
>Deployment template validation failed: 'The template 'copy' definition at line '223' and >column '20' has an invalid copy count. The copy count must be a postive integer value and cannot exceed '800'.
This was the workaround I was able to come up with - because of the condition on the access policy resources, the copy never gets evaluated if the array is empty.

201-timeseriesinsights-environment-with-eventhub/azuredeploy.json
sandshadow: As above.

201-timeseriesinsights-environment-with-eventhub/azuredeploy.json
sandshadow: I can't simplify this one because of the problem with copy not excepting a length of 1.

201-timeseriesinsights-environment-with-eventhub/azuredeploy.json
sandshadow: I suppose an alternative is to conditionally set the copy.count to 1 when the array is empty. But that feels like just as much a hack and one that I can't document in the template. I thought the default value I picked would help make the reason for the workaround more apparent.

201-timeseriesinsights-environment-with-eventhub/azuredeploy.json
sandshadow: Addressed. I think this must have been left over from earlier attempts to work around the copy.length problem.

201-timeseriesinsights-environment-with-eventhub/azuredeploy.json
sandshadow: I think I can do this (I haven't tested it yet) if I go with the alternative of conditionally setting copy.lenth to 1 if 'accessPolicyContributorObjectIds' is empty. But I honestly think someone who sees that is just going to think it's a template authoring bug, where my current workaround is a little more obvious. I could put something in the description indicating why the default value is ["thisArrayIsReallyEmpty"]. But this would bleed through to the deployment in the portal. Do you have any real preference for either of the two workarounds?
(I wish I could just add a comment in the template.)

201-timeseriesinsights-environment-with-eventhub/azuredeploy.json
```diff
@@ -194,16 +208,50 @@
...
         }
       ]
+    },
+    {
+      "condition": "[if(equals(first(parameters('accessPolicyReaderObjectIds')), 'thisArrayIsReallyEmpty'), bool('false'), bool('true'))]",
+      "type": "Microsoft.TimeSeriesInsights/environments/accesspolicies",
+      "name": "[concat(parameters('environmentName'), '/', 'readerAccessPolicy', copyIndex())]",
+      "copy": { 
+        "name": "accessPolicyReaderCopy", 
+        "count": "[length(parameters('accessPolicyReaderObjectIds'))]"
```
bmoore-msft: You're right we have a "bug" there...  for count let's use:

if(empty(parameters('array')), 1, length(parameters('array'))

The the rest of the items can be simplified per individual comments.

sandshadow: Done

PR #4112, opened 2017-12-01
Allow creating an event source connected to an existing event hub.

sandshadow:
### Best Practice Checklist

### Changelog

* Allow creating an event source connected to an existing event hub. This will be the common scenario when someone wants to create a Time Series Insights environment for production use, as opposed to creating the event hub as part of the deployment, which is more of a sample.

### DIFF
--- a/201-timeseriesinsights-environment-with-eventhub/azuredeploy.json
+++ b/201-timeseriesinsights-environment-with-eventhub/azuredeploy.json
@@ -205,7 +205,6 @@
       "tags": "[variables('environmentTagsValue')]",
       "resources": [
         {
-          "condition": "[empty(parameters('existingEventHubResourceId'))]",
           "type": "eventsources",
           "name": "[parameters('eventSourceName')]",
           "apiVersion": "2017-11-15",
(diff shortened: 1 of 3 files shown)


201-timeseriesinsights-environment-with-eventhub/azuredeploy.json
```diff
@@ -192,24 +205,47 @@
       "tags": "[variables('environmentTagsValue')]",
       "resources": [
         {
+          "condition": "[empty(parameters('existingEventHubResourceId'))]",
           "type": "eventsources",
-          "name": "[parameters('eventSourceName')]",
+          "name": "[if(empty(parameters('existingEventHubResourceId')), parameters('eventSourceName'), 'avoidDuplicateResourceNameWhenCreatingEventHub')]",
```
bmoore-msft: since this resource is conditional, you don't need extra logic on the name property - just use whatever property you would use when it's being created

201-timeseriesinsights-environment-with-eventhub/azuredeploy.json
```diff
@@ -192,24 +205,47 @@
...
             "[concat('Microsoft.TimeSeriesInsights/environments/', parameters('environmentName'))]",
-            "[concat('Microsoft.EventHub/namespaces/', parameters('eventHubNamespaceName'), '/eventHubs/', parameters('eventHubName'), '/consumerGroups/', parameters('consumerGroupname'))]"
+            "[resourceId('Microsoft.EventHub/Namespaces/EventHubs/ConsumerGroups', parameters('eventHubNamespaceName'), parameters('eventHubName'), parameters('consumerGroupName'))]",
+            "[resourceId('Microsoft.EventHub/Namespaces/EventHubs/AuthorizationRules', parameters('eventHubNamespaceName'), parameters('eventHubName'), parameters('eventSourceKeyName'))]"
+          ]
+        },
+        {
+          "condition": "[not(empty(parameters('existingEventHubResourceId')))]",
+          "type": "eventsources",
+          "name": "[if(empty(parameters('existingEventHubResourceId')), 'avoidDuplicateResourceNameWhenNotCreatingEventHub', parameters('eventSourceName'))]",
```
bmoore-msft: same as above, remove the if()

201-timeseriesinsights-environment-with-eventhub/azuredeploy.json
sandshadow: Unfortunately, validation doesn't seem to understand conditionals and thinks both resources are going to be deployed with the template. Without this "if" hack, I get the following validation failure:
>Deployment template validation failed: 'The resource 'Microsoft.TimeSeriesInsights/environments/GEN-UNIQUE-environmentName/eventsources/GEN-UNIQUE-eventSourceName' at line '209' and column '14' is defined multiple times in a template.

sandshadow:
Unfortunately I cant make the requested changes because they would introduce validation errors.

PR #4162, opened 2018-01-03
OMS cloud foundry solution update 2018.1

HDYA:
- [ * ] - Please check this box once you've submitted the PR if you've read through the Contribution Guide and best practices checklist.

### Changelog

* Upgrade all alerts to KQL and remove workaround due to language limitation
* Add alerts for new KPIs including internal MySQL KPIs
* Remove views for etcd server from default views list according to component change in PCF2.0
* Rename alerts and saved searches for clarity

### DIFF
--- a/oms-cloudfoundry-solution/nested/omsCustomViews.json
+++ b/oms-cloudfoundry-solution/nested/omsCustomViews.json
@@ -126,7 +126,7 @@
                         "value": "[parameters('omsWorkspaceName')]"
                     },
                     "workspaceapiversion": {
-                        "value": "2015-11-01-preview"
+                        "value": "2017-03-03-preview"
                     }
                 }
             }
@@ -161,7 +161,7 @@
                         "value": "[parameters('omsWorkspaceName')]"
                     },
                     "workspaceapiversion": {
-                        "value": "2015-11-01-preview"
+                        "value": "2017-03-03-preview"
                     }
                 }
             }
@@ -191,7 +191,7 @@
                         "value": "[parameters('omsWorkspaceName')]"
                     },
                     "workspaceapiversion": {
-                        "value": "2015-11-01-preview"
+                        "value": "2017-03-03-preview"
                     }
                 }
             }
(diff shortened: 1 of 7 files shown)


oms-cloudfoundry-solution/nested/omsCustomViews.json
```diff
@@ -87,16 +100,51 @@
...
+                        "value": "[parameters('omsWorkspaceRegion')]"
+                    },
+                    "subscriptionId": {
+                        "value": "[subscription().subscriptionId]"
+                    },
+                    "workspace": {
+                        "value": "[parameters('omsWorkspaceName')]"
+                    },
+                    "workspaceapiversion": {
+                        "value": "2015-11-01-preview"
```
bmoore-msft: this needs to be a literal value in the destination template, not passed in from the deployment - apiVersions can change the shape of the resource and a change here would likely break deployment.

Also, why are we not including all the dependent templates in the sample?

HDYA: They are actually OMS view file (.omsview) generated by OMS portal displaying and analyzing data forwarded from [OMS log analytics firehose nozzle](https://github.com/Azure/oms-log-analytics-firehose-nozzle), thus they are under the same repository as the nozzle.

This parameter is set required without a default value in these automatically generated view files(actually also templates). It seems it can only be passed in from the deployment.

bmoore-msft: Ok, a couple things:
1) if we must pull the omsview files from another repo, then use a variable for the location with a context specific name (e.g. omsViewsLocation) so as not to confuse with the BPs in the repo.

Also, it looks like the files have a defaultValue of "" for the apiVersion, so you can remove them from here.  Passing in the apiVersion for a resource is an anti-pattern.

HDYA: Updated.

I tried to remove the parameter but seems they work like this by design. I should talk to OMS team about this.

PR #4202, opened 2018-01-18
AKS zero-downtime deployment with Jenkins

ArieShout:
### Best Practice Checklist

### Changelog

* Add quick start template for AKS zero-downtime deployment with Jenkins


bmoore-msft:
Why are we skipping ci?

ArieShout:
@bmoore-msft that's not intended. I copied the base files template from `301-jenins-aptly-spinnaker-vmss`. Just removed `.ci_skip`.

bmoore-msft:
Thanks @ArieShout - check the CI failure... since containers have limited region availability, add a parameter named 'location' and set the defaultValue for one of the allowedValues

ArieShout:
Hi @bmoore-msft ,

I tried to add the `location` parameter in the `parameters` section, but it's still failing the same. I can see the `location` is restricted when I deploy it manually, but it seems the CI is not regarding the `location` parameter when it creates the resource group before deploying the target template.
So I updated the `Microsoft.ContainerService/managedClusters` resource to use the value in `location` parameter, rather than `resourceGroup().location` as suggested. Now it's failing on the service principal.
The `GEN-UNIQUE` as used in other template doesn't work for the service principal field because the managed cluster expects a working service principal to provision necessary resources. Do we have any workaround on this kind of situation?

ArieShout:
Hi @bmoore-msft , could you advise how we can workaround the issue that the template requires a working service principal to pass the deployment CI test?

Thanks,

bmoore-msft:
you'll need to use an MSI - https://docs.microsoft.com/en-us/azure/active-directory/msi-overview

Some samples: https://github.com/Azure/azure-quickstart-templates/tree/master/201-vm-msi-linux

ArieShout:
@bmoore-msft MSI is not applicable in this case. The AKS deployment requires a working service principal explicitly. There is no ARM template configuration entry to restrict it to use MSI instead. 

Can we skip the CI build in this case?

ArieShout:
As confirmed, ACS/AKS requires a working service principal configuration in the ARM template explicitly. (When we create ACS/AKS from Azure CLI we can omit the service principal parameters and it will generate one before the underlying ARM deployment, which leads to the previous confusion).

As a result, we cannot inject some dummy values for the service principal parameter, or use MSI as a replacement.

As I checked, some ACS related templates also fail the CI build (dummy reference https://github.com/Azure/azure-quickstart-templates/pull/4250). 

@bmoore-msft  Can we add `.ci_skip` to skip the CI build verification in this case?

PR #4219, opened 2018-01-25
Add jenkins-cicd-container

TylerLu:


bmoore-msft:
Let's update this to use MSI for the SP

PR #4233, opened 2018-01-31
Refact the templates

sunbinzhu:
1. Refact the templates as per the BP.
2. Enable "accelerated networking" option.


sunbinzhu:
The CI validation failure is because a storage account(which stores the custom image) in the CI platform had been deleted. Please bring it back or just ignore this failure.

bmoore-msft:
Couple notes here:

1) we're not owning any custom VHDs required for quickstarts, in fact don't support it... this might have slipped by a while back but it won't work going forward (not sure where that image would have ever been hosted)

2) the refactoring to share resources across samples isn't allowed, everything needed for a sample must come from the samples directory

3) related, it seems like everything in these samples can be combined into one and the correct workload selected at deploy time (if it couldn't the shared approach wouldn't work).

If you want a smaller PR to get SLES fixes in that should be straightforward - but the refactoring needs more attention

sunbinzhu:
@bmoore-msft Thanks for your comments:
1. Since it is mandatory to put everything needed into one sample directory, I had copied the share resources to the other two sample directories to meet the BP.
2. Refactoring is required for us because the old templates didn't take advantage of some *new* features in ARM template language, for example "condition". We had created two nested templates for every condition, this is not acceptable anymore, given the language itself already supports condition.

bmoore-msft:
Please break this into smaller PRs where CI can be run.

PR #4221, opened 2018-01-25
add jenkins-cicd-webapp

alexchx:

PR #4238, opened 2018-02-02
add jenkins-cicd-container

alexchx:
We investigated to use MSI for SP as per advice from @bmoore-msft, as the articles below, however, MSI appears not supported in managedAKS which is required in our scenario, so can we keep the current implementation and go ahead to use .ci_skip to bypass the Travis CI workflow? Thanks.

https://docs.microsoft.com/en-us/azure/active-directory/msi-overview
https://docs.microsoft.com/en-us/azure/templates/microsoft.containerservice/managedclusters


bmoore-msft:
What permissions are needed on the SP (couldn't find it in the readme)?

alexchx:
@bmoore-msft It's using the default role "Contributor", just updated the README, it's used to create Azure resources, deployment, etc.

bmoore-msft:
What scope?

alexchx:
@bmoore-msft It defaults to the current subscription, updated the README to clarify, thanks again.

bmoore-msft:
Is that what it *needs* or just what we default to?  What is the SP doing?

alexchx:
@bmoore-msft Yes, we need that role so that the SP could do the following at least.
1. Create managedAKS
2. Authenticate Jenkins build for AKS deployment
3. Authenticate Grafana to read the AKS VM info and Azure Monitor data

alexchx:
@bmoore-msft Thanks for all your advice, so do you have any other feedback? Or are you ok to merge it now?

bmoore-msft:
I'm not able to deploy it... check your use of best practices, for starters the template uri's do not include the sasToken.  I suspect that will be true with #4220 and #4221 as well...

alexchx:
@bmoore-msft there's no sasToken used in our template uris, that token is used by for the artifacts location only, the "Deploy to Azure" button in our README might not work as the PR hasn't merged yet, could you try this link if you want to deploy this template?

https://portal.azure.com/#create/Microsoft.Template/uri/https%3A%2F%2Fraw.githubusercontent.com%2FTylerLu%2Fazure-quickstart-templates%2Fmaster%2Fjenkins-cicd-container%2Fazuredeploy.json

alexchx:
@bmoore-msft Thanks a lot, I'm just aware of that issue as our goal was to support the **Deploy to Azure** button only, I just revised and committed the templates/scripts for PowerShell deployment, however, I received the following error during the deployment, that appears the **Deploy-AzureResourceGroup.ps1** passed the wrong sasToken to the ARM template, could you advise? Thanks again.

```json
"uri": "[concat(parameters('_artifactsLocation'), '/nested/grafana.json', parameters('_artifactsLocationSasToken'))]"
```

Unable to download deployment content from
'https://stage3b7912c3ad06426e862.blob.core.windows.net/jenkins-cicd-container-stageartifacts/nested/grafana.jsonSystem.Security.SecureString'

bmoore-msft:
Make sure you have the latest version of the scripts - I'm still not able to deploy with the latest changes... the VM Extensions fail.

PR #4263, opened 2018-02-08
Zerto Templates - update Zerto appliance to version zerto60ga

navalev:
1. Add visualise button to readme
2. Update Zerto appliance to version zerto60ga


navalev:
@bmoore-msft Seems that moving to a new version on the zerto appliance resulted in the same problem with the subscription, that was resolved by manually deploying the image from marketplace.

PR #4301, opened 2018-02-20
Updating ADLS deployment template w/ KeyVault encryption

matt1883:
### Best Practice Checklist

### Changelog

* Updated 101-data-lake-store-encryption-key-vault to fully configure and enable KeyVault encryption on the Data Lake Store account that is created by the template.
* This change updates the RBAC permissions on the KeyVault key to allow the ADLS account's MSI to access the key to enable encryption. This change also notifies the ADLS account that the permissions have been updated, so that the encryption settings can be confirmed.

### DIFF
--- a/101-data-lake-store-encryption-key-vault/azuredeploy.json
+++ b/101-data-lake-store-encryption-key-vault/azuredeploy.json
@@ -1,65 +1,121 @@
 {
-    "$schema": "http://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#",
-    "contentVersion": "1.0.0.0",
-    "parameters": {
-        "accountName": {
-            "type": "string",
-            "metadata": {
-                "description": "The name of the Data Lake Store account to create."
-            }
-        },
-        "location": {
-            "type": "string",
-            "allowedValues": [
-                "East US 2",
-                "Central US",
-                "North Europe"
-            ],
-            "defaultValue": "East US 2",
-            "metadata": {
-                "description": "The location in which to create the Data Lake Store account."
-            }
-        },
-        "keyVaultResourceId": {
-            "type": "string",
-            "metadata": {
-                "description": "The Azure Key Vault resource ID."
-            }
-        },
-        "encryptionKeyName": {
-            "type": "string",
-            "metadata": {
-                "description": "The Azure Key Vault encryption key name."
-            }
-        },
-        "encryptionKeyVersion": {
-            "type": "string",
-            "metadata": {
-                "description": "The Azure Key Vault encryption key version."
-            }
-        }
+  "$schema": "http://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#",
+  "contentVersion": "0.9.0.0",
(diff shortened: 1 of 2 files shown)


matt1883:
@bmoore-msft Please take a look when you have a chance.

This template expects a KeyVault to exist with a key, and the template needs the name of the KeyVault, the name of the key, and the version of the key.

The template creates an ADLS account, sets permissions on the key, and notifies the ADLS account that those permissions have been set.

Currently, the build is failing because the KeyVault, as named, doesn't exist. I believe this is because the KeyVault doesn't exist in the deployment's targeted resource group. Any recommendations?

101-data-lake-store-encryption-key-vault/azuredeploy.json
```diff
@@ -1,65 +1,121 @@
...
-        {
-            "apiVersion": "2016-11-01",
-            "name": "[parameters('accountName')]",
-            "location": "[parameters('location')]",
-            "type": "Microsoft.DataLakeStore/accounts",
-            "properties": {
-                "newTier": "Consumption",
-                "encryptionState": "Enabled",
+    {
+      "type": "Microsoft.Resources/deployments",
```
bmoore-msft: does this need to be nested?

bmoore-msft:
I updated the param file to reference the key/version for the encryption key - the error seems to be the same though,  seems like there's something with the nesting or permissions that are still off...  The vault does exist, but if you need the reference from another resourceGroup (looks like you do) then you need to add a param for that in the resourceId function...  You can use GEN-KEYVAULT-RESOURCEGROUP-NAME to get that from CI

matt1883:
@bmoore-msft The current error is a result of the template having the nested deployment depend on the completion of the "add" action on the accessPolicies list of the Key Vault. What can the dependsOn field be replace with in order for the template to successfully validate?

101-data-lake-store-encryption-key-vault/azuredeploy.json
matt1883: Yes; an error is raised when an ARM template modifies the same resource twice.

matt1883:
@bmoore-msft It looks like the "add" action on the accessPolicies list of the Key Vault needs to target the KeyVault where it exists its resource group, but I'm not sure where to specify the resource group. Can a single template target different resource groups?

101-data-lake-store-encryption-key-vault/azuredeploy.json
```diff
@@ -1,65 +1,127 @@
...
+      "apiVersion": "2017-05-10",
+      "name": "updateAdlsAccount",
+      "dependsOn": [
+        "[resourceId(parameters('keyVaultResourceGroupName'), 'Microsoft.KeyVault/vaults/accessPolicies', parameters('keyVaultName'), 'add')]"
```
bmoore-msft: I think your validation error is coming from here, try:

[concat(parameters('keyVaultName'), '/add')]

matt1883:
Addressed feedback and reopened here: #4659

PR #4306, opened 2018-02-22
Refact the templates

sunbinzhu:
1. Refact the templates as per the BP
2. Enable "accelerated networking" option.
3. Update the supported Linux distro list


sunbinzhu:
@bmoore-msft I had revert the modification to the sample create-hpc-cluster-custom-image to make CI happy. I will submit another PR to modify the sample create-hpc-cluster-custom-image after this PR is merged.

bmoore-msft:
What are we refactoring this is huge?  Seems we added some nesting around new/existing pattern, we should be using the condition property for that.

sunbinzhu:
@bmoore-msft I am trying my best to use the condition property. But sometimes i have to use new/existing/no pattern when i found condition doesn't help, for example, In my template, if user chooses to create a storage account, i will get the storage key, if user chooses not to create a storage account, i set the storage key to empty. I can use the condition property when creating the storage account, but then when i try to get the storage key as below, it fails when "NewStorage" is false, because the language parser will always execute the "listKeys" function even when its value is not used. 
if(parameters('NewStorage'), listKeys(variables('storageAccountId'), '2015-06-15').key1, '')
I tried to use condition property as much as possible to reduce the nesting templates, if you have some advice to work around above problem i mentioned with condition, please let me know, thanks.
In fact, our customers have been asking for more options in the deployment template, but due to the language limitation before (no condition support, no logic function), we cannot satisfy the customer. Now since the condition and logic function are supported, we added some more options when refacting the templates, for example, use public IP or not, use storage account or not, use accelerated networking or not, use VMSS or not...

sunbinzhu:
I will break into several smaller PRs.

PR #4320, opened 2018-02-28
Dsc extension azure automation pullserver

mgreenegit:
We need to publish an updated ReadMe that explains to customers that this example is no longer required.  Many people are using the included .zip file so we cannot remove it yet.


mgreenegit:
there is a problem with the branching of this request.  closing and starting new.

PR #4348, opened 2018-03-15
new template for ubuntu VM with dynatrace extension installed

jetinski:
new template for ubuntu vm (different vm sizes) with dynatrace vm extension pre-installed


jetinski:
I copied the JSON (parameters + template) to dedicated files and deployed to a new resource group:
$ az group deployment create --resource-group ps-rg --template-file .\azuredeploy1.json --parameters .\azuredeploy.parameters1.json -o json
no errors with my deployment (the exact JSON that was output by Travis CI) but still Travis does not accept my template. 
Is there any additional resource I can look at to find and fix the error? Thanks! :)

bmoore-msft:
You can't reference keyvaults across tenants - what specifically are you trying to pull from keyVault?

jetinski:
The use case is the following: in the quickstart template a vm extension that needs credentials to install is included.
Since the credentials are personalized, I can not commit them to github, since I would have to remove them once the CI pipeline is validated. Which on the other hand would cause the pipeline to fail once they are removed. However, there is no option for me to provide credentials that are not personalized and could be used. (the only workaround would be to provide valid credentials and invalidate them on our server after the validation is passed, but that would mean they still will be kept in the github)

Right now I’ve put these credentials in the keyvault to somehow obscure them and to make them manageable via the portal instead of having them in the quickstart template. I am aware that this might not be the perfect solution but I’ve hoped it would do the trick, which it doesn’t unfortunately.

Hope I could clarify my situation.
Is there a workaround for this issue?

Thanks for you help.

bmoore-msft:
sounds like you need a "trial" set of creds that can be used publicly

jetinski:
Thx. I’ve decided we with a different approach and publish the template in our own github repository here: https://github.com/dynatrace-innovationlab/azure-quickstart-vm

jetinski:
closing pull request, since it is now hosted here: https://github.com/dynatrace-innovationlab/azure-quickstart-vm

PR #4355, opened 2018-03-18
correct fromImage to FromImage

Nepomuceno:
### Changelog

* Correcting fromImage to From image to be compliant with a schema.

### DIFF
--- a/201-aci-wordpress/README.md
+++ b/201-aci-wordpress/README.md
@@ -1,0 +1,41 @@
+# Create a WordPress site on a Container Instance
+
+Create a WordPress site (and its MySQL database) on a Container Instance
+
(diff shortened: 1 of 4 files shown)


101-aci-dynamicsnav/azuredeploy.json
```diff
@@ -60,7 +60,7 @@
       {
               "name": "[parameters('name')]",
               "type": "Microsoft.ContainerInstance/containerGroups",
-              "apiVersion": "2017-10-01-preview",
+              "apiVersion": "2016-05-15",
```
bmoore-msft: this is not a valid apiVersion for this RP...  I think you can just revert this change

PR #4356, opened 2018-03-18
Acs engine

Nepomuceno:
### Changelog

* Correcting ACS engine api version


bmoore-msft:
Let's separate the apiVersion changes from the FromImage changes - some of the apiVersions are not correct and we're not getting a good CI on some of the samples

PR #4494, opened 2018-03-30
temporarily use older extension schema

mgreenegit:
Please accept this modification to the prior PR.  There will be a follow up change soon that shifts to a more simple model but we need an additional week.


mgreenegit:
#sign-off @bmoore-msft thanks for your patience

PR #4514, opened 2018-04-04
Take in Azure/Moodle changes to azure-quickstart-templates

hosungsmsft:
### Best Practice Checklist

### Changelog

* Take in changes made in [our development repo](https://github.com/Azure/Moodle) since last updates. Notable changes are as follows:
  * Added more configuration switches, like redis on/off, various Moodle plugins, HTML local copy feature, ...
  * Added Microsoft SQL server on Azure as a DB option
  * Updated Azure MySQL/PostgreSQL APIs for GA
  * Various other improvements


hosungsmsft:
@rgardler -- FYI

rgardler:
+1

moodle-scalable-cluster-ubuntu/azuredeploy.json
```diff
@@ -30,58 +30,96 @@
...
+        "htmlLocalCopySwitch": {
+            "defaultValue": true,
             "metadata": {
-                "description": "Storage Account type"
+                "description": "Switch to create a local copy of /moodle/html or not"
+            },
+            "type": "bool"
+        },
+        "siteURL": {
+            "defaultValue": "www.example.org",
```
bmoore-msft: is this valid for all deployments?

moodle-scalable-cluster-ubuntu/azuredeploy.json
```diff
@@ -30,58 +30,96 @@
...
-            ],
+        "sshPublicKey": {
             "metadata": {
-                "description": "File server type: GlusterFS, Azure Files (CIFS)--disabled due to too slow perf, NFS--not highly available"
+                "description": "ssh public key"
+            },
+            "type": "string"
+        },
+        "sshUsername": {
+            "defaultValue": "azureadmin",
```
bmoore-msft: remove this - increases surface area

moodle-scalable-cluster-ubuntu/azuredeploy.json
```diff
@@ -96,227 +134,262 @@
...
             "metadata": {
-                "description": "VM size for the elastic search nodes"
+                "description": "Database type"
             },
             "type": "string"
         },
-        "firewallRuleName": {
-            "defaultValue": "open-to-the-world",
+        "dbLogin": {
+            "defaultValue": "dbadmin",
```
bmoore-msft: remove this increases surface area

moodle-scalable-cluster-ubuntu/azuredeploy.json
```diff
@@ -587,6 +678,7 @@
             "dbServerType": "[parameters('dbServerType')]",
             "dbUsername": "[concat(parameters('dbLogin'), '@', parameters('dbServerType'), '-', variables('resourceprefix'))]",
             "elasticVmSku": "[parameters('elasticVmSku')]",
+            "dbDNS": "[if(equals(parameters('dbServerType'),'mssql'),concat(parameters('dbServerType'), '-',variables('resourcePrefix'), '.database.windows.net'), concat(parameters('dbServerType'), '-', variables('resourcePrefix'), '.', parameters('dbServerType'), '.database.azure.com'))]",
```
bmoore-msft: use reference() to get the FQDNs

moodle-scalable-cluster-ubuntu/azuredeploy.json
```diff
@@ -394,7 +476,7 @@
                     }
                 },
                 "templateLink": {
-                    "uri": "[concat(variables('moodleCommon').baseTemplateUrl,'redis.json',parameters('_artifactsLocationSasToken'))]"
+                    "uri": "[concat(variables('moodleCommon').baseTemplateUrl,'redis', if(parameters('redisDeploySwitch'), '', '-none'), '.json',parameters('_artifactsLocationSasToken'))]"
```
bmoore-msft: any reason we can't use the condition property for this?

moodle-scalable-cluster-ubuntu/azuredeploy.parameters.json
```diff
@@ -2,26 +2,11 @@
     "$schema": "https://schema.management.azure.com/schemas/2015-01-01/deploymentParameters.json#",
     "contentVersion": "1.0.0.0",
         "parameters": {
-                "autoscaleVmCount":            { "value": 10 },
-                "autoscaleVmSku":              { "value": "Standard_DS2_v2" },
-                "azureBackupSwitch":           { "value": true },
-                "storageAccountType":          { "value": "Standard_LRS"},
-                "controllerVmSku":             { "value": "Standard_DS1_v2" },
-                "dbServerType":                { "value": "mysql" },
+                "redisDeploySwitch":           { "value": false },
```
bmoore-msft: remove the non GEN values here and just use defaultValues in the template

moodle-scalable-cluster-ubuntu/docs/Cleanup.md
```diff
@@ -0,0 +1,22 @@
+# Cleanup All Resource
```
bmoore-msft: Resources

moodle-scalable-cluster-ubuntu/azuredeploy.json
rgardler: Yes. The install scripts take either an empty value or this default value to mean "configure it for me". In this situation the scripts generate a DNS name based on the LB name and create a self-signed certificate. This behavior is documented (not that I expect you to read the docs in a review, just letting you know it is documemented).

moodle-scalable-cluster-ubuntu/azuredeploy.json
hosungsmsft: We took the approach of providing as many reasonable default values as possible to minimize the number of unfilled required params, allowing beginning customers to try out the templates without having to figure out all the parameters. Once they get confidence and ready to deploy for their production, they can customize all these parameters (just like the siteURL parameter as above). I think we'd like to keep it this way, as it has always been. @rgardler -- Would you agree?

moodle-scalable-cluster-ubuntu/azuredeploy.json
hosungsmsft: Same response as above...

moodle-scalable-cluster-ubuntu/azuredeploy.json
hosungsmsft: According to the [documentation](https://docs.microsoft.com/en-us/azure/azure-resource-manager/resource-group-template-functions-resource#reference), we can't use `reference()` in the variables section (mentioned in Remarks). Is this changed?

moodle-scalable-cluster-ubuntu/azuredeploy.json
hosungsmsft: That's exactly what I tried at the beginning, and it kept failing, leading me to conclude that the ARM's `if(,,)` function is eager and causes this issue. Let me explain further in detail.

Basically, I wanted something like the following for the `redisKey` value in `controller.json`:
```
if([condition],listKeys([redisResourceId]).primaryKey),'None')
```
The `[condition]` would be true iff the `redisDeploySwitch` parameter is `true`. However, even when the switch is false, ARM keeps evaluating the `listKeys([redisResourceId]).primaryKey)` and complaining that there's no such resource (because no redis will be deployed when the switch is false). This was unexpected to me (as I was expecting a programming-language-like behavior with `if(,,)`), but then I realized that the parameter evaluations in `if(,,)` might be eager, call-by-value, and all 3 parameters are evaluated even if the condition is false, which is clearly different from the usual programming-language-like behavior of `if ... else ...` construct.

I've been wondering if this understanding of mine is correct, and who to ask about this. Could you confirm if this is the currently expected behavior? If not, could you let me know who to ask about this?

Bottom line is that I did/do want to get rid of this hack, but the ARM's `if(,,)` semantics I experienced doesn't allow that, so I couldn't avoid this hack. Please advise.

moodle-scalable-cluster-ubuntu/azuredeploy.parameters.json
hosungsmsft: Actually, the default value change to `redisDeploySwitch` (or any other params) requires some internal discussion (@rgardler -- please chime in). We chose the default values in `azuredeploy.json` with some direction in mind, but here in the parameters file, we need to change it to minimize the test deployment time, so that's why it's changed here.

@rgardler , I'm in favor of changing this default to false in `azuredeploy.json`, but I wouldn't do that without your agreement first.

moodle-scalable-cluster-ubuntu/docs/Cleanup.md
hosungsmsft: Fixed

moodle-scalable-cluster-ubuntu/azuredeploy.parameters.json
rgardler: I'm happy to follow your guidance on this one. My reading of the load testing results is that Reddis is not needed in anything but the most active scenarios. In which case I would support false by default.

hosungsmsft: Thanks @rgardler . Updated the default, parameters.json, and pushed the commit.

moodle-scalable-cluster-ubuntu/azuredeploy.json
bmoore-msft: sounds good

moodle-scalable-cluster-ubuntu/azuredeploy.json
bmoore-msft: Right approach, but not with something that compromises security... having a default username makes it easier to attack the app

moodle-scalable-cluster-ubuntu/azuredeploy.json
bmoore-msft: you can't you need to move this into the properties section that needs it.

moodle-scalable-cluster-ubuntu/azuredeploy.json
hosungsmsft: Understood

moodle-scalable-cluster-ubuntu/azuredeploy.json
hosungsmsft: Fixed in the newly added commit

moodle-scalable-cluster-ubuntu/azuredeploy.json
hosungsmsft: Thanks. It's working now as desired and fixed in the newly added commit.

bmoore-msft:
@hosungsmsft - can you address the merge conflicts?

hosungsmsft:
@bmoore-msft -- Your commit 0552ef815c96912fec58f26e7bdca758943db421 introduced a whole lot of merge conflicts (mostly for tab width, but also for introducing the `location` parameter to all template .json files), which will take some time for me to resolve (I'm engaged right now with some urgent customer support case). The change will also have to come from our Azure/Moodle repo, so that's how I'll resolve these conflicts later.

hosungsmsft:
@bmoore-msft -- Merge conflicts are now resolved. I modified your `location` parameter change so that it aligns with linked templates (all of the linked/nested resources should be all in the same resource group, not individually specify-able).

hosungsmsft:
@bmoore-msft -- Brian, can we get this PR finally merged? I know there were repeated merge conflicts due to this PR being opened too long (your ongoing changes on this long running PR kept causing merge conflicts, even after I resolved them manually by merging using 'theirs' strategy). I just reset my master to the current upstream master and overwrote the moodle-scalable-cluster-ubuntu directory with the changed content that were all reviewed in this PR. So it'd be greatly appreciated if this PR gets merged once the Travis CI passes. Thanks.

moodle-scalable-cluster-ubuntu/azuredeploy.json
bmoore-msft: this still needs to be removed

PR #4537, opened 2018-04-10
Modify check on resourceGroup.location

zjpjack:
### Best Practice Checklist

### Changelog

*
*
*


test/solution-template-validation-tests/test/mainTemplateTests.js
```diff
@@ -79,15 +79,16 @@ describe('template files - ', () => {
 
         });
 
-        /** The location parameter MUST NOT have allowedValues property and MUST have defaultValue property whose value MUST be '[resourceGroup().location]' */
+        /** The location parameter MUST NOT have allowedValues property and if it has defaultValue property whose value MUST be '[resourceGroup().location]' */
```
vivsriaus: The location parameter MUST NOT have allowedValues property and if it's value MUST be '[resourceGroup().location]'

test/solution-template-validation-tests/test/mainTemplateTests.js
```diff
@@ -118,6 +119,7 @@ describe('template files - ', () => {
             /** resourceGroup().location should NOT be present anywhere in template, EXCEPT as a defaultValue */
             it.each(templateObject, 'resourceGroup().location must NOT be be used in the template file ' + templateJSONObject.filename + ', except as a default value for the location parameter. Please correct similar errors in the file', function(element, next) {
                 var templateFileContent = JSON.stringify(templateObject);
```
vivsriaus: Please check if there is a "minify" json method that removes spaces/carriage return, so we wouldn't have to check for space/special chars in our regexes

zjpjack: Unfortunately, the stringify is the only json method to transfer json to string. If we would like a method that removes spaces/carriage return, we need to implement one in util.js.

test/solution-template-validation-tests/test/mainTemplateTests.js
zjpjack: Here we need to check two properties, "allowedValues" and "defaultValue", so I think the original comment is more accurate.

vivsriaus: /** The location parameter MUST NOT have allowedValues property and if it has defaultValue property, it's value MUST be '[resourceGroup().location]' */

PR #4587, opened 2018-04-26
Disable encryption templates for ADE without AAD

SudhakaraReddyEvuri:
### Best Practice Checklist

### Changelog

*
*
*

### DIFF
--- a/201-decrypt-running-windows-vm-without-aad/azuredeploy.json
+++ b/201-decrypt-running-windows-vm-without-aad/azuredeploy.json
@@ -1,0 +1,50 @@
+{
+  "$schema": "https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#",
+  "contentVersion": "1.0.0.0",
+  "parameters": {
+    "vmName": {
+      "type": "string",
+      "metadata": {
+        "description": "Name of the virtual machine"
+      }
+    },
+    "volumeType": {
+      "type": "string",
+      "defaultValue": "All",
+      "metadata": {
+        "description": "Type of the volume OS or Data to perform encryption operation"
+      }
+    },
+    "forceUpdateTag": {
+      "type": "string",
+      "defaultValue": "1.0",
+      "metadata": {
+        "description": "Pass in an unique value like a GUID everytime the operation needs to be force run"
+      }
+    }
+  },
+  "variables": {
+    "extensionName": "AzureDiskEncryption",
+    "extensionVersion": "2.2",
+    "encryptionOperation": "DisableEncryption"
+  },
+  "resources": [
+    {
+      "type": "Microsoft.Compute/virtualMachines/extensions",
+      "name": "[concat(parameters('vmName'),'/', variables('extensionName'))]",
+      "location": "[resourceGroup().location]",
+      "apiVersion": "2017-03-30",
+      "properties": {
+        "publisher": "Microsoft.Azure.Security",
+        "type": "[variables('extensionName')]",
+        "typeHandlerVersion": "[variables('extensionVersion')]",
+        "autoUpgradeMinorVersion": true,
+        "forceUpdateTag": "[parameters('forceUpdateTag')]",
+        "settings": {
+            "EncryptionOperation": "[variables('encryptionOperation')]",
+            "VolumeType": "[parameters('volumeType')]"
+        }
+      }
+    }
+  ]
+}
--- a/201-decrypt-running-windows-vm-without-aad/prereqs/prereq.azuredeploy.json
+++ b/201-decrypt-running-windows-vm-without-aad/prereqs/prereq.azuredeploy.json
@@ -1,0 +1,170 @@
+{
+  "$schema": "https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#",
+  "contentVersion": "1.0.0.0",
+  "parameters": {
+    "adminUsername": {
+      "type": "string",
+      "metadata": {
+        "description": "User name for the Virtual Machine."
+      }
+    },
+    "adminPassword": {
+      "type": "securestring",
+      "metadata": {
+        "description": "Password for the Virtual Machine."
+      }
+    },
+    "keyVaultName": {
+      "type": "string",
+      "metadata": {
+        "description": "Name of the KeyVault to place the volume encryption key"
+      }
+    },
+    "keyVaultResourceGroup": {
+      "type": "string",
+      "metadata": {
+        "description": "Resource group of the KeyVault"
+      }
+    },
+    "keyEncryptionKeyURL": {
+      "type": "string",
+      "defaultValue": "",
+      "metadata": {
+        "description": "URL of the KeyEncryptionKey used to encrypt the volume encryption key"
+      }
+    }
+  },
+  "variables": {
+    "nicName": "myVMNic",
+    "subnetName": "Subnet",
+    "vmName": "MyWindowsVM",
+    "virtualNetworkName": "MyVNET",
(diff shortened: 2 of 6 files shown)


201-decrypt-running-windows-vm-without-aad/azuredeploy.json
```diff
@@ -0,0 +1,50 @@
...
+    },
+    "forceUpdateTag": {
+      "type": "string",
+      "defaultValue": "1.0",
+      "metadata": {
+        "description": "Pass in an unique value like a GUID everytime the operation needs to be force run"
+      }
+    }
+  },
+  "variables": {
```
bmoore-msft: remove these variables and just use the values in place on the extension

201-decrypt-running-windows-vm-without-aad/prereqs/prereq.azuredeploy.json
```diff
@@ -0,0 +1,170 @@
...
+      "apiVersion": "2017-03-30",
+      "properties": {
+        "publisher": "Microsoft.Azure.Security",
+        "type": "[variables('extensionName')]",
+        "typeHandlerVersion": "[variables('extensionVersion')]",
+        "autoUpgradeMinorVersion": true,
+        "forceUpdateTag": "1.0",
+        "settings": {
+            "EncryptionOperation": "[variables('encryptionOperation')]",
+            "KeyVaultURL": "[reference(variables('keyVaultResourceId'),'2016-10-01').vaultUri]",
```
bmoore-msft: update apiVersion

201-decrypt-running-windows-vm-without-aad/azuredeploy.json
```diff
@@ -0,0 +1,50 @@
...
+      "type": "Microsoft.Compute/virtualMachines/extensions",
+      "name": "[concat(parameters('vmName'),'/', variables('extensionName'))]",
+      "location": "[resourceGroup().location]",
+      "apiVersion": "2017-03-30",
+      "properties": {
+        "publisher": "Microsoft.Azure.Security",
+        "type": "[variables('extensionName')]",
+        "typeHandlerVersion": "[variables('extensionVersion')]",
+        "autoUpgradeMinorVersion": true,
+        "forceUpdateTag": "[parameters('forceUpdateTag')]",
```
bmoore-msft: This shouldn't be needed (s/b an idempotent operation) - can you confirm?

201-decrypt-running-windows-vm-without-aad/azuredeploy.json
SudhakaraReddyEvuri: Removed 'encryptionOperation' variable. For variables 'extensionName' , 'extensionVersion' I would like to retain them as they can be used to quickly change during deployment to deploy a test extension instead of PROD extension.

201-decrypt-running-windows-vm-without-aad/prereqs/prereq.azuredeploy.json
SudhakaraReddyEvuri: As per https://docs.microsoft.com/en-us/rest/api/keyvault/key-vault-versions , 2016-10-01 is the latest KeyVault API version. Am I missing something?

201-decrypt-running-windows-vm-without-aad/azuredeploy.json
SudhakaraReddyEvuri: This is required in cases where user attaches additional data disks and wants  to perform encryption operations. In this case, CRP & GuestAgent won't invoke ADE extension unless there is a change in config. ForceUpdateTag is used as a mechanism to trigger re-running extension though there is no change in the extension config.

SudhakaraReddyEvuri:
@bmoore-msft  Travis CI seems to be failing because it didn't deploy the pre-reqs to create the VM.

201-decrypt-running-windows-vm-without-aad/azuredeploy.json
bmoore-msft: The name (as we're using it also for the type property) & extension version determines the shape of the resource so not a sound practice to assume one can change without the other.

201-decrypt-running-windows-vm-without-aad/azuredeploy.json
bmoore-msft: ok, then let's add a defaultValue to the parameter of:

[uniqueString(deployment().name)]

201-decrypt-running-windows-vm-without-aad/prereqs/prereq.azuredeploy.json
bmoore-msft: You can use: 2018-02-14-preview

201-decrypt-running-windows-vm-without-aad/azuredeploy.json
SudhakaraReddyEvuri: @bmoore-msft  . Didn't get you. This is intentionally left as a variable so that the template can be easily changed during deployment time to deploy extension of type & name 'ADETest' and version '2.3' .

bmoore-msft: That's an uncommon pattern and not one we would recommend users follow.  Very few extensions and RPs work that way.  The type and the version affect the shape of the properties object across ARM.  A change in one of those affects the other.  Changing it is not a simple variable change, it requires intentional changes to the property bag as well.  Types and versions must be explicitly specified.

201-decrypt-running-windows-vm-without-aad/prereqs/prereq.azuredeploy.json
bmoore-msft: Unfortunately there's nothing in place that will make sure that happens.  In the repo we don't want any apiVersions < 12 months old - at 2 years, we block.

201-decrypt-running-windows-vm-without-aad/azuredeploy.json
bmoore-msft: Portal, marketplace and the scripts in the repo will provide a timestamped deployment name so will work in the scenarios appropriate for the repo.  It's also just a default, it can be changed.

SudhakaraReddyEvuri:
@bmoore-msft  addressed all your comments. Please review the PR

201-decrypt-running-windows-vm-without-aad/azuredeploy.json
```diff
@@ -0,0 +1,45 @@
...
+    "volumeType": {
+      "type": "string",
+      "defaultValue": "All",
+      "metadata": {
+        "description": "Type of the volume OS or Data to perform encryption operation"
+      }
+    },
+    "forceUpdateTag": {
+      "type": "string",
+      "defaultValue": "1.0",
```
bmoore-msft: use [uniqueString(resourceGroup().id, deployment().name)]

SudhakaraReddyEvuri: @bmoore-msft  As I mentioned in my comment in earlier iteration, if user does "New-AzureRmResourceGroupDeployment -Name 'ecryptVM'.." then [uniqueString(resourceGroup().id, deployment().name)] doesn't produce unique value for two different deployments. That's the reason to leave this as is.

bmoore-msft: The current defaultValue doesn't produce anything unique either... and we're optimizing for the scenarios in which these templates are most commonly deployed 1) azure portal and 2) scripts in the repo.  Both of those methods do generate unique deployment names.

SudhakaraReddyEvuri:
@bmoore-msft  fixed the last comment about generating forceupdate tag from unique deployment name

PR #4613, opened 2018-05-07
Some updates in HPC templates

sunbinzhu:
1. Use literal apiVersion rather than variable
2. Update supported Linux operating system list.
3. Use managed disk to create VMs


sunbinzhu:
Since my previous PR (which was submitted several months ago) was too huge to do code review, as per Brian's advice, i am breaking it into several PRs, this is the first one.

sunbinzhu:
@bmoore-msft could you help to do CR and merge if the PR is acceptable? thanks

PR #4635, opened 2018-05-19
Updated the template to use Managed Disks

ramankumarlive:
Updated the template to use Managed Disks


docker-simple-on-ubuntu/metadata.json
```diff
@@ -1,4 +1,5 @@
 {
```
bmoore-msft: update the data modified as well

PR #4639, opened 2018-05-19
Updated to use Managed Disks - Removed Technical Preview

JackStromberg:
Updated template for Managed Disks
Windows Server 2016 is GA for Containers.  Removed the temporary scripts to enable containers and used the default gallery image.

Note: I think we should delete this template now that we have an image specifically for containers in the marketplace.  Windows Server 2016 is no longer in Technical Preview.


bmoore-msft:
Good call - please submit a PR to remove this instead...

PR #4643, opened 2018-05-19
Updated to use Managed Disks

JackStromberg:
Removed newStorageAccountName parameter
Renamed storageAccountType to diskType parameter to reflect managed disks


dsc-extension-iis-server-windows-vm/azuredeploy.json
```diff
@@ -64,13 +62,6 @@
       "metadata": {
         "description": "DSC configuration function to call"
       }
-    },
-    "location": {
```
bmoore-msft: revert this change - need to leave the location parameter in place

Issue #4574, opened 2018-04-19
Unable to deploy SonarQube - Validation error


felipecruz91:
[sonarqube-azuresql](https://github.com/Azure/azure-quickstart-templates/tree/master/sonarqube-azuresql)

### Issue Details
I am trying to deploy a SonarQube VM with Azure SQL DB but I get an error in the Azure portal.

### Repro steps
1. Click on the "Deploy to Azure" button located in the README.md file.
2. An Azure portal window will appear. Populate the fields that appear in the Azure quickstart template.
3. Click on "Purchase"

Please see below the validation error:

{"telemetryId":"acdf70c5-e478-4159-99df-b7b10a62ecec","bladeInstanceId":"Blade_e4462c916f604d798254dda74a586e5d_0_0","galleryItemId":"Microsoft.Template","createBlade":"DeployToAzure","code":"InvalidTemplateDeployment","message":"The template deployment 'Microsoft.Template' is not valid according to the validation procedure. The tracking id is '6001db1f-c047-4fd8-9c7a-d895404b8e4f'. See inner errors for details. Please see https://aka.ms/arm-deploy for usage details.","details":[{"code":"InvalidParameter","target":"imageReference","message":"The following list of images referenced from the deployment template are not found: Publisher: Microsoft, Offer: JDK, Sku: JDK_8, Version: latest. Please refer to https://docs.microsoft.com/en-us/azure/virtual-machines/windows/cli-ps-findimage for instructions on finding available images.","innererror":{"internalErrorCode":"ReferencedImagesNotFound"}}]}

StephanBis:
Has there been any progress regarding this issue?

felipecruz91:
Haven't heard from Microsoft yet. As soon as I receive an answer back I will post it here ;)

felipecruz91:
I've received an email with the updated [template ](https://drive.google.com/file/d/1WCRVc1yphrJOvHzNc3HAq1UXqkWEW26K/view?usp=sharing) and the following instructions:

1. Deploy the attached template by using PowerShell to do so ([Link](https://docs.microsoft.com/en-us/azure/azure-resource-manager/resource-group-template-deploy)) 
a. New-AzureRmResourceGroup -Name SonarQubeSample-RS01 -Location "South Central US"
b. New-AzureRmResourceGroupDeployment -Name InitialDeployment -ResourceGroupName SonarQubeSample-RS01 -TemplateFile ".\azuredeploy.json"
[...]

PR #4649, opened 2018-05-21
Updated to use Managed Disks + Image Reference

JackStromberg:
Updated to use Managed Disks
Updated references image to use 2012-R2-Datacenter.  The images referenced in the previous version no longer exist in the azure marketplace.


JackStromberg:
Note, the commits made also references fixes provided here: https://github.com/Azure/azure-quickstart-templates/issues/4574#issuecomment-386221588

PR #4676, opened 2018-05-25
 Auto retrieval of a LetsEncrypt cert, better ports

tfenster:
### Best Practice Checklist

### Changelog

*
*
*


tfenster:
Hi @bmoore-msft, do I need to change anything here to get it merged?

101-aci-dynamicsnav/azuredeploy.json
```diff
@@ -99,6 +105,14 @@
                 { 
                   "name": "PublicDnsName",
                   "value": "[variables('publicdnsname')]"
+                },
+                {
+                  "name": "folders",
+                  "value": "c:\\run\\my=https://github.com/NAVDEMO/LetsEncrypt/archive/master.zip\\LetsEncrypt-master"
```
bmoore-msft: this file needs to come from this repo

101-aci-dynamicsnav/azuredeploy.json
```diff
@@ -99,6 +105,14 @@
                 { 
                   "name": "PublicDnsName",
                   "value": "[variables('publicdnsname')]"
```
bmoore-msft: this needs to come from the reference() function - as this is written will fail when the location parameter is used

101-aci-dynamicsnav/azuredeploy.json
tfenster: @bmoore-msft Then I can't really test before the merge, right?

101-aci-dynamicsnav/azuredeploy.json
tfenster: @bmoore-msft I think we've had that discussion here https://github.com/Azure/azure-quickstart-templates/pull/4616/files/a3380d5851b8caad0ce09faafe22412ed178b5c7 before or am I misunderstanding you? I need to have the fqdn when the container starts and don't know how to achieve that

bmoore-msft: Alright - we need to find a solution for this... we're not going to have this pattern repeated in the repo.  I would recommend finding out where that re-entrant requirement is coming from and see how that can be addressed.

tfenster: @bmoore-msft The only thing I can come up with would be the following:

1. Run a script at the very beginning of container startup which calls an Azure Function
2. That Azure Function tries to take the requesting IP (and I hope this is the public IP of the container)
3. I run something like nslookup to determine the FQDN of that IP

Kind of crude but I don't see an alternative. Do you see any obvious problems with that approach

tfenster:
@bmoore-msft any feedback on this?

101-aci-dynamicsnav/azuredeploy.json
```diff
@@ -60,17 +66,27 @@
...
+      "defaultValue": "[resourceGroup().location]",	
+      "metadata": {	
+        "description": "Location for all resources."	
       }
     }
   },
   "variables": {
     "image": "[parameters('navRelease')]",
-    "publicdnsname": "[concat(parameters('dnsPrefix'),'.',resourceGroup().location,'.azurecontainer.io')]"
+    "publicdnsname": "[concat(parameters('dnsPrefix'),'.',resourceGroup().location,parameters('azurecontainerSuffix'))]"
```
bmoore-msft: need to use parameters('location') here...

101-aci-dynamicsnav/azuredeploy.json
bmoore-msft: See [BP guide](https://github.com/Azure/azure-quickstart-templates/blob/master/1-CONTRIBUTION-GUIDE/best-practices.md#deployment-artifacts-nested-templates-scripts) for how to handle artifacts

tfenster:
@bmoore-msft thanks a lot for your help! I finally understood how the staging works and now the ACI starts and works as expected. Do you need any more changes?

PR #4679, opened 2018-05-25
Disabling auto upgrade of minor version for custom script extension

ganesp:
Disabling auto upgrade of minor version for custom script extension


ganesp:
@bmoore-msft - 
We download few binaries/scripts using user defined script in the custom script extension.
And we face intermittent issues due to auto upgrade of the custom script extension. 
We have already disabled this auto upgrade behavior in this template already but for different user input condition. And we missed updating this particular change that time.

PR #4683, opened 2018-05-25
Updated template to use Managed disks

varunbeesu:
Updated apiversion for VM 
Removed vhd Uri reference to use Managed Disks
Updated date and added $schema in metadata

### DIFF
--- a/visual-studio-vstsbuildagent-vm/azuredeploy.json
+++ b/visual-studio-vstsbuildagent-vm/azuredeploy.json
@@ -224,9 +224,6 @@
           },
           "osDisk": {
             "name": "[concat(parameters('vmName'),'_OSDisk')]",
-            "vhd": {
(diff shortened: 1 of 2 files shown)


visual-studio-vstsbuildagent-vm/azuredeploy.json
```diff
@@ -223,7 +223,7 @@
             "version": "latest"
           },
           "osDisk": {
-            "name": "VMOSDisk",
+            "name": "[concat(parameters('vmName'),'_OSDisk')]",
             "vhd": {
               "uri": "[concat(reference(concat('Microsoft.Storage/storageAccounts/', variables('storageAccountName')), '2016-01-01').primaryEndpoints.blob, variables('vmStorageAccountContainerName'), '/', variables('vmOSDiskName'), '.vhd')]"
```
bmoore-msft: This is still using an unmanaged disk

varunbeesu:
updated to use MD

visual-studio-vstsbuildagent-vm/azuredeploy.json
varunbeesu: updated

reneloehde:
Is the Availability Set missing something like:
    
"sku": {
        "name": "Aligned"
    }

After the after the added Managed Disks?

PR #4866, opened 2018-07-02
updated to use managed disks

varunbeesu:
Updated API Version
Removed vhd uri Reference


bmoore-msft:
@varunbeesu - rather than deal with the merge conflicts and CI failures, I think we can just remove this sample, it's dated, broken and not showing anything we don't have in many other samples...  if you want to submit a PR to just remove that probably easiest

PR #4956, opened 2018-07-20
Updated to use Managed disks

varunbeesu:
Removed vhd uri reference to use managed disks


bmoore-msft:
@varunbeesu can you check the CI failure on this one?

PR #4977, opened 2018-07-26
Removed length validation from password param

cawilliamson:
The validation does not work on securestring fields on the latest Azure modules.

This PR fixes one of the templates - feel free to use the same fix over other templates.


bmoore-msft:
@cawilliamson - can you expand on this:

The validation does not work on securestring fields on the latest Azure module

That sounds like something that needs to be fixed (not the templates)

bmoore-msft:

PR #4987, opened 2018-08-01
updated to use Managed Disks

varunbeesu:
Removed VHD reference
Updated API Version


varunbeesu:
@bmoore-msft - Could you please verify this, it is terminated due to max timeout

bmoore-msft:
There's something wrong with the commit range - it's running CI on everything in the repo.  Try creating a new branch/pr and resubmit.

PR #4999, opened 2018-08-01
updated to use Managed Disks

varunbeesu:
Removed VHD URI reference
Updated API Version


varunbeesu:
@bmoore-msft - Unable to identify issue in this, aborted due to warnings.

PR #5029, opened 2018-08-07
use exact match on filename tests

bmoore-msft:
use exact match on the filename to avoid confusing "createUIDefinition.json" with "._createUIDefinition.json" created by zip programs


bmoore-msft:
@zjpjack can you have a look?

PR #5048, opened 2018-08-10
updated to use Managed Disks

varunbeesu:
Removed VHD URI reference
Updated API Version


bmoore-msft:
@varunbeesu check the build on this one, pretty sure this one worked prior to the change...

varunbeesu:
@bmoore-msft  -  Job exceeded the maximum time limit for jobs, and has been terminated.

PR #5067, opened 2018-08-14
updated to use Managed Disks

varunbeesu:
Removed VHD URI reference
Updated API Version


varunbeesu:
@bmoore-msft - Facing this error "The provided value for the template parameter 'adminPassword' is not valid" even the parameter value is set to GEN-PASSWORD

PR #5072, opened 2018-08-14
updated to use Managed Disks

varunbeesu:
Removed VHD URI reference
Updated API Version


bmoore-msft:
looks like we also missed an instance of the os disk

varunbeesu:
Removed vhd uri reference and unused storage

PR #5084, opened 2018-08-15
updated to use Managed Disks

varunbeesu:
Removed VHD URI reference
Updated API Version


bingosummer:
@varunbeesu We expect this template to run both on Azure and Azure Stack. This PR will break Azure Stack scenario, because Azure Stack doesn't support managed disks.

PR #5101, opened 2018-08-16
Vsts tomcat ubuntu vm branch

varunbeesu:

### DIFF
--- a/mongodb-high-availability/arbiter-resources.json
+++ b/mongodb-high-availability/arbiter-resources.json
@@ -58,7 +58,7 @@
     {
       "type": "Microsoft.Compute/virtualMachines",
       "name": "vmArbiter",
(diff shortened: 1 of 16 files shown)


varunbeesu:
@bmoore-msft  - Error when processing extension 'installmongodb'

bmoore-msft:
I just did a deployment of what's currently in master and that's working ok... try deploying locally and see what result you get.

varunbeesu:
I deployed locally with my branch and it worked successfully with out any error.

bmoore-msft:
ok, just noticed that you have 2 samples in this PR, mongo is failing, is that the one you tested?

bmoore-msft:
Actually, nevermind - the mongo template is referencing files that do not exist - remove that sample from this PR (or just open a new one with tomcat only).

PR #5103, opened 2018-08-16
Updated to use Managed disks

varunbeesu:


varunbeesu:
To rollback changes and raise new PR

PR #5126, opened 2018-08-22
Updated to use Managed disks

varunbeesu:
Removed unused storage


bmoore-msft:
I'm going to remove this sample, the config scripts are missing from original location (so it will never succeed)

PR #5136, opened 2018-08-24
Fixed Log Analytics Per GB SKU

ruslanlbondar:
Fixed SKU name (from "Per GB" to "pergb2018") for the Log Analytics and added "pergb2018" SKU to other files where it is necessary.

Without these changes, the service can not be deployed, because the Per GB SKU was not added everywhere where it is necessary, that when checking the template and parameters, it produced an error "Deployment template validation failed: 'The provided value 'Per GB' for the template parameter 'omsLogAnalyticsSku' at line '35' and column '31' is not valid. The parameter value is not part of the allowed value(s): '**free,standalone,pernode**'.'. (Code: InvalidTemplate)".

### DIFF
--- a/oms-azure-storage-analytics-solution/azuredeploy.json
+++ b/oms-azure-storage-analytics-solution/azuredeploy.json
@@ -38,7 +38,7 @@
                 "free",
                 "standalone",
                 "pernode",
-                "Per GB"
+                "pergb2018"
             ],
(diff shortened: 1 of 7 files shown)


bmoore-msft:
@ruslanlbondar - check the CI failure

bmoore-msft:

PR #5146, opened 2018-08-27
Updated to use Managed disks

varunbeesu:
Removed vhd uri reference to use MD
Removed storage account info


varunbeesu:
Already updated to MD

PR #5147, opened 2018-08-28
Updated to use Managed disks

varunbeesu:
Removed vhd uri to use MD
Removed unused storage
updated imagepublisher version


bmoore-msft:
@varunbeesu - check CI on this one, looks disk related...

varunbeesu:
Verified MD storage disks logic , It looks like issue is during extension "Installation"

PR #5153, opened 2018-08-29
Updated to use Managed disks

varunbeesu:
Removed vhd uri to use MD
Updated parameters

### DIFF
--- a/201-vm-os-disk-and-data-disk-existing-vnet/azuredeploy.json
+++ b/201-vm-os-disk-and-data-disk-existing-vnet/azuredeploy.json
@@ -42,12 +42,6 @@
         "description": "Name of the existing VNET"
       }
     },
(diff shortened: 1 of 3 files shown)


201-vm-os-disk-and-data-disk-existing-vnet/azuredeploy.json
```diff
@@ -42,12 +42,6 @@
         "description": "Name of the existing VNET"
       }
     },
-    "existingVirtualNetworkResourceGroup": {
```
bmoore-msft: This is still needed in the subnet variable - should be the first parameter of the resourceId function

varunbeesu: Added back existingVirtualNetworkResourceGroup parameter and updated at subnetref

PR #5154, opened 2018-08-29
Updated to use Managed disks

varunbeesu:
Removed vhd uri to use MD
updated parameter

### DIFF
--- a/201-vm-efficientip-vhd/azuredeploy.json
+++ b/201-vm-efficientip-vhd/azuredeploy.json
@@ -42,6 +42,13 @@
         "description": "Diskspace size of the VM"
       }
     },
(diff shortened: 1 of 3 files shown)


201-vm-efficientip-vhd/azuredeploy.json
```diff
@@ -48,12 +48,6 @@
         "description": "Name of the existing VNET"
       }
     },
-    "existingVirtualNetworkResourceGroup": {
```
bmoore-msft: this is still needed in the subnet ref - should be the first parameter of the resourceId function

varunbeesu: Added back existingVirtualNetworkResourceGroup parameter and updated at subnetref

PR #5159, opened 2018-08-30
Add templates for SQL Threat Detection

guywald1:
This PR adds templates for Threat Detection under the Advanced Threat Protection offering:

* A template that deploys an Azure SQL server with TD enabled (`401-sql-threat-detection-server-only`)
* A template that deploys an Azure SQL server with TD enabled + Azure SQL database (`401-sql-threat-detection-server-policy-single-db`)
* A template that deploys an Azure SQL server with a user-defined number of databases (each with TD enabled on a DB-policy level) with configurable email addresses for receiving alerts (`401-sql-threat-detection-db-policy-multiple-databases`)


401-sql-threat-detection-db-policy-multiple-databases/azuredeploy.json
```diff
@@ -0,0 +1,96 @@
+{
+  "$schema": "https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#",
+  "contentVersion": "1.0.0.0",
+  "parameters": {
+    "serverName": {
+      "type": "string",
```
bmoore-msft: add a defaultValue using uniqueString

401-sql-threat-detection-db-policy-multiple-databases/azuredeploy.json
```diff
@@ -0,0 +1,96 @@
...
+      "metadata": {
+        "description": "Name for the SQL server"
+      }
+    },
+    "databaseNames": {
```
bmoore-msft: add a defaultValue

401-sql-threat-detection-db-policy-multiple-databases/azuredeploy.json
```diff
@@ -0,0 +1,96 @@
...
+        "description": "Location for server and DBs"
+      }
+    },
+    "adminPassword": {
+      "type": "securestring",
+      "metadata": {
+        "description": "Password for admin"
+      }
+    },
+    "emailAddresses": {
```
bmoore-msft: add a defaultValue

401-sql-threat-detection-db-policy-multiple-databases/azuredeploy.json
```diff
@@ -0,0 +1,96 @@
...
+  },
+  "variables": {
+    "databaseServerName": "[toLower(parameters('serverName'))]",
+    "databaseNames": "[parameters('databaseNames')]",
+    "databaseServerLocation": "[parameters('location')]",
+    "databaseServerAdminLogin": "[parameters('adminUser')]",
+    "databaseServerAdminLoginPassword": "[parameters('adminPassword')]",
+    "emailAddresses": "[parameters('emailAddresses')]"
+  },
+  "resources": [{
```
bmoore-msft: format the json (e.g. SHIFT+ALT+F in VS Code)

401-sql-threat-detection-db-policy-multiple-databases/azuredeploy.parameters.json
```diff
@@ -0,0 +1,21 @@
+{
+  "$schema": "http://schema.management.azure.com/schemas/2015-01-01/deploymentParameters.json#",
+  "contentVersion": "1.0.0.0",
+  "parameters": {
+    "serverName": {
```
bmoore-msft: once you add a defaultValues in the template remove the ones that aren't needed

401-sql-threat-detection-db-policy-multiple-databases/README.md
```diff
@@ -0,0 +1,14 @@
+# Deploy an Azure SQL server and Azure SQL databases with Threat Detection enabled and configurable email addresses
```
bmoore-msft: let's rename the folders from 401 to 201

401-sql-threat-detection-server-only/azuredeploy.json
```diff
@@ -0,0 +1,75 @@
+{
```
bmoore-msft: let's combine these into a single sample - the deltas are pretty small and you can use the condition property to toggle features if needed.

401-sql-threat-detection-db-policy-multiple-databases/azuredeploy.json
guywald1: You mean formatting in 401-sql-threat-detection-**server-only** template, I presume? That's only one with a small inconsistency (line break between `[` and `{`). All JSONs are formatted, except for that difference in the other template

bmoore-msft: yes, but read the last comment, let's combine these into a single sample first, they are all very similar and it's likely more than one would be applied to a given server/db

guywald1:
@bmoore-msft 

Thanks!

We purposefully intented to differentiate between a server policy and database policies in the templates, two different functionalities that we intented to keep separate.
So currently, the templates are:
* Server with Threat Detection and multiple DBs without TD (don't require it because of the server policy)
* Server without TD and DBs with TD, and are therefore configurable with different email addresses for alerts, an option

Best option I could think of for merging the two is adding a condition parameter (DB policy / server policy), but this would entail ignoring the `emailAddresses` parameter for the DBs in case of selecting the server policy option and perhaps other nuances like this which to my mind could prove to be confusing for users.

So from our point of view, it would be best to keep them separate but of course that they will be merged if you think it is more appropriate.

Would love to hear your thoughts, thanks!

bmoore-msft:
Thanks - that helps... I think the nature of the child resource (TD) flipping between a child and grandchild may not simplify much.  You explanation is really good - we should add that to the readme and metadata... maybe not literally but to let users know there is more than one way to do this...

PR #5193, opened 2018-09-07
Updated to use Managed disks

varunbeesu:
Removed vhd uri to use managed disks
Added image resource


bmoore-msft:
@varunbeesu - check ci on this one, looks like it might need more of an update...

PR #5214, opened 2018-09-12
Updated to use Managed disks

varunbeesu:
Removed vhd uri to use managed disks
Updated availability set properties


bmoore-msft:
@varunbeesu - check the CI failure...

PR #5364, opened 2018-11-03
Test

mythdevishere:


bmoore-msft:
We'll need to break this up into fewer commits

PR #5377, opened 2018-11-09
Removed old WordPress templates

ItalyPaleAle:
The two templates in this PR are using outdated practices and should not be recommended. Instead, users should use https://github.com/Azure/app-service-quickstart-docker-images/blob/master/wordpress-alpine-php/0.61/README.md (direct URL is http://aka.ms/linux-wordpress )

+ @yiliaomsft


bmoore-msft:
What about people who want a config-as-code option?

ItalyPaleAle:
You can generate an ARM template from http://aka.ms/linux-wordpress Do you think that's not enough?

bmoore-msft:
Not really - users generally start by search here or on acom… nothing will appear there if we remove these (so they simply won't find it)

ItalyPaleAle:
That would be the goal. There are currently too many old repos with non-recommended practices anymore, and they're confusing customers coming in from search engines.

I could see if I can propose a new template to replace these?

bmoore-msft:
I think we still want customers to have as many entry points into WP on Linux as possible...  Even if we just want to take the automation script and replace the legacy ones, that's better than removing altogether.

bmoore-msft:
mysql is still used quite a bit by customers, I'll remove cleardb via another PR

PR #5466, opened 2018-12-30
Remove HTML in Readme template

kjacobsen:
The README template contains HTML which isn't valid Markdown (it will render but it isn't valid). Replaced it with correct Markdown.


kjacobsen:
Could probably switch to https://azuredeploy.net/deploybutton.png

Any idea why one image is hosted on GitHub and the other on a resource domain?

PR #5713, opened 2019-01-28
[AUTO-2019-01-25] for -oms-azure-storage-analytics-solution

Nepomuceno:
Changes to comply to [schema](https://github.com/Azure/azure-resource-manager-schemas) and validations made with [validator](https://github.com/Nepomuceno/arm-validation) 
 Most of this changes were auto generated


bmoore-msft:
downgrade

PR #5758, opened 2019-01-31
Azure Databricks Custom VNet Updates

tw3lveparsecs:
### Best Practice Checklist

### Changelog

* Added parameters vnetName and vnetResourceGroup
* Updated template to generate VNet ID based off vnetName and vnetResourceGroup parameters instead of having it as a string value that the user needs to retrieve.
* Removed customVirtualNetworkID parameter as no longer required.
* Added parameter managedResourceGroupName to allow users to control the name of the additional resource group that gets created.

### DIFF
--- a/101-databricks-workspace-with-vnet-injection/azuredeploy.json
+++ b/101-databricks-workspace-with-vnet-injection/azuredeploy.json
@@ -2,42 +2,54 @@
   "$schema": "https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#",
   "contentVersion": "1.0.0.0",
   "parameters": {
-    "workspaceName": {
+    "vnetName": {
       "type": "string",
       "metadata": {
-        "description": "The name of the Azure Databricks workspace to create."
+        "description": "Name of the existing virtual network for Azure Databricks"
       }
     },
-    "pricingTier": {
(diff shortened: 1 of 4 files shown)


botang-databricks:
For the `managedResourceGroupName` change, we prefer changing it back to the auto-generated one. It would be confusing for customer to ask them specify a managed resource group. We also required the managed resource group cannot be any existing resource group.

tw3lveparsecs:
@botang-databricks  just some feedback to consider as i have deployed this for two established customers now and is the reason i submitted this change.

- Each customer already had a VNet deployed and do not use the prereqs templates, therefore having the customer input the resource id for the VNet as a parameter is unacceptable in my opinion and leads to confusion.

- Each customer has naming standards they need to follow and would like flexibility to names things as they wish, hence the change for `managedResourceGroupName `

- Agreed the managed resource group cannot be any existing resource group, which is why the description states it will create a new resource group.

Perhaps you may want to consider using this template as another version of Databricks, perhaps a 102 version to cater for existing VNets and customer naming standards?

bmoore-msft:
There was a breaking change in a new version of node/npm - update your .travis.yml file to pick up that change or open a new PR after pulling from master.

bmoore-msft:

PR #5800, opened 2019-02-11
Mongodb on ubuntu updated

SychevIgor:
script required some update:
### Changelog
drop storage account
removed second subnet
updated to production api versions
mongodb 4 script update
current vm generation

### DIFF
--- a/mongodb-on-ubuntu/azuredeploy.json
+++ b/mongodb-on-ubuntu/azuredeploy.json
@@ -46,37 +46,7 @@
       "defaultValue": "Standard_A1_v2",
       "metadata": {
         "description": "VM Size"
-      },
-      "allowedValues": [
-        "Standard_A1_v2",
-        "Standard_A2m_v2",
-        "Standard_A2_v2",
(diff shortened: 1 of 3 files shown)


mongodb-on-ubuntu/azuredeploy.json
```diff
@@ -42,17 +36,47 @@
...
     },
     "vmSize": {
       "type": "string",
-      "defaultValue": "Standard_A0",
+      "defaultValue": "Standard_A1_v2",
       "metadata": {
         "description": "VM Size"
-      }
+      },
+      "allowedValues": [
```
bmoore-msft: remove this since it creates a maint problem later

SychevIgor: you mean- allowedValues? or new default value?

bmoore-msft: remove the allowedValues property

SychevIgor: @bmoore-msft done

PR #5847, opened 2019-02-27
Fix/duplicate values

pplu:
Strict JSON parsers don't admit duplicate keys. (JSON parsers in general shouldn't admit duplicate keys, since behaviour is undefined).

I've parsed all templates files with a strict parser, correcting duplicate keys. From what I've observed, it seems like the AzureARM JSON parser sticks to the value of last key parsed.

I'd be good to not accept JSON with duplicate keys, as this can make templates very difficult to understand.

### DIFF
--- a/101-backup-oms-monitoring/azuredeploy.json
+++ b/101-backup-oms-monitoring/azuredeploy.json
@@ -556,9 +556,8 @@
                             }
                         ],
                         "OverviewTile": {
-                            "Id": "SingleQueryDonutBuilderTileV1",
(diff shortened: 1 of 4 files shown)


bmoore-msft:
With these changes we'll want to see a passing CI before merging... can you check the failures?

pplu:
Hi. It looks like the CI failure is due to a problem in the subscription: `A new free workspace cannot be created in this subscription.  The limit is 10 per subscription.`

bmoore-msft:
break each sample into it's own PR and try again...

PR #5852, opened 2019-02-28
Delete duplicate osDisk

pplu:
The osDisk key is duplicate


pplu:
@bmoore-msft : I can't get this to pass. There seems to be an error unrelated to deleting the duplicate osDisk key. This stack seems to be broken before my changes.

PR #5853, opened 2019-02-28
Delete duplicate location key

pplu:
The location key is duplicate


bmoore-msft:
The location property is required for samples in the repo

PR #5892, opened 2019-03-12
bug fixes

bmoore-msft:


bmoore-msft:
@StartAutomating can you have a look?

PR #5893, opened 2019-03-13
password bug in simple vm

marlonsingleton:
### Best Practice Checklist

### Changelog

* forced adminPassword variable to be null when authn is sshPublicKey

### Reason

* current logic makes sshPublicKey null when authn is password but the reverse is not true...con't
* when authn is sshPublicKey, adminPassword is also assigned the sshPublicKey string (bug)

### DIFF
--- a/101-vm-simple-linux/azuredeploy.json
+++ b/101-vm-simple-linux/azuredeploy.json
@@ -19,10 +19,16 @@
         "description": "Type of authentication to use on the Virtual Machine. SSH key is recommended."
       }
     },
-    "adminPasswordOrKey": {
+    "adminPassword": {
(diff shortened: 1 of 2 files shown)


marlonsingleton:
Here's a validation script showing my 'test' sshPublicKey being assigned to both ssh publicKeys under osProfile and adminPassword, which appears to be the root cause of this reported bug

ssh_and_password

Once I forced the adminPassword to not take the ssh key, I stopped seeing the issue below

ssh_bug_101-vm-simple-linux

bmoore-msft:
The template is correct, the bug is in the portal's template deployment blade and needs to be fixed there.

PR #6160, opened 2019-05-08
SCCM TP: avoid copying file by AzCopy in the same time

yizhongwu:
### Best Practice Checklist

### Changelog

*
*
*

### DIFF
--- a/sccm-technicalpreview/scripts/WorkFlow-DPMP.ps1
+++ b/sccm-technicalpreview/scripts/WorkFlow-DPMP.ps1
@@ -128,7 +128,7 @@ if ($Configuration.WaitForDC.Status -eq 'NotStart') {
     $Configuration.WaitForDC.Status = 'Running'
     $Configuration.WaitForDC.StartTime = Get-Date -format "yyyy-MM-dd HH:mm:ss"
     UploadConfigFile
-    Start-Sleep -Second 600
+    Start-Sleep -Seconds 600
     $Result = WaitFor-DC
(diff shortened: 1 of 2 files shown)


sccm-technicalpreview/scripts/WorkFlow-DPMP.ps1
```diff
@@ -128,6 +128,7 @@ if ($Configuration.WaitForDC.Status -eq 'NotStart') {
     $Configuration.WaitForDC.Status = 'Running'
     $Configuration.WaitForDC.StartTime = Get-Date -format "yyyy-MM-dd HH:mm:ss"
     UploadConfigFile
+    Start-Sleep -Seconds 600
```
bmoore-msft: is there a more deterministic option here?

yizhongwu: We are using AzCopy to sync status between machines. Per some feedback, AzCopy may have problem if it write and read in the same time. This fix gave the enough time to let write finishing. Meanwhile, it won't increase the overall deployment time since this DPMP deployment workflow is not the critical path of the overall deployment time. We are changing our script to DSC which is a big change. With that, we will have more deterministic logic to handle this. To fix the issues that users are currently experiencing, I think it is safe and simple fix.

yizhongwu:
@bmoore-msft some users are experiencing failures. I am working a better fix but it may take a while. could you help approve this fix?

PR #6164, opened 2019-05-10
Updated App Service Environment creation template

ccompy:
The new ASE creation template solves several problems that customers have today with their ASEs. 
Because the ASE is itself still an RDFE cloud service under the covers, there is only so much that can be done around automated testing, GEN field values, and similar things. The user of the template must have a pre-existing VNet in order to run the template. The four values they are required to provide include:
name of the ASE, location the ASE is deployed into, resource ID of the subnet the ASE is deployed into, and the internal load balancing mode.  
We will explore being able to perform automated deployment testing with travis CI but this update will skip it it like with the previous version.

### DIFF
--- a/201-web-app-asev2-create/azuredeploy.json
+++ b/201-web-app-asev2-create/azuredeploy.json
@@ -1,5 +1,5 @@
 {
-    "$schema": "http://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json",
+    "$schema": "https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json",
     "contentVersion": "1.0.0.0",
     "parameters": {
         "aseName": {
--- a/201-web-app-asev2-create/azuredeploy.json
+++ b/201-web-app-asev2-create/azuredeploy.json
@@ -38,7 +38,6 @@
         "properties": {
           "name": "[parameters('aseName')]",
           "location": "[parameters('aseLocation')]",
-          "ipSslAddressCount": 0,
           "internalLoadBalancingMode": "[parameters('internalLoadBalancingMode')]",
           "virtualNetwork": {
             "Id": "[parameters('existingSubnetResourceId')]",
--- a/201-web-app-asev2-create/azuredeploy.json
+++ b/201-web-app-asev2-create/azuredeploy.json
@@ -1,5 +1,5 @@
 {
-    "$schema": "https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json",
+    "$schema": "http://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json",
     "contentVersion": "1.0.0.0",
     "parameters": {
         "aseName": {
@@ -10,36 +10,27 @@
         },
         "aseLocation": { 
             "type": "string",
-            "defaultValue": "South Central US",
             "metadata": { 
                 "description": "Location of the App Service Environment" 
             } 
-        }, 
-        "existingVirtualNetworkName": {
+         },
+        "existingSubnetResourceId": {
             "type": "string",
             "metadata": {
-              "description": "Name of the existing VNET"
+                "description": "ARM Url reference for the subnet that will contain the ASE.  Only Microsoft.Network is supported for ASEv2.  /subscriptions/subIDGoesHere/resourceGroups/rgNameGoesHere/providers/Microsoft.Network/virtualNetworks/vnetnamegoeshere/subnets/subnetnamegoeshere"
             }
         },
-        "existingVirtualNetworkResourceGroup": {
-            "type": "string",
+        "internalLoadBalancingMode": {
+            "type": "int",
+            "allowedValues": [0,3],
             "metadata": {
-              "description": "Name of the existing VNET resource group"
-            }
-        }, 
-        "subnetName": {
-            "type": "string",
-            "metadata": {
-                "description": "Subnet name that will contain the App Service Environment"
+                "description": "0 = public VIP only, 3 = both ports 80/443 and FTP ports are mapped to an ILB VIP."
             }
         }
     },
-    "variables":{
-        "vnetID": "[resourceId(parameters('existingVirtualNetworkResourceGroup'), 'Microsoft.Network/virtualNetworks', parameters('existingVirtualNetworkName'))]"
-    },
     "resources": [
       {
-        "apiVersion": "2015-08-01",
+        "apiVersion": "2019-02-01",
         "type": "Microsoft.Web/hostingEnvironments",
         "name": "[parameters('aseName')]",
         "kind": "ASEV2",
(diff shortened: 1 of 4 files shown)


201-web-app-asev2-create/azuredeploy.json
```diff
@@ -1,5 +1,5 @@
 {
-    "$schema": "https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json",
+    "$schema": "http://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json",
```
bmoore-msft: revert this change

201-web-app-asev2-create/azuredeploy.json
```diff
@@ -10,48 +10,39 @@
...
         "aseLocation": { 
             "type": "string",
-            "defaultValue": "South Central US",
             "metadata": { 
                 "description": "Location of the App Service Environment" 
             } 
-        }, 
-        "existingVirtualNetworkName": {
+         },
+        "existingSubnetResourceId": {
```
bmoore-msft: revert this change - pass in the name of the vnet and the rg.  The resourceId can be constructed from that

201-web-app-asev2-create/azuredeploy.parameters.json
```diff
@@ -1,21 +1,18 @@
...
+    "$schema": "http://schema.management.azure.com/schemas/2015-01-01/deploymentParameters.json",
     "contentVersion": "1.0.0.0",
     "parameters": {
         "aseName": {
-            "value": "GEN_UNIQUE"
+            "value": "GEN-UNIQUE"
         },
         "aseLocation": {
-            "value": "South Central US"
+            "value": "changeme"
```
bmoore-msft: use a valid location

201-web-app-asev2-create/azuredeploy.parameters.json
```diff
@@ -1,21 +1,18 @@
...
         },
-        "existingVirtualNetworkName": {
-            "value": "GET-PREREQ-virtualNetwork"
+        "existingSubnetResourceId": {
```
bmoore-msft: use the prereq values in the master branch (i.e. revert this change)

201-web-app-asev2-create/azuredeploy.parameters.json
```diff
@@ -1,21 +1,18 @@
...
+            "value": "/subscriptions/changeme/resourceGroups/changeme/providers/Microsoft.Network/virtualNetworks/changeme/subnets/changeme"
         },
-        "existingVirtualNetworkResourceGroup": {
-            "value": "GET-PREREQ-rgName"
-        },
-        "subnetName": {
-            "value": "GET-PREREQ-subnet"
+        "internalLoadBalancingMode": {
+            "value": "changeme"
```
bmoore-msft: use a valid defaultValue

201-web-app-asev2-create/azuredeploy.parameters.json
ccompy: This field needs to be the same location that the VNet was created in that the ASE is deployed into.  Why put in a specific location?

201-web-app-asev2-create/azuredeploy.parameters.json
ccompy: Can't do that.  This is the resource ID of the subnet.  It is not the VNet.  It is also a resource ID and not a name. 
If you were to make a resource ID from the various GEN values you could have: /subscriptions/SUBSCRIPTION ID/resourceGroups/GEN-VNET-RESOURCEGROUP-NAME/providers/Microsoft.Network/virtualNetworks/GEN-VNET-NAME/subnets/GEN-VNET-SUBNET1-NAME

I doubt that would work with your validators but do you want that instead?  

or do you have something that would be a reference to the resource ID of the subnet?

ccompy: I will try GEN-VNET-SUBNET-RESOURCE-ID and see if you are cool with that  :-)

ccompy:
Brian, I made some changes that hopefully sit better but I don't see how to automate testing offhand due to the subnet RID input value.  
We are getting CRIs because the API version is out there but I don't have the template that uses it which I can update documentation with. I would like to push it out without removing the ci_skip file and update with something around ci after that.

201-web-app-asev2-create/azuredeploy.json
ccompy: I will look into how to construct a RID in a template.  Would need the subnet name also

ccompy: Ok, I realized I can't do this.  I must take the RID of the subnet.  Many of our customers create their ASEs that are not just in a different resource group than the VNet but also they are in different subscriptions.  Collecting the subscription ID, resource group name, VNet name, and subnet name as separate items is a poor substitute for copy/paste of the subnet RID in the first place.

ccompy:
@bmoore-msft 
I don't see a clean way to do what you are looking for.  Happy to discuss but we are getting CRIs by not having this out there as a reference. 
We were able to fix a required permission item with this new API version which handles the subnet RID as one thing. Used to require join permission on the VNet.  With this change we can limit that to just the subnet. Some other items in there too that drove format changes in the new API version.

ccompy:
PR is stuck in odd state.  All items are resolved but the PR is stuck still in a requested change state.
closing and reopening

ccompy:
Reopening

ccompy:
@bmoore-msft 
I think this PR needs your acceptance of my resolutions, or some such thing.  Either way, I am happy to chat with you about the template.  Writing templates isn't something I do often and welcome any feedback. 
I do believe what I have submitted here is better than what was.  At a minimum it solves several customer problems and is up to date with the API. Can we get this update out and then iterate on it?

201-web-app-asev2-create/azuredeploy.json
bmoore-msft: you can add a defaultValue for each using the functions for retrieving the values - make the common case simple.

201-web-app-asev2-create/azuredeploy.parameters.json
bmoore-msft: so we can run ci/cd and test the common scenario

201-web-app-asev2-create/azuredeploy.parameters.json
bmoore-msft: if it passes CI/CD - should be fine

bmoore-msft:
still need to remove the .ci_skip file

ccompy:
@bmoore-msft  
An ASE deployment takes 1.5 hours to complete.  When I last was involved with working a template, my understanding was that any operation had to complete in a few minutes.

bmoore-msft:
we've increased the travis timeout to 2 hours - if that fails we can run a manual test...

201-web-app-asev2-create/azuredeploy.json
bmoore-msft: this still needs to be done

bmoore-msft:

PR #6184, opened 2019-05-16
Fixes and Enhancements

StartAutomating:
* Changing Default to Not Use Pester (can still be run in Pester with -Pester)
* Fixed Location-Should-Not-Be-Hardcoded ( templateObject was being passed incorrectly)
* Fixed Min-and-Max-Value-Are-Numbers ( incomplete error messages and bad pairing)
* Enhanced test for blanks (added whitelist of acceptable blanks)
* Enhanced test for apiVersions (enabling more exotic preview versions, handling resource type inheritance)

### DIFF
--- a/test/template-tests/testcases/deploymentTemplate/Location-Should-Not-Be-Hardcoded.test.ps1
+++ b/test/template-tests/testcases/deploymentTemplate/Location-Should-Not-Be-Hardcoded.test.ps1
@@ -3,7 +3,7 @@
 [string]$TemplateText,
 
 [Parameter(Mandatory=$true,Position=1)]
-[string]$TemplateObject,
+[PSObject]$TemplateObject,
 
 [Parameter(Mandatory=$true,Position=1)]
 [switch]$IsMainTemplate
@@ -26,7 +26,7 @@ if ($locationParameter -and
     "$($locationParameter.defaultValue)".Trim() -ne 'global' -and 
     $IsMainTemplate) {
     # If it wasn't, write an error
-    Write-Error "Location parameter must not be hardcoded.  The default value should be [resourceGroup().location]." -ErrorId Location.Parameter.Hardcoded -TargetObject $parameter
+    Write-Error "Location parameter must not be hardcoded.  The default value should be [resourceGroup().location]. It is $($locationParameter.defaultValue)" -ErrorId Location.Parameter.Hardcoded -TargetObject $parameter
 }
 
 # Now check that the rest of the template doesn't use [resourceGroup().location] 
--- a/test/template-tests/testcases/deploymentTemplate/Location-Should-Not-Be-Hardcoded.test.ps1
+++ b/test/template-tests/testcases/deploymentTemplate/Location-Should-Not-Be-Hardcoded.test.ps1
@@ -22,11 +22,10 @@ $locationParameter = $templateObject.parameters.location
 
 # Make sure that the template parameter's default is the expression [resourceGroup().location] 
 if ($locationParameter -and 
-    "$($locationParameter.defaultvalue)".Trim() -ne '[resourceGroup().location]' -and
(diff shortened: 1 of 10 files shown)


test/template-tests/testcases/deploymentTemplate/Location-Should-Not-Be-Hardcoded.test.ps1
```diff
@@ -26,7 +26,7 @@ if ($locationParameter -and
     "$($locationParameter.defaultValue)".Trim() -ne 'global' -and
```
bmoore-msft: if we're just checking the defaultValue property here, then I think we can remove this, it should never be global.  "global" is only allowed as value on a resource.

test/template-tests/testcases/deploymentTemplate/Location-Should-Not-Be-Hardcoded.test.ps1
```diff
@@ -26,7 +26,7 @@ if ($locationParameter -and
     "$($locationParameter.defaultValue)".Trim() -ne 'global' -and 
     $IsMainTemplate) {
     # If it wasn't, write an error
-    Write-Error "Location parameter must not be hardcoded.  The default value should be [resourceGroup().location]." -ErrorId Location.Parameter.Hardcoded -TargetObject $parameter
+    Write-Error "Location parameter must not be hardcoded.  The default value should be [resourceGroup().location]. It is $($locationParameter.defaultValue)" -ErrorId Location.Parameter.Hardcoded -TargetObject $parameter
```
bmoore-msft: Change message:
The defaultValue of the location parameter must not be a specific location.  The defaultValue must be... (the rest is fine)

test/template-tests/testcases/deploymentTemplate/apiVersions-Should-Be-Recent.test.ps1
```diff
@@ -39,23 +39,32 @@ foreach ($av in $allApiVersions) { # Then walk over each object containing an Ap
...
     }
+    $apiDate = [DateTime]::new($matches.Year, $matches.Month, $matches.Day) # now coerce the apiVersion into a DateTime
+
     
 
     # Now find all of the valid versions from this API
-    $validApiVersions = $AllAzureResources.$FullResourceType | 
-        Select-Object -ExpandProperty apiVersions 
-    #! sort this yourself, because it might not be
+    $validApiVersions = # This is made a little tricky by the fact that some resources don't directly have an API version
```
bmoore-msft: what's the scenario here?  The only case I can think of is when apiProfiles are used, in which case we would just bypass the check on the resource.

test/template-tests/testcases/deploymentTemplate/apiVersions-Should-Be-Recent.test.ps1
```diff
@@ -65,22 +74,17 @@ foreach ($av in $allApiVersions) { # Then walk over each object containing an Ap
         continue
     }
     
-    if ($av.ApiVersion -like '*-preview') {
+    if ($av.ApiVersion -like '*-*-*-*') {
         #! Determine the index without respect to preview versions
         $howRecent? = $validApiVersions.IndexOf($av.ApiVersion) 
         if ($howRecent?) {
-            Write-Error "$FullResourceType uses a -preview version when there are $($howRecent?) more recent versions available" -TargetObject $av
+            Write-Error "$FullResourceType uses an preview version ( $($av.apiVersion) ) when there are $($howRecent?) more recent versions available" -TargetObject $av
```
bmoore-msft: typo - "a preview version"

bmoore-msft:
Also found another (unrelated bug) here:

https://github.com/Azure/azure-quickstart-templates/blob/master/test/template-tests/testcases/deploymentTemplate/Parameters-Must-Be-Referenced.test.ps1#L7

$TemplateText needs to be passed in

PR #6194, opened 2019-05-23
Apim appgw keyvault

chrismason:
### Best Practice Checklist

### Changelog

* Example template connecting API Management and Application Gateway both with Key Vault integration

### DIFF
--- a/api-management-application-gateway/README.md
+++ b/api-management-application-gateway/README.md
@@ -37,22 +37,6 @@ Set-AzureKeyVaultSecret -VaultName $vaultName -Name $secretName -SecretValue $Se
 ```
 
 ```powershell
-# Find trusted root public certificate
-$cerFile = "CER_CERTIFICATE_FILE_PATH"
-
-$certificate = New-Object Security.Cryptography.X509Certificates.X509Certificate2
-$certificate.Import($cerFile)
-
-$chain = New-Object Security.Cryptography.X509Certificates.X509Chain
-$chain.Build($certificate)
-$rootCert = $chain.ChainElements[$chain.ChainElements.Count - 1].Certificate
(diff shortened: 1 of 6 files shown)


masonch:
@bmoore-msft I am trying to get this PR into the quickstarts and it has a dependency on a key vault being deployed with a certificate loaded into it. I am not sure there is a good way to do this via the prereqs templates. Can this be merged without passing that check?

bmoore-msft:
@masonch - check the CI failure, it looks like the problem might be that the:

"apiManagementGatewayCustomDomain"
and
"apiManagementPortalCustomDomain"

Are using the same parameter value - are those mean to be distinct?

masonch:
@bmoore-msft they are distinct URLs however I am using GEN-CUSTOM-FQDN-NAME for the CI process which is producing the same URL for both. Is there a different value I can use to inject unique parameters?

bmoore-msft:
So you need two different certs?

bmoore-msft:
@masonch - ping on that last question...

masonch:
It depends on what certs a customer has when they deploy. If they have a wildcard cert, then a single cert would work for both endpoints, if they want to be specific, they would need 1 cert for each url. I guess the question remains though is is there another way to generate unique names for each of these endpoints since GEN-CUSTOM-FQDN-NAME always seems to generate the same value?

bmoore-msft:
so with how this sample is authored, what are the requirements for certs and/or FQDNs?  The GEN values for cert/fqdn will provide the right parameter values for 1 cert and matching FQDN for that cert.  If we need 2 certs, we can set that up in CI.  If we just need a different FQDN (and cert doesn't matter), just hardcode something in the param file (e.g. contoso.com)

bmoore-msft:

PR #6195, opened 2019-05-23
Fix location attribute for AppInsights resource

aws-scripting-guy:
App Insights resource location should reference dynamic location from Resource Group as other resources.
* defaultValue set to resourceGroup().location instead of static eastus


bmoore-msft:
app insights are not globally available so there's no guarantee that they will be available in the resourceGroup location...  you can add a separate parameter for them if you want to make that change.

PR #6233, opened 2019-06-04
Making ApiVersions test apply to all templates and checking for providers().apiVersion

StartAutomating:
* Making ApiVersions apply  to all templates
* Enforcing ban on providers().apiVersions


bmoore-msft:
let's break this out into a separate file for it's own test

PR #6267, opened 2019-07-01
Azure Bastion Template

robotechredmond:
### Best Practice Checklist

### Changelog

*
*
*

### DIFF
--- a/101-azure-bastion/azuredeploy.json
+++ b/101-azure-bastion/azuredeploy.json
@@ -38,7 +38,10 @@
         },
         "location": {
             "type": "string",
-            "defaultValue": "[resourceGroup().location]"
+            "defaultValue": "[resourceGroup().location]",
+            "metadata": {
+                "description": "Azure region for Bastion and virtual network"
+            }
         }
     },
     "variables": {
--- a/101-azure-bastion/azuredeploy.json
+++ b/101-azure-bastion/azuredeploy.json
@@ -38,6 +38,7 @@
         },
         "location": {
             "type": "string",
+            "defaultValue": "[resourceGroup().location]",
             "allowedValues": [
                 "westus",
                 "eastus",
--- a/101-azure-bastion/azuredeploy.json
+++ b/101-azure-bastion/azuredeploy.json
@@ -218,6 +218,7 @@
             "location": "[parameters('location')]",
             "dependsOn": [
                 "[resourceId('Microsoft.Network/publicIpAddresses', variables('public-ip-address-name'))]",
+                "[resourceId('Microsoft.Network/virtualNetworks', parameters('vnet-name'))]",
                 "[resourceId('Microsoft.Network/virtualNetworks/subnets', parameters('vnet-name'), variables('bastion-subnet-name'))]"
             ],
             "properties": {
--- a/101-azure-bastion/azuredeploy.json
+++ b/101-azure-bastion/azuredeploy.json
@@ -21,7 +21,7 @@
                 "existing"
             ],
             "metadata": {
-                "description": "IP prefix for available addresses in vnet address space"
+                "description": "Specify whether to provision new vnet or deploy to existing vnet"
             }
         },
         "bastion-subnet-ip-prefix": {
--- a/101-azure-bastion/azuredeploy.json
+++ b/101-azure-bastion/azuredeploy.json
@@ -11,7 +11,7 @@
         },
         "vnet-ip-prefix": {
             "type": "string",
-            "defaultValue": "10.1.0.0/16"
+            "defaultValue": "10.1.0.0/16",
             "metadata": {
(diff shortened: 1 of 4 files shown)


101-azure-bastion/azuredeploy.json
```diff
@@ -0,0 +1,242 @@
...
+            "type": "Microsoft.Network/publicIpAddresses",
+            "name": "[variables('public-ip-address-name')]",
+            "location": "[parameters('location')]",
+            "sku": {
+                "name": "Standard"
+            },
+            "properties": {
+                "publicIPAllocationMethod": "Static"
+            },
+            "tags": {}
```
bmoore-msft: remove empty properties

101-azure-bastion/azuredeploy.json
```diff
@@ -0,0 +1,242 @@
...
+                    }
+                ]
+            }
+        },
+        {
+            "apiVersion": "2019-02-01",
+            "name": "[parameters('vnet-name')]",
+            "type": "Microsoft.Network/virtualNetworks",
+            "location": "[parameters('location')]",
+            "condition": "[equals(parameters('vnet-new-or-existing'), 'new')]",
```
bmoore-msft: move this to the top of the resource definition

101-azure-bastion/azuredeploy.json
```diff
@@ -0,0 +1,242 @@
...
+                                "id": "[resourceId('Microsoft.Network/networkSecurityGroups', variables('nsg-name'))]"
+                            }
+                        }
+                    }
+                ]
+            }
+        },
+        {
+            "apiVersion": "2019-02-01",
+            "type": "Microsoft.Network/virtualNetworks/subnets",
```
bmoore-msft: does it make sense to have this sample always deploy a new vnet?  This pattern is going to be pretty fragile...  For starters, it will be common to have the vnet in another resourceGroup.  As well, the subnet prefix will have to be precise (unused, no overlap).  Users will be more successful in the new pattern if that is appropriate for this scenario.

101-azure-bastion/azuredeploy.parameters.json
```diff
@@ -0,0 +1,24 @@
...
+      "value": "new"
+    },
+    "bastion-subnet-ip-prefix": {
+      "value": "10.1.1.0/27"
+    },
+    "bastion-host-name": {
+      "value": "GEN-UNIQUE-8"
+    },
+    "location": {
+      "value": "southcentralus"
```
bmoore-msft: is the preview only available in certain regions?

101-azure-bastion/azuredeploy.parameters.json
```diff
@@ -0,0 +1,24 @@
+{
+  "$schema": "https://schema.management.azure.com/schemas/2015-01-01/deploymentParameters.json#",
+  "contentVersion": "1.0.0.0",
+  "parameters": {
+    "vnet-name": {
+      "value": "vnet01"
```
bmoore-msft: remove the non-GEN values from here and just use defaultValues in the template

101-azure-bastion/azuredeploy.json
robotechredmond: done.

101-azure-bastion/azuredeploy.json
robotechredmond: done.

101-azure-bastion/azuredeploy.json
robotechredmond: We have a large # of customers who have a need for both scenarios - while it's common to have networking resources, such as VNETs in a different resource group from other resources deployed for applications, AzureBastion and the other deployed resources are all Networking elements, so customers would like deploy in the same resource group.  I'm defaulting to "new" to help ensure success for default deployments and testing, but I think there's value in providing the option for "existing" in this case.

101-azure-bastion/azuredeploy.parameters.json
robotechredmond: done. I left "ussouthcentral" for location in the parameters file to ensure CI testing, since the preview is only available in a subset of regions currently.

101-azure-bastion/azuredeploy.parameters.json
robotechredmond: Yes, the preview is currently only available in a subset of regions, but is in the process of rolling out to more regions.  By GA, all regions will be supported.

101-azure-bastion/azuredeploy.json
bmoore-msft: Ok, I'll merge this as you have it now - the scenario that will be a challenge here is deploying to a vnet in another resourceGroup (template won't work for that as written).  The template can be authored for that but we may need changes in the CI engine to support if you want to support that.

PR #6269, opened 2019-07-02
Update version

### LINKED ISSUE #6262: AKS quickstart template fails due to version default value being out of date.
### Issue Details
AKS quickstart template fails with the following error when run with defaults as per docs (https://docs.microsoft.com/en-us/azure/aks/kubernetes-walkthrough-rm-template):
```
Azure Error: InvalidTemplateDeployment
Message: The template deployment 'template' is not valid according to the validation procedure. The tracking id is '55c1314d-b0cb-4caa-ba1e-c371412afe3b'. See inner errors for details. Please see https://aka.ms/arm-deploy for usage details.
Exception Details:
        Error Code: InvalidParameter
        Message: Provisioning of resource(s) for container service aks101cluster in resource group <my rg> failed. Message: {
  "code": "InvalidParameter",
  "message": "The value of parameter orchestratorProfile.OrchestratorVersion is invalid. Please see https://aka.ms/aks-naming-rules for more details.",
  "target": "orchestratorProfile.OrchestratorVersion"
 }. Details:
```

This is due to the fact that the default version (1.12.6) is not available any longer.  Version default should be changed to 1.12.8 and allowed values to something like:
1.13.5
1.12.8
1.12.7
1.11.9
1.11.8
1.10.13 
1.10.12
(AKS team should validate, but that is what comes up for "get-versions" in westus and westeurope)

Side note: It would be nice if the error message could be more helpful in cases like this.

Side side note: Setting a default template value to something other than allowed-values gives a good message on the portal:
`Deployment template validation failed: 'The provided value '1.12.8' for the template parameter 'kubernetesVersion' at line '1' and column '1933' is not valid. The parameter value is not part of the allowed value(s): '1.10.13,1.11.8,1.12.6'.'.`
but an unhelpful one on using the az cli: 
`400 Client Error: Bad Request for url: https://management.azure.com/subscriptions/<my info>` 
It would be nice if the portal's message could be piped to the command line.

montebhoover:
### Best Practice Checklist

### Changelog

* Update kubernetesVersion default value from 1.12.6 to 1.13.5
* Remove version 1.12.6 from allowedValues and add 1.12.7 and 1.13.5

Currently with 1.12.6 as the default value for kubernetesVersion, this fails when a user follows the quickstart from https://docs.microsoft.com/en-us/azure/aks/kubernetes-walkthrough-rm-template for most regions (all regions in N.A., Europe, and Asia).  There are a few choices for what should be the default, so please let me know if there is a better one.  

The current default for AKS in most regions is 1.12.8, but that is not available and fails for a user in Brazil or South Africa regions.  1.12.7 and 1.13.5 are currently available for all regions, and I opted for 1.13.5 as the default.

One option would be to remove the default and allowed values altogether, but providing them allows for a dropdown box in the portal view for the template quickstart.

Fixes #6262

### DIFF
--- a/101-aks/metadata.json
+++ b/101-aks/metadata.json
@@ -4,7 +4,7 @@
     "itemDisplayName": "Azure Container Service (AKS)",
     "description": "Deploy a managed cluster with Azure Container Service (AKS)",
     "summary": "Deploy a managed cluster with Azure Container Service (AKS)",
-    "githubUsername": "montebhoover",
+    "githubUsername": "vyta",
     "dateUpdated": "2019-07-02"
 }
 
--- a/101-aks/metadata.json
+++ b/101-aks/metadata.json
@@ -4,7 +4,7 @@
     "itemDisplayName": "Azure Container Service (AKS)",
     "description": "Deploy a managed cluster with Azure Container Service (AKS)",
     "summary": "Deploy a managed cluster with Azure Container Service (AKS)",
-    "githubUsername": "vyta",
-    "dateUpdated": "2019-03-21"
+    "githubUsername": "montebhoover",
+    "dateUpdated": "2019-07-02"
 }
 
--- a/101-aks/azuredeploy.json
+++ b/101-aks/azuredeploy.json
@@ -83,11 +83,12 @@
         },
         "kubernetesVersion": {
             "type": "string",
-            "defaultValue": "1.12.6",
+            "defaultValue": "1.13.5",
             "allowedValues": [
                 "1.10.13",
                 "1.11.8",
-                "1.12.6"
+                "1.12.7",
+                "1.13.5"
             ],
             "metadata": {
                 "description": "The version of Kubernetes."


101-aks/metadata.json
```diff
@@ -4,7 +4,7 @@
     "itemDisplayName": "Azure Container Service (AKS)",
     "description": "Deploy a managed cluster with Azure Container Service (AKS)",
     "summary": "Deploy a managed cluster with Azure Container Service (AKS)",
-    "githubUsername": "vyta",
-    "dateUpdated": "2019-03-21"
+    "githubUsername": "montebhoover",
```
bmoore-msft: do you want to take ownership of this sample?

montebhoover: Ah I didn't realize that's what I was signifying there.  I don't mind taking ownership, but I can undo that line if you prefer.

101-aks/metadata.json
```diff
@@ -5,6 +5,6 @@
     "description": "Deploy a managed cluster with Azure Container Service (AKS)",
     "summary": "Deploy a managed cluster with Azure Container Service (AKS)",
     "githubUsername": "vyta",
```
montebhoover: Original username restored.

PR #6272, opened 2019-07-03
update templates with new index policy defaults

markjbrown:
### Best Practice Checklist

### Changelog
updated the ARM templates for SQL and Gremlin to remove the range index types on strings and numbers. The previous values are now default in the back-end service. The result is they look like what they have been updated to now.
*
*
*

### DIFF
--- a/101-cosmosdb-gremlin/azuredeploy.json
+++ b/101-cosmosdb-gremlin/azuredeploy.json
@@ -196,8 +196,7 @@
                         "indexingMode": "consistent",
                         "includedPaths": [
                             {
-                                "path": "/*",
-                                "indexes": []
+                                "path": "/*"
                             }
                         ],
                         "excludedPaths": [
@@ -229,8 +228,7 @@
                         "indexingMode": "consistent",
                         "includedPaths": [
                             {
-                                "path": "/*",
-                                "indexes": []
+                                "path": "/*"
                             }
                         ]
                     },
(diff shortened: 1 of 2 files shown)


101-cosmosdb-gremlin/azuredeploy.json
```diff
@@ -197,18 +197,7 @@
...
-                                        "dataType": "number",
-                                        "precision": -1
-                                    },
-                                    {
-                                        "kind": "Range",
-                                        "dataType": "string",
-                                        "precision": -1
-                                    }
-                                ]
+                                "indexes": []
```
bmoore-msft: Can we just remove the indexes property since it's empty?

markjbrown: We actually would prefer these remain empty arrays's since they might be used for customers implementing a custom index policy and/or adding a spatial index type. The empty property will make it easier for customers to visually know where to put them.

Thanks Brian.

bmoore-msft: The problem with empty properties on a put is it will remove any values already set, which is a common mistake users make. GET is showing the correct property value since it shows what is...  Putting a valid index in is fine, but empty has an explicit sometimes unexpected meaning.

markjbrown: I'm not sure I follow your logic. 

If someone changes the properties of a deployed resource outside of the template or does not update the template to reflect the changes made elsewhere, then that would seem to be the actual problem, not the empty array which reflects what was deployed initially. This would be true for any property, there or not.

That said I'm more interested in getting this merged than speculating on this so I will update as you suggested with a new commit.

markjbrown:
@bmoore-msft I don't know why this check failed. Can you take a look? I'll fix. Need to get this merged.

Thanks.

PR #6347, opened 2019-07-15
NSGs

ashjain:
Update template to fix NSGs

### Best Practice Checklist

### Changelog

*
*
*

### DIFF
--- a/101-azure-bastion/azuredeploy.json
+++ b/101-azure-bastion/azuredeploy.json
@@ -116,20 +116,7 @@
                         }
                     },
                     {
-                        "name": "bastion-subnet-out-allow",
-                        "properties": {
-                            "protocol": "*",
-                            "sourcePortRange": "*",
-                            "sourceAddressPrefix": "[parameters('bastion-subnet-ip-prefix')]",
-                            "destinationPortRange": "*",
-                            "destinationAddressPrefix": "[parameters('bastion-subnet-ip-prefix')]",
-                            "access": "Allow",
-                            "priority": 110,
-                            "direction": "Outbound"
-                        }
-                    },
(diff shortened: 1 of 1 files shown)


101-azure-bastion/azuredeploy.json
```diff
@@ -116,20 +116,7 @@
                         }
                     },
                     {
-                        "name": "bastion-subnet-out-allow",
```
robotechredmond: The bastion-subnet-out-allow NSG rule was intended to permit unrestricted communication between VMSS bastion instances on the AzureBastionSubnet.  Otherwise, this communication will get blocked by the VNET deny rule.  Do the VMSS bastion instances not need to communicate between themselves?

101-azure-bastion/azuredeploy.json
```diff
@@ -138,7 +125,7 @@
                             "destinationAddressPrefix": "GatewayManager",
```
robotechredmond: Since this is being changed to an inbound rule, should the "destinationAddressPrefix" be "*" and the "sourceAddressPrefix" be "GatewayManager"?

101-azure-bastion/azuredeploy.json
ashjain: This NSG is applied on the subnet and not on VM nic.  I doubt if the NSG applied on the subnet would block intra-vnet traffic within that subnet.

In either case, as of today, the Bastion instances do not talk to each other.

101-azure-bastion/azuredeploy.json
ashjain: Yes, good catch.  Please update.

PR #6387, opened 2019-07-28
Updated route tables

DazzleJim:
The templates were failing to deploy due to the network intent policy identifying configuration problems with routing. I’ve updated the templates to include the network routes as per this documentation in order to be compliant with the SQL MI network intent policy.

https://docs.microsoft.com/en-us/azure/sql-database/sql-database-managed-instance-connectivity-architecture#network-requirements


DazzleJim:
Included 2 files by mistake, will close this one and reopen two individual requests.

PR #6388, opened 2019-07-28
Updated network route table

DazzleJim:
## 101-sql-managed-instance-azure-environment
Fixing issue where this template was failing to deploy due to the SQL MI network intent policy identifying configuration problems with network routing. The update to the template includes new network routes as per the documentation to be compliant with the SQL MI network intent policy.


DazzleJim:
Closed with errors.

PR #6427, opened 2019-08-17
GEN-PASSWORD fix and CUID test fix

bmoore-msft:
### Changelog

* remove dashes from GEN-PASSWORD as some resources don't like a dash in the password
* added additional outputs to the exclusion list for CUID tests
*


bmoore-msft:
@garyke - can you have a look?

PR #6438, opened 2019-08-19
Attempting to avoid MacOSX zip files

StartAutomating:
Avoiding pulling nested files from folders ending with __macosx and source.temp

### DIFF
--- a/test/template-tests/Expand-AzureRMTemplate.ps1
+++ b/test/template-tests/Expand-AzureRMTemplate.ps1
@@ -151,8 +151,7 @@ function Expand-AzureRMTemplate
             } else {                
                 $HasCreateUIDefinition = $false
                 $createUiDefinitionFullPath = $null 
-            }
-        
+            }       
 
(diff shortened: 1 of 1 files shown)


test/template-tests/Expand-AzureRMTemplate.ps1
```diff
@@ -151,17 +151,20 @@ function Expand-AzureRMTemplate
...
 
             #*$FolderFiles (a list of objects of each file in the directory)
             $FolderFiles = 
                 @(Get-ChildItem -Path $templateFolder.FullName -Recurse |
                     Where-Object { -not $_.PSIsContainer } |
                     ForEach-Object {
 
-                        # All FolderFile objects will have the following properties:
                         $fileInfo = $_
+                        if ($fileInfo.DirectoryName -like '*__macosx' -or $fileInfo.DirectoryName -like '*source.temp') {
```
bmoore-msft: any reason why we shouldn't do an exact match on this?  Also, we can remove source.temp from the exemption...

PR #6441, opened 2019-08-21
Update location ID to Data Center location mapping

morgangrobin:
Due to historical reasons, the location ID string that is used does not always correspond to the data center from which the test is performed. I've updated the table to correctly identify the data center that is displayed in portal - which is the actual webtest location - when you create a test using ARM templates.

### Best Practice Checklist

### Changelog

* Update documentation on location ID to actual location mapping

### DIFF
--- a/201-dynamic-web-tests/README.md
+++ b/201-dynamic-web-tests/README.md
@@ -23,26 +23,26 @@ This template will help you quickly spin up any number of [Application Insights]
 }
 ```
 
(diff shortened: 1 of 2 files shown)


morgangrobin:
I updated the metadata.json, but the other build errors aren't caused by my PR. Can we merge anyway? It is important that the location to id mapping is updated, as this is causing active support tickets.

PR #6463, opened 2019-08-29
allowing for a newGuid in a defaultValue

bmoore-msft:


bmoore-msft:
@StartAutomating - can you have a look?

PR #6482, opened 2019-09-07
added outputs tests, unit-test and update regex

bmoore-msft:
### Changelog

* added new tests for secrets in outputs
* added some unit-tests for the tests
* fixed/expanded regex's in other tests to account for whitespace in searches


bmoore-msft:
@StartAutomating - please have a look, note I added "\s*" as a whitespace check around functions seems to work with the unit tests, but double check that...

StartAutomating:
If whitespace is valid between [ and parameters, then checking for it is not a bad idea.  Is that valid?

A quick review of the \s you've got though indicate a lot of them are still greedy matches.  I'd replace \s* with \s+ in most instances.

bmoore-msft:
@StartAutomating - ok, PR updates from discussion today

PR #6563, opened 2019-10-01
Ttk updates

StartAutomating:
* Adding Import-JSON (to clearly display filenames when loading bad JSON)
* Improving performance of ConvertFrom-JSON (skipping comment stripping when not required)
* Adding $TemplateFileName as a "well-known variables" (variables any test can use)
* PreReqs.azuredeploy.json files are now considered "Main" templates
* Excluding parameters by default in expansion
* Adding -Skip to Test-AzureRMTemplate


test/template-tests/Expand-AzureRMTemplate.ps1
```diff
@@ -153,7 +153,8 @@ function Expand-AzureRMTemplate
             #*$TemplateFileName (the name of the azure template file)
             $templateFileName = $TemplateFullPath | Split-Path -Leaf
             #*$IsMainTemplate (if the TemplateFileName is named mainTemplate.json)
-            $isMainTemplate = 'mainTemplate.json', 'azureDeploy.json' -contains $templateFileName
+            $isMainTemplate = 'mainTemplate.json', 'azureDeploy.json' -contains $templateFileName -or
```
bmoore-msft: why the "or" and not just add prereq.azuredeploy.json on line 156 in the collection?

test/template-tests/Format-AzureRMTemplate.ps1
```diff
@@ -23,6 +23,46 @@
             '$schema','contentVersion', 'apiProfile', 
             'parameters','functions','variables',
             'resources', 'outputs'
+
+        $resourceOrder = 'comments', 'condition', 'type', 'apiVersion', 'name', 
+            'location', 'sku', 'kind', 'dependsOn', 'tags', 'copy'
+
+        $sortProperties = {
```
bmoore-msft: have we tested nested inline resources?

test/template-tests/Test-AzureRMTemplate.ps1
```diff
@@ -182,7 +183,7 @@ Each test script has access to a set of well-known variables:
         #*Test-Group (executes a group of tests)
         function Test-Group {                
             $testQueue = [Collections.Queue]::new(@($GroupName))
-            while ($testQueue.Count) {
+            :nextTest while ($testQueue.Count) {
```
bmoore-msft: nit - does the label need to be on the same line as the while stmt?

test/template-tests/Expand-AzureRMTemplate.ps1
StartAutomating: Originally I thought that prereqs would be wildcarded.  Feel free to just add that to the list of -contains.

test/template-tests/Format-AzureRMTemplate.ps1
StartAutomating: At current this would only sort one level of resource depth.  You're right, this should be addressed.  Lmk if you'd prefer a rework of this function in this PR or a new one.

test/template-tests/Test-AzureRMTemplate.ps1
StartAutomating: _kinda_ technically it just needs to preceed a keyword.  The point of using it here is to avoid extra processing of the current test if we're skipping it.

PR #6602, opened 2019-10-08
201-rbac-multipleprincipals-maps

stack111:
### Best Practice Checklist

N/A
2. Staged artifacts use _artifactsLocation & _artifactsLocationSasToken
N/A
3. Use a parameter for resource locations with the defaultValue set to resourceGroup().location
Yes
4. Folder names for artifacts (nestedtemplates, scripts, DSC)
N/A
5. Use literal values for apiVersion (no variables)
Yes
6. Parameter files (GEN-UNIQUE for value generation and no "changemeplease" values)
Yes
7. $schema and other uris use https
Yes
8. Use uniqueString() whenever possible to generate names for resources.  While this is not required, it's one of the most common failure points in a deployment. 
Yes
9. Update the metadata.json with the current date
Yes

### Changelog
Azure Maps supports Azure AD RBAC for data-plane, I would like to provide samples to quick start template authoring to use this feature. Once this is approved I plan to create some 3xx and scenarios using App Service.

* Added 201-rbac-multipleprincipals-maps

### DIFF
--- a/201-rbac-multipleprincipals-maps/README.md
+++ b/201-rbac-multipleprincipals-maps/README.md
@@ -3,8 +3,8 @@
 <IMG SRC="https://azurequickstartsservice.blob.core.windows.net/badges/201-rbac-multipleprincipals-maps/PublicLastTestDate.svg" />&nbsp;
 <IMG SRC="https://azurequickstartsservice.blob.core.windows.net/badges/201-rbac-multipleprincipals-maps/PublicDeployment.svg" />&nbsp;
 
-<IMG SRC="https://azurequickstartsservice.blob.core.windows.net/badges/201-rbac-managedidentity-maps/FairfaxLastTestDate.svg" />&nbsp;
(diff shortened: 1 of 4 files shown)


stack111:
I expect CI to fail:

- the service principal used does not have User Access Administrator role
- the princpals will not exist in the CI subscription's Azure AD directory.

Anyone who wants to use this quickstart must provide the principal id's and provide User Access Adminstrator permissions.

bmoore-msft:
see comments in #6601 and apply the same here...

stack111:
@bmoore-msft this is an interesting question, how do we provide principal ids which do not exist the Azure AD tenants?

bmoore-msft:
This one isn't terribly different that the other so probably not worth a separate sample to maintain... lmk if I'm missing anything.

stack111:
> This one isn't terribly different that the other so probably not worth a separate sample to maintain... lmk if I'm missing anything.

Ok as long as the idea is clear, we support another 201-rbac version which does have assigned principal ids as well.

PR #6662, opened 2019-10-23
301 nested vms in virtual network

RyanLWilliams:
new quickstart template 301-nested-vms-in-virtual-network


RyanLWilliams:
Readme check generates warning

PR #6695, opened 2019-10-31
TTK Fixes 

StartAutomating:
* Improving test for AllowedValues (no longer failing when the control is used elsewhere in CreateUIDefinition, fuzzy matching when direct matching fails)
* Fixing bug in min/max test (both [int] and [long] are acceptable types)
* Adding test for DeploymentTemplate schema versions.
* Fixing bug in ConvertFrom-JSON (Regex timeout with partial block comments)
* Improving error display in Import-JSON
* Adding parameter -TestParameter(s) to Test-AzTemplate.

### DIFF
--- a/test/arm-ttk/testcases/CreateUIDefinition/Allowed-Values-Should-Actually-Be-Allowed.test.ps1
+++ b/test/arm-ttk/testcases/CreateUIDefinition/Allowed-Values-Should-Actually-Be-Allowed.test.ps1
@@ -1,22 +1,39 @@
-﻿param(
+﻿<#
+.Synopsis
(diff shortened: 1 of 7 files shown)


test/arm-ttk/testcases/CreateUIDefinition/Allowed-Values-Should-Actually-Be-Allowed.test.ps1
```diff
@@ -1,22 +1,39 @@
-﻿param(
+﻿<#
```
bmoore-msft: Let's add some unit tests for this one...

PR #6698, opened 2019-10-31
added schemas for other scopes

bmoore-msft:
### Changelog

* added schema values for schemas at other scopes
*
*


benkaday3218 (commented):
help

PR #6706, opened 2019-11-02
ARM Templates to Copy data from Azure Blob Storage to Azure Database for PostgreSQL

ramnov:
ARM Templates to Copy data from Azure Blob Storage to Azure Database for PostgreSQL

### Best Practice Checklist

### Changelog

*
*
*


101-data-factory-v2-blob-to-postgresql-copy/azuredeploy.json
```diff
@@ -0,0 +1,205 @@
+{
+  "contentVersion": "1.0.0.0",
+  "$schema": "https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#",
+  "parameters": {
+    "dataFactoryName": {
+      "type": "string",
```
bmoore-msft: add a globally unique defaultValue using uniqueString

101-data-factory-v2-blob-to-postgresql-copy/prereqs/prereq.azuredeploy.json
```diff
@@ -0,0 +1,84 @@
+{
```
bmoore-msft: doesn't look like you're actually using these - is that intentional?

101-data-factory-v2-blob-to-postgresql-copy/azuredeploy.json
ramnov: sure

101-data-factory-v2-blob-to-postgresql-copy/prereqs/prereq.azuredeploy.json
ramnov: Ideally, we are supposed to use outputs from prereq 'AzureStorage_connectionString' and 'AzurePostgreSqlDatabase_connectionString'. But the PR check always fails with message - output should not contain secrets. So removed the outputs. 

The following ARM template does have secrets in output : 
https://github.com/Azure/azure-quickstart-templates/blob/master/101-data-factory-v2-blob-to-sql-copy/prereqs/prereq.azuredeploy.json

ramnov: @bmoore-msft : I have updated the ARM Templates using your suggestion to pass storage account resource id.

101-data-factory-v2-blob-to-postgresql-copy/azuredeploy.json
```diff
@@ -53,6 +73,8 @@
     }
   },
   "variables": {
+    "storageAccountName": "[last(split(parameters('storageAccountResourceId'), '/'))]",
+    "azurePostgreSqlDatabaseConnectionString": "[concat('Server=', parameters('postgreSqlServerName'),'.postgres.database.azure.com;Port=5432;Database=' ,parameters('postgreSqlDatabase'), ';UID=', parameters('postgreSqlUserName'), '@', parameters('postgreSqlServerName'), ';Password=', parameters('postgreSqlPassword'), ';SSL Mode=Require')]",
```
bmoore-msft: use reference to get the FQDN - you'll need to move this from variables to where it's used below (reference cannot be used in a variable)

ramnov: sure. Used reference

bmoore-msft: This still needs to be removed (and the one instance below updated...

ramnov: I have indeed removed it.

PR #6709, opened 2019-11-04
Updated cloud storage template

adityabalaji-msft:
### Best Practice Checklist

### Changelog

*
*
*

### DIFF
--- a/101-backup-la-reporting/README.md
+++ b/101-backup-la-reporting/README.md
@@ -34,7 +34,7 @@ Please select the relevant log, “AzureBackupReport” in this case, to be sent
 
 
 
-<a href="https://portal.azure.com/#create/Microsoft.Template/uri/https%3A%2F%2Fraw.githubusercontent.com%2FAzure%2Fazure-quickstart-templates%2Fmaster%2F101-backup-la-reporting%2Fazuredeploy.json" target="_blank">
+<a href="https://portal.azure.com/#create/Microsoft.Template/uri/https%3A%2F%2Fraw.githubusercontent.com%2Fadityabalaji-msft%2Fazure-quickstart-templates%2Faditya-wb-nov-update%2F101-backup-la-reporting%2Fazuredeploy.json" target="_blank">
 <img src="https://raw.githubusercontent.com/Azure/azure-quickstart-templates/master/1-CONTRIBUTION-GUIDE/images/deploytoazure.png"/>
 </a>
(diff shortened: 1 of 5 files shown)


adityabalaji-msft:
Hi @bmoore-msft - Made an update to one of the nested templates. Build fails with ResourceNotFound error. I guess it is because the pipeline places the prereqs and the templates in different resource group, which is not allowed for this particular solution (as discussed in one of my earlier PRs [here](https://github.com/Azure/azure-quickstart-templates/pull/6341)).

I have deployed the template locally and it works.

If there are no other changes to be made from my end, please merge these commits. Thanks!

adityabalaji-msft:
Hi @bmoore-msft / @krnese - This template update was made in order to mitigate an ICM, so would be extremely grateful if you could take a look at the same at the earliest, so that I can inform CSS. Thanks!

adityabalaji-msft:
Any update?

adityabalaji-msft:
Hi @bmoore-msft  - The errors which show up in the pipeline results are of the following categories:

1)	**Validate metadata.json (Display name of the template is too large)** – I have now shortened the display name.
2)	**Run Best practices test (Property: Dashboard must use resourceId() instead of a static id – shows up for all the nested templates)** – ‘Dashboard’ is not really an ARM Resource. The resources being deployed are of type Microsoft.OperationalInsights/workspaces/views, of which ‘Dashboard’ is one of the properties ([doc link](https://docs.microsoft.com/en-us/azure/azure-monitor/insights/solutions-resources-views)). The static id which is populated for the dashboard in the json file comes from the LA View Designer tool, when we export the views as json. Hence there is no ‘Resource Type’ as such we can use to create an ARM id for the Dashboard property. This appears to be more like a warning and not a blocking issue.
3)	**Deploy Main Template (Resource X under Resource Group Y not found)** – As discussed in one of my earlier PRs [here](https://github.com/Azure/azure-quickstart-templates/pull/6341), I believe this error is appearing because the pipeline places the prereqs and nested templates in different resource groups. However, in the case of this template, the prereq (LA Workspace) and the nested templates (views) need to be part of the same resource group.

I have deployed the template locally and it works.

Please let me know if anything else is needed from my end.

PS – After adding the change to the display name, I now see that there is an error in the step ‘Write Test Results to readme.md’: “You cannot call a method on a null-valued expression”. Could you help me understand this error?

Would be grateful if you could provide guidance on getting this merged.

adityabalaji-msft:
@bmoore-msft Any update on this?

bmoore-msft:
It looks like there are still a few problems with CI and prereqs - the prereqs should be deployed into the same group now, but are missing parameters... let's focus on getting a successful deployment, we can look at the BP tests separately once deployment succeeds.

bmoore-msft:

PR #6847, opened 2019-11-18
update advancedThreatProtectionSettings version

neryaco:
Updated advancedThreatProtectionSettings API version from "2017-08-01-preview" to "2019-01-01".


neryaco:
@Michael-Kuritzky

PR #6850, opened 2019-11-18
Image SKU parametrization

OmegaMadLab:
I added an ImageSKU parameter to the main template. In this way, it's possible to create a cluster based on SQL Server Developer Edition images for testing or demo purposes.
I've also update ARM Template schema definitions and added a note on readme.md, to specify this template is still working fine, at least on Azure Public, even if testing badges report it as broken.


bmoore-msft:
@OmegaMadLab - if the template works, let's get a passing CI so others can make it work as well.  The readme doesn't really help with what users need to do...

Take a look at this: https://github.com/Azure/azure-quickstart-templates/tree/master/1-CONTRIBUTION-GUIDE#template-pre-requisites

PR #6886, opened 2019-11-24
301-create-ad-forest-with-subdomain: Improved reliability; made DSC dynamic

wkasdorp:
### Best Practice Checklist

### Changelog

* Updated DSC to dynamic modules
* Removed 2012 as supported OS (does not work with dynamic DSC)
* Made reliability updates; main concern was that the swapfile was too small on small VMs. This is a 2016/2019 problem.
* updated readme.md accordingly


wkasdorp:
@bmoore-msft : the Travis CI is failing on metadata.json with a script error. I reverted to the default (-100) version, and it still errors out. Can you take a look? 

image

wkasdorp:
@bmoore-msft I think we are ready for this PR. The main problem here was that I started using the script extension, which is apparently bugged: the working directory in the VM depends on the nature of the download source: is it a blob or something else? This is the workaround:
    
```
    "UrlIsBlob": "[or(contains(parameters('_artifactsLocation'), 'blob.core.windows.net'), contains(parameters('_artifactsLocation'), 'blob.core.usgovcloudapi.net'))]",
    "scriptExecutionPath": "[if(variables('UrlIsBlob'), variables('scriptFolderName'), '')]", 
    "commandToExecute": "[concat('powershell.exe -ExecutionPolicy Unrestricted -File ./', variables('scriptExecutionPath'),variables('scriptName'),' -SkipNetworkProfileCheck $true')]"
```
I logged an issue at our github page [Custom Script Extension for Windows](https://docs.microsoft.com/en-us/azure/virtual-machines/extensions/custom-script-windows#feedback)

301-create-ad-forest-with-subdomain/nested/CreateAndPrepnewVM.json
```diff
@@ -92,7 +101,15 @@
...
     "vmDataDiskSize": 10,
     "imagePublisher": "MicrosoftWindowsServer",
-    "imageOffer": "WindowsServer"
+    "imageOffer": "WindowsServer",
+
+    "scriptFolderName": "scripts/",
+    "scriptName": "Configure-DSC.ps1",
+    "scriptDownLoad": "[uri(parameters('_artifactsLocation'), concat(variables('scriptFoldername'), variables('scriptName'), parameters('_artifactsLocationSasToken')))]",
+    
+    "UrlIsBlob": "[or(contains(parameters('_artifactsLocation'), 'blob.core.windows.net'), contains(parameters('_artifactsLocation'), 'blob.core.usgovcloudapi.net'))]",
```
bmoore-msft: let's just move this one script file to the root of the sample and remove this logic...

301-create-ad-forest-with-subdomain/azuredeploy.json
```diff
@@ -68,16 +67,14 @@
         "Standard_DS3_v2"
       ],
       "metadata": {
-        "description": "The size and type of the VM. *_DS* or *s types support Premium Disks."
+        "description": "The size and type of the VM. *_DS* or *s types support Premium Disks. Hint: the DS* series seem most reliable for this complex template."
```
bmoore-msft: DS is no more or less reliable than any other size - what problem are we trying to solve here?

301-create-ad-forest-with-subdomain/azuredeploy.json
```diff
@@ -87,14 +84,14 @@
...
-      "defaultValue": "StandardSSD_LRS",
+      "defaultValue": "Premium_LRS",
       "allowedValues": [
         "StandardSSD_LRS",
         "Standard_LRS",
         "Premium_LRS"
       ],
       "metadata": {
-        "description": "The Storage type of the data Disks. Use Premium_LRS only with *s or *DS* VM types."
+        "description": "The Storage type of the data Disks. Use Premium_LRS only with *s or *DS* VM types. Warning: non-premium may be too slow."
```
bmoore-msft: why?

301-create-ad-forest-with-subdomain/azuredeploy.json
wkasdorp: Speed is the problem, the combined speed of CPU and disk. If it is too slow, the template times out on DSC stuff. Server 2016 is especially vulnerable for this.  Apparently it is very busy just after deployment, so busy the template gets too few resources. So, I tried to select sku and disk combinations that are fast enough, just to stabilize the template. This particular (fast) combination has not failed me yet. 

I never had any problem with 2012 R2 on slower disks and sku's. That worked on an A1 with SATA disks.

bmoore-msft: that seems a bit odd, would imply that the DSC extension just doesn't work (which would be a regression because it used to).  For now, let's use standard_ssd, and set allowedValues to a set of vms that have the core/memory requirements for AD...  Update the description for the requirement, not the perception of reliability.  If we see regular failures we can escalate that...

wkasdorp: > that seems a bit odd, would imply that the DSC extension just doesn't work (which would be a regression because it used to).

It's not really the extension, it's the DSC code that is used to install and configure Active Directory. This is resource intensive. Windows Server 2016 simply seems a resource hog just after deployment.

I'll make the suggested change and test if Server 2016 is stable. Server 2019 should be OK.

wkasdorp: I tried three times with Windows 2016 on standard SSD. It succeeded twice, and failed once with a DSC timeout: 

image

And that's typical. With full SSD, this does not happen -- at least I have not seen it yet. 

@bmoore-msft I'd like to move back to full SSD just to avoid disappointments. Thoughts?

301-create-ad-forest-with-subdomain/nested/CreateAndPrepnewVM.json
wkasdorp: It is now recognized as a bug: https://github.com/MicrosoftDocs/azure-docs/issues/43816

wkasdorp:
@bmoore-msft Can you have a look a the current state? Anything left to be done? Thx!

PR #7001, opened 2019-12-11
fix line endings per git attribute file

kucharskim:
Done with `find . -type f -name '*.sh' -print0 | xargs -r0 dos2unix`


kucharskim:
When experimenting with templates, git normalized files unrelated to my experiment. To avoid those unexpected file modifications, I converted all shell scripts (files with .sh extension, per git attribute file) to Unix line endings.

PR #7040, opened 2019-12-15
Creation of manage identity when crating SQL server for vulnerability assessments

shimonar1171:
Creation of manage identity when crating SQL server for vulnerability assessments

1. Add role assignments as StorageBlobContributor to the new storage
2. Add enableVA parameter and add condition for creation of Managed Identity that depend on enableVA

### DIFF
--- a/101-sql-logical-server/azuredeploy.json
+++ b/101-sql-logical-server/azuredeploy.json
@@ -29,6 +29,13 @@
             "description": "Enable Advanced Data Security."
           }
       },
+      "enableVA": {
+            "defaultValue": false,
+            "type": "Bool",
+          "metadata": {
+            "description": "To enable vulnerability assessments, the user deploying the template must have an administrator or owner permissions."
+          }
+      },
       "allowAzureIPs": {
           "defaultValue": true,
           "type": "bool",
@@ -56,7 +63,9 @@
        "serverResourceGroupName": "[resourceGroup().name]",
        "subscriptionId": "[subscription().subscriptionId]",
        "uniqueStorage": "[uniqueString(variables('subscriptionId'), variables('serverResourceGroupName'), parameters('location'))]",
-       "storageName": "[tolower(concat('sqlva', variables('uniqueStorage')))]"
+       "storageName": "[tolower(concat('sqlva', variables('uniqueStorage')))]",
+       "uniqueRoleGuid": "[guid(resourceId('Microsoft.Sql/servers', parameters('serverName')))]",
+       "StorageBlobContributor": "[subscriptionResourceId('Microsoft.Authorization/roleDefinitions', 'ba92f5b4-2d11-453d-a403-e96b0029c9fe')]"
    },
   "resources": [
       {
@@ -64,6 +73,7 @@
           "name": "[parameters('serverName')]",
           "apiVersion": "2018-06-01-preview",
           "location": "[parameters('location')]",
+          "identity": "[if(and(parameters('enableADS'),parameters('enableVA')), json('{\"type\":\"SystemAssigned\"}'), json('null'))]",
           "properties": {
               "administratorLogin": "[parameters('administratorLogin')]",
               "administratorLoginPassword": "[parameters('administratorLoginPassword')]",
@@ -98,7 +108,7 @@
                   }
               },
               {
-                  "condition": "[parameters('enableADS')]",
+                  "condition": "[and(parameters('enableADS'),parameters('enableVA'))]",
                   "type": "vulnerabilityAssessments",
                   "name": "Default",
                   "apiVersion": "2018-06-01-preview",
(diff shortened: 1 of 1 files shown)


shimonar1171:
@bmoore-msft , @talhers

101-sql-logical-server/azuredeploy.json
```diff
@@ -108,8 +118,7 @@
                       "[resourceId('Microsoft.Storage/storageAccounts', variables('storageName'))]"
                   ],
                   "properties": {
-                      "storageContainerPath": "[if(parameters('enableADS'), concat(reference(resourceId('Microsoft.Storage/storageAccounts', variables('storageName')), '2018-07-01').primaryEndpoints.blob, 'vulnerability-assessment'), '')]",
-                      "storageAccountAccessKey": "[if(parameters('enableADS'), listKeys(resourceId('Microsoft.Storage/storageAccounts', variables('storageName')), '2018-02-01').keys[0].value, '')]",
+                      "storageContainerPath": "[if(and(parameters('enableADS'),parameters('enableVA')), concat(reference(resourceId('Microsoft.Storage/storageAccounts', variables('storageName')), '2018-07-01').primaryEndpoints.blob, 'vulnerability-assessment'), '')]",
```
bmoore-msft: I think you can remove the conditional here since the resource itself is deployed with the same condition

101-sql-logical-server/azuredeploy.json
```diff
@@ -130,15 +139,33 @@
...
+                        "[resourceId('Microsoft.Sql/servers', parameters('serverName'))]",
+                        "[resourceId('Microsoft.Storage/storageAccounts', variables('storageName'))]"
+                    ],
+                    "properties": {
+                        "roleDefinitionId": "[variables('StorageBlobContributor')]",
+                        "principalId": "[reference(resourceId('Microsoft.Sql/servers', parameters('serverName')), '2018-06-01-preview', 'Full').identity.principalId]",
+                        "scope": "[resourceId('Microsoft.Storage/storageAccounts', variables('storageName'))]",
+                        "principalType": "ServicePrincipal"
+                    },
+                    "condition": "[and(parameters('enableADS'),parameters('enableVA'))]"
```
bmoore-msft: move this to the top of the resource definition

101-sql-logical-server/azuredeploy.json
```diff
@@ -56,14 +63,17 @@
        "serverResourceGroupName": "[resourceGroup().name]",
        "subscriptionId": "[subscription().subscriptionId]",
        "uniqueStorage": "[uniqueString(variables('subscriptionId'), variables('serverResourceGroupName'), parameters('location'))]",
-       "storageName": "[tolower(concat('sqlva', variables('uniqueStorage')))]"
+       "storageName": "[tolower(concat('sqlva', variables('uniqueStorage')))]",
+       "uniqueRoleGuid": "[guid(resourceId('Microsoft.Sql/servers', parameters('serverName')))]",
```
bmoore-msft: this should be based on your scope, role and prinicipal... you won't be able to incorporate principal since it's an MSI, but you can fix the scope and role, e.g.

guid(variables(resourceId('Microsoft.Storage/storageAccounts, variables('storageName'), variables('storageBlobContributor'))

101-sql-logical-server/azuredeploy.json
shimonar1171: I removed it in the first commit and the deploy failed (you can see the first commit error https://dev.azure.com/azurequickstarts/azure-quickstart-templates/_build/results?buildId=9910&view=results)

101-sql-logical-server/azuredeploy.json
shimonar1171: Why ?

101-sql-logical-server/azuredeploy.json
shimonar1171: I can set the guid to be :
 guid(variables(resourceId('Microsoft.Storage/storageAccounts, variables('storageName'), variables('storageBlobContributor'), resourceId('Microsoft.Sql/servers', parameters('serverName'))) 
???

shimonar1171: I can't use only the storage name and the storageBlobContributor variable.
My storage name is created form - subscription, resource group and location.
If I run this template twice for different servers the last role id will override the first .
So, because of this, I need to add the server name when I create the guid.

bmoore-msft: sounds good

101-sql-logical-server/azuredeploy.json
bmoore-msft: remove the apiVersion parameter from the reference call

101-sql-logical-server/azuredeploy.json
bmoore-msft: We're trying to establish some consistency:

https://github.com/Azure/azure-quickstart-templates/blob/master/1-CONTRIBUTION-GUIDE/best-practices.md#sort-order-of-properties

101-sql-logical-server/azuredeploy.json
shimonar1171: Ok

PR #7092, opened 2020-01-09
Add sku to azure sql database resource arm template to make sure its DTU based model as intended

melborp:

### Changelog

* Add sku element to 4 templates that are intending to use DTU based model, but dont specify it. Without the sku element the database can end up being vCore based instead (as per latest tests)


bmoore-msft:
@melborp - you'll need to split this into separate PRs - one for each sample.

PR #7144, opened 2020-01-30
Josuemb parametrize deploy

josuemb:
### Changelog

*Promote all variables to parameters in order to be easy to change all names
*Change some parameters name for an easier comprehension of it.


josuemb:
I am cancelling this since I will be requesting a new (updated) one.

PR #7145, opened 2020-01-30
Parametrized template

josuemb:
### Changelog

*All variables promoted to parameters in order to easier customization of artifact names and preserving same easy use by letting default names.
*Some parameters renamed for an easier comprehension.


josuemb:
### To consider:

- Same logic and "spirit" preserved just added some flexibility. :thumbsup:
- Already probed in Azure deployments. :ok_hand:

PR #7165, opened 2020-02-07
Azure Viya Quickstart 1.4

chlync:
### Best Practice Checklist

### Changelog

* Added Support for Viya 3.5
* Added Support for MPP
* Other bug fixes and improvements


sas-viya/azuredeploy.json
```diff
@@ -101,20 +117,18 @@
...
     "VirtualNetworkApplicationGatewaySubnetCIDR": "10.0.129.0/24",
 
-    "LoadBalancerName": "PrimaryViyaLoadbalancer",
-    "LoadBalancerNetworkSecurityGroup": "[concat(variables('LoadBalancerName'), '_NetworkSecurityGroup')]",
-    "AnsibleController_name": "Ansible",
-    "CASController_name": "controller",
-    "Services_name": "services",
+
     "DiagnosticStorageGroupName": "[toLower(concat('rg4diag', variables('resourceGroupUniqueString')))]",
+    "DiagnosticStorageGroupName_long": "[concat('Microsoft.Storage/storageAccounts/', variables('DiagnosticStorageGroupName'))]",
```
bmoore-msft: use resourceId() instead of concat()

sas-viya/azuredeploy.json
```diff
@@ -101,20 +117,18 @@
...
     "AzureFilesViyaShare": "viyashare",
     "PrimaryUserName": "vmuser",
     "ExtensionAnsibleURI": "[concat(parameters('_artifactsLocation'), 'nestedtemplates/run_ansible_script.json', parameters('_artifactsLocationSasToken'))]",
     "NetworkCreateTemplateURI": "[concat(parameters('_artifactsLocation'), 'nestedtemplates/createNetworkSubtemplate.json', parameters('_artifactsLocationSasToken'))]",
     "AlternateDomain": "[concat('viya-', subscription().subscriptionId, '-', variables('resourceGroupUniqueString'))]",
     "DomainActual": "[ substring(variables('AlternateDomain'),0,if(lessOrEquals(length(variables('AlternateDomain')),60), length(variables('AlternateDomain')), 60))]",
     "Base64AdminPass": "[base64(parameters('SASAdminPass'))]",
-    "Base64UserPass": "[base64(parameters('SASUserPass'))]"
+    "Base64UserPass": "[base64(parameters('SASUserPass'))]",
+    "DUMMY_LICENSE_STRING": "<REPLACE THIS WITH A LINK TO YOUR LICENSE>"
```
bmoore-msft: can we make this a parameter?  And just name it licenseString (or something that doesn't say "dummy" - you can make the defaultValue whatever, e.g.

This is a dummy license, you will need to provide a valid license to run this workload in production

Or words to that effect...

sas-viya/azuredeploy.json
```diff
@@ -699,10 +713,10 @@
...
+            "value": "[concat('mkdir -p /var/log/sas/install && set -o pipefail; ./ansiblecontroller_startup.sh \"1\" \"', parameters('_artifactsLocation'), '\" \"', parameters('_artifactsLocationSasToken'), '\" \"', parameters('DeploymentDataLocation'),  '\" \"', variables('PrimaryUserName'),'\" \"', variables('Base64AdminPass'),'\" \"', variables('Base64UserPass'), '\" \"', variables('VirtualNetworkPrivateSubnetCIDR'), '\" \"', reference(resourceId('Microsoft.Network/publicIPAddresses', 'PrimaryViyaLoadbalancer_PublicIP')).dnsSettings.fqdn, '\" \"', parameters('DeploymentMirror'), '\" \"', variables('DiagnosticStorageGroupName'), '\" \"', variables('AzureFilesViyaShare'), '\" \"', listKeys(variables('DiagnosticStorageGroupName'),'2017-10-01').keys[0].value, '\" \"', parameters('CAS Node Count'), '\" \"', variables('DUMMY_LICENSE_STRING'), '\" 2>&1 | tee /var/log/sas/install/runAnsiblePhase1_SetupPrerequisites.log')]"
```
bmoore-msft: update the apiVersion for listKeys

sas-viya/azuredeploy.json
```diff
@@ -1153,13 +1167,169 @@
...
+      "properties": {
+        "publisher": "Microsoft.Azure.Extensions",
+        "type": "CustomScript",
+        "typeHandlerVersion": "2.0",
+        "autoUpgradeMinorVersion": true,
+        "settings": {
+          "skipDos2Unix": false
+        },
+        "protectedSettings": {
+          "commandToExecute": "[ concat('set -o pipefail; ./nonansiblecontroller_prereqs.sh \"', variables('PrimaryUserName'), '\" \"', variables('DiagnosticStorageGroupName'), '\" \"', variables('AzureFilesViyaShare'), '\" \"', listKeys(variables('DiagnosticStorageGroupName'),'2017-10-01').keys[0].value,'\" \"CasControllerServers\" 2>&1 | tee /tmp/prerequisites.log')]",
```
bmoore-msft: update apiVersion for listKeys

sas-viya/azuredeploy.json
```diff
@@ -1171,18 +1341,28 @@
           "skipDos2Unix": false
         },
         "protectedSettings": {
-          "commandToExecute": "[ concat('set -o pipefail; ./nonansiblecontroller_prereqs.sh \"', variables('PrimaryUserName'), '\" \"', variables('DiagnosticStorageGroupName'), '\" \"', variables('AzureFilesViyaShare'), '\" \"', listKeys(variables('DiagnosticStorageGroupName'),'2019-06-01').keys[0].value,'\" \"CasControllerServers\" 2>&1 | tee /tmp/prerequisites.log')]",
+          "commandToExecute": "[ concat('set -o pipefail; ./nonansiblecontroller_prereqs.sh \"', variables('PrimaryUserName'), '\" \"', variables('DiagnosticStorageGroupName'), '\" \"', variables('AzureFilesViyaShare'), '\" \"', listKeys(variables('DiagnosticStorageGroupName'),'2017-10-01').keys[0].value,'\" \"CasControllerServers\" 2>&1 | tee /tmp/prerequisites.log')]",
```
bmoore-msft: update apiVersion for listKeys()

sas-viya/metadata.json
```diff
@@ -5,6 +5,9 @@
...
   "summary": "This Quickstart is a reference architecture for users who want to deploy the SAS Viya platform, using microservices and other cloud-friendly technologies.",
   "githubUsername": "sassoftware",
-  "dateUpdated": "2019-12-04"
+  "dateUpdated": "2020-02-06",
+  "environments": [
+    "AzureCloud"
```
bmoore-msft: this doesn't work in Azure Gov?  (I thought it did)

sas-viya/azuredeploy.json
chlync: This variable is just the value compared against the valued in parameter 'DeploymentDataLocation'. I made it a variable because we do the comparison multiple times.

chlync: In theory, the only way they run into this string is if they submit the azuredeploy.parameters.json, which also has replace values for your CI system. This value causes it to not do a full deploy, but just a resources only test deploy.

bmoore-msft: Here's what I think would make it a bit cleaner:

- Set the defaultValue of the DeploymentDataLocation to an empty string
- remove the variable
- in your conditional checks for a valid license use:
    if(empty(parameters('DeploymentDataLocation;))...

The concern I have with the way it is written is that it looks like the user has to update the variable in the template to provide a valid license (and then the variable is still called "dummy license").  Using only the parameter makes it a bit clearer as to your intention for the users.

sas-viya/metadata.json
chlync: We do not have any resources that should disqualify us from us gov, but I do not have a test environment with access to US Gov. In running against your automated test, we ran into issues where the dns of the storage account is different from the standard. While this is a simple enough fix, without a test environment we cannot run this through QA, so for now I have to limit our support to the general cloud only.

sas-viya/azuredeploy.json
chlync: So, the issue we have with setting a default of "" is that that will cause it to allow an empty string submission of the DeploymentDataLocation on the Web UI without complaint. On the cli, it will also not fail, but run a full resource deploy and put in the outputs that this was a dummy deploy, which is not a great experience. Since you need a license for the deployment to fully release sas, we don't want to give the impression that you can submit without it, so we tried to make the dummy value suggest you needed to replace it with a valid license in the off chance that they encountered it.

sas-viya/azuredeploy.json
chlync: Change made

sas-viya/azuredeploy.json
chlync: Change Made

sas-viya/azuredeploy.json
chlync: Change Made

sas-viya/azuredeploy.json
chlync: change made

sas-viya/metadata.json
bmoore-msft: ok, lmk if you want to revisit or need help debugging...

sas-viya/azuredeploy.json
bmoore-msft: So you just want to make the parameter required?  (which it isn't if you have a defaultValue defined) - if so, then just use the empty string in the param file for the quickstart... then the UI will require it (since it doesn't use the param file)

sas-viya/azuredeploy.json
bmoore-msft: closed

sas-viya/azuredeploy.json
bmoore-msft: closed

sas-viya/azuredeploy.json
bmoore-msft: closed

sas-viya/azuredeploy.json
bmoore-msft: closed

sas-viya/azuredeploy.json
chlync: Updated as requested.

PR #7336, opened 2020-03-24
Update SQL MI template

jftl6y:
# Best Practice Checklist

## Changelog

* Update the template for revised management IPs, API versions and updated NSG rules
*
*


jftl6y:
@bmoore-msft I do not know if this is going to ever pass the check as it is timing out deploying the SQL Managed Instance.

bmoore-msft:
We can do a manual test - how do we know if it completed correctly?  The log shows "Accepted" and "Failed"

jftl6y:
@bmoore-msft I did a test deploy and it was failing because the GEN-PASSWORD token for the administrator password does not meet the password length/complexity requirements.  Generally I omit this parameter for a manual deploy (prompts the user) or pull in from a key vault for an automated deployment.  Thoughts/suggestions?

bmoore-msft:
I don't think that's what caused the CI failure (Accepted and Failed) but aside from that, how do we validate that this works?  We can either automate the test (and fix the password complexity issue (though it seems to work for other SQL workloads so I'm not sure that's related) or perform a manual test... if manual, how to we validate that this succeeds? since even after the timeout, the test deployment did not succeed...

jftl6y:
Close to re-submit.

PR #7473, opened 2020-04-29
Cosmos DB Free Tier template

markjbrown:
## Changelog

* New Template.
* Needs badges after merge.
*

### DIFF
--- a/101-cosmosdb-free/README.md
+++ b/101-cosmosdb-free/README.md
@@ -1,0 +1,10 @@
+# Create a free-tier Azure Cosmos DB account for Core (SQL) API
+
+This template will create a free-tier Azure Cosmos account for Core (SQL) API in a single region with Session level consistency and one database with 400 RU/s that can be shared with up to 25 containers. Accounts in free tier will not be billed for usage of 400 RU/s or 5GB of data or less.
+
+Below are the parameters which can be user configured in the parameters file including:
+
+- **Location:** Enter location for primary region.
+- **Database Name:** Enter the database name for the account.
+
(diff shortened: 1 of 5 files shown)


101-cosmosdb-free/README.md
```diff
@@ -0,0 +1,10 @@
+# Create a free-tier Azure Cosmos DB account for Core (SQL) API
```
bmoore-msft, line 1: Need to add the badges, see the sample readme or the pipeline task will have the exact MD you need to add.  AzDO pipelines are broken for PS Core error reporting so the task succeeds even though failures are detected.

101-cosmosdb-free/azuredeploy.parameters.json
```diff
@@ -0,0 +1,15 @@
+{
+  "$schema": "https://schema.management.azure.com/schemas/2019-04-01/deploymentParameters.json#",
+  "contentVersion": "1.0.0.0",
+  "parameters": {
```
bmoore-msft, line 4: remove the parameters for which you have defaultValues

101-cosmosdb-free/azuredeploy.parameters.us.json
```diff
@@ -0,0 +1,15 @@
+{
```
bmoore-msft, line 1: remove this file entirely since the defaults should work for Gov

101-cosmosdb-free/README.md
markjbrown: Added

101-cosmosdb-free/azuredeploy.parameters.json
markjbrown: Removed the default values from template to keep param files

101-cosmosdb-free/azuredeploy.parameters.us.json
markjbrown: So this will run for US Gov even without a "us" param file?

PR #7590, opened 2020-05-27
Update template for Azure ML workspace creation

guanyu-240:
# Best Practice Checklist

## Changelog

* Added AdbWorkspace property
* Remove Container registry provisioning since we only support ACR brought by customer
* Update the config for existing subnet to ensure it won't block private endpoint creation


101-machine-learning-create/azuredeploy.json
```diff
@@ -56,7 +56,7 @@
       },
       "storageAccountName": {
         "type": "string",
-        "defaultValue": "[concat('sa',uniqueString(resourceGroup().id))]"
+        "defaultValue": "[concat('sa',uniqueString(resourceGroup().id, deployment().name))]"
```
bmoore-msft, line 59: these won't be idempotent - what's the scenario you ran into for needing a change here?

guanyu-240: In order to ensure the uniqueness when using different deployment names.

guanyu-240: Without this change, if we deploy the second time(without deployment name parameter), the resource names are the same at sometime.

bmoore-msft: That's the intent - when deploying to the same RG you can redeploy idempotently.  If I want a new instance, I can change the param value, if I don't, this approach makes it difficult to do (and goes against the patterns in the repo.

So let's remove that from the seed

guanyu-240: Got it.

PR #7596, opened 2020-05-28
Windows

msangapu-msft:
We're working with Michael Blythe on creating ARM quickstarts for docs.microsoft.com/azure/app-service. This is the windows version. It's different than existing ARM templates in that, it has: a FREE SKU and flexibility to change programming languages easily in the template - which is a requirement for Quickstarts (day 1 scenarios). Please review and approve.

# Best Practice Checklist

## Changelog

*
*
*


msangapu-msft:
Hi @bmoore-msft  Please take a look when you have a moment.
We're working with Michael Blythe on creating ARM quickstarts for docs.microsoft.com/azure/app-service. This is the windows version. It's different than existing ARM templates in that, it has: a FREE SKU and flexibility to change programming languages easily in the template - which is a requirement for Quickstarts (day 1 scenarios). Please review and approve.

bmoore-msft:
@msangapu-msft - this (and the linux sample) are pretty much the same as the 101-webapp-basic* samples.  Any reason we can leverage those?

msangapu-msft:
> @msangapu-msft - this (and the linux sample) are pretty much the same as the 101-webapp-basic* samples. Any reason we can leverage those?

Thanks @bmoore-msft  for taking a look. I spent a couple days trying the existing templates, but couldn't use them for various reasons. For "quickstarts" we want as little changes (to the templates) as possible before deploying - in this case, there are no changes needed. The existing templates require modifications before deployment for Day 1 scenarios. We've kept these new templates as consistent as possible and can be easily modifiable to use other programming language stacks (.Net, Ruby, Node, PHP, Java, Python, etc.)

Here's some things off the top of my head:
* SKU needs to be free for Day 1 scenarios - customers shouldn't get charged for "trying" Azure products
* webAppName needs to be randomly generated
* siteConfig property is missing for Windows
* appSettings property is missing for Windows (Node.js on Windows uses this)
* location is set in advance with flexibility to change this through CLI
* programming language stacks can be changed easily (Java, Python, Node.js, and PHP have different property names on Windows)

Now if you're open to me replacing the existing Windows/Linux templates with these new ones, I'm totally up for it! Given the reasons, I believe these are streamlined for a wider number of customers.

Let me know your thoughts. Thank you for taking a look.

bmoore-msft:
I'm not talking about line by line differences, the spirit of the sample is the same.  Provision a "hello world" web app and we probably don't need a lot of those in the repo, just makes it hard for users to find.  We can easily add a site config property for it (as well as others that we need to convey how to best get started with the service).

I would prefer that we modify the existing ones since some set of users are already familiar with them.

@tfitzmac is looking into ways we can "tag" these samples to know they are docs - separating them isn't enough to ensure they stay relevant to the doc.

bmoore-msft:
I think SKU should be a parameter with a defaultValue for free...  But in general yes, a good doc sample is going to be a good sample for everyone.

msangapu-msft:
Thanks @bmoore-msft. One question - I checked the ARM Template documentation and unable to figure out how to do this:

Each language requires a slightly different siteConfig parameters. 

For example:
- PHP requires the phpVersion parameter.
- NodeJS requires an appSettings parameter.
- Java requires javaVersion, javaContainer, and javaContainerVersion parameters.

Depending on which language is chosen, different parameters are used. How can I handle this programmatically - so users don't have to modify the templates?

I looked at conditionals, if/then, user-defined functions, etc, but couldn't figure out. Do you know if this is possible?

bmoore-msft:
If you can "predefine" the related properties, you can use an object variable associated with single parameter to set the config.  It's a little more "sophisticated" than hello world but an extremely valuable construct for users to learn...  Here's a sample:

https://github.com/bmoore-msft/AzureRM-Samples/blob/master/leap-2020/modules/module.ultimate-vm.json#L271-L290

So user would select PHP/Node/Java via a single parameter (like you have in the doc) and then you can provide the variable that has the supporting properties.

msangapu-msft:
Thank you @bmoore-msft !! I didn't exactly that! Let me push the latest changes.

msangapu-msft:
@bmoore-msft Please review.

101-webapp-basic-windows/README.md
```diff
@@ -1,4 +1,5 @@
-# Deploy a app service plan and a basic Windows web app
+# This template allows you to deploy an App Service Plan and a Windows web app.
```
bmoore-msft, line 1: Change this to title "voice", e.g.

Windows App Service Plan and Basic Web App

101-webapp-basic-windows/azuredeploy.json
```diff
@@ -2,57 +2,87 @@
   "$schema": "https://schema.management.azure.com/schemas/2019-04-01/deploymentTemplate.json#",
   "contentVersion": "1.0.0.0",
   "parameters": {
-    "webAppName": {
-      "type": "string",
-      "metadata": {
-        "description": "Base name of the resource such as web app name and app service plan"
+      "webAppName": {
+          "type": "string",
+          "metadata": {
```
bmoore-msft, line 7: add a globally unique defaultValue using uniqueString

101-webapp-basic-windows/azuredeploy.json
```diff
@@ -2,57 +2,87 @@
...
+          "type": "string",
+          "defaultValue": "F1",
+          "metadata": {
+              "description": "The SKU of App Service Plan "
+          }
+      },
+      "language": {
+          "type": "string",
+          "defaultValue": "default",
+          "metadata": {
```
bmoore-msft, line 29: add allowedValues constraint

101-webapp-basic-windows/azuredeploy.json
```diff
@@ -2,57 +2,87 @@
...
+              "description": "The language stack of the app. Default = siteConfig. Java = javaSiteConfig. See objects below. "
+          }
+      },
+      "siteConfig": {
```
bmoore-msft, line 33: move both site config params to a variable like in the example I sent... using params creates a few problems:
1) object params are not easy for humans to manipulate so the "hellow world" simplicity is lost
2) these params and become incompatible with the language param and that will cause deployment to fail

101-webapp-basic-windows/azuredeploy.json
```diff
@@ -2,57 +2,87 @@
...
-      ]
-    }
+      {
+          "apiVersion": "2019-08-01",
+          "type": "Microsoft.Web/sites",
+          "kind": "app",
+          "name": "[parameters('webAppName')]",
+          "location": "[parameters('location')]",
+          "properties": {
+              "siteConfig": "[if( equals(parameters('language'),'java'),parameters('javaSiteConfig'),parameters('siteConfig') )]",
```
bmoore-msft, line 80: there should be no if() here, but a object parameter reference - see the sample I linked to

101-webapp-basic-windows/azuredeploy.json
```diff
@@ -2,57 +2,87 @@
...
+              "serverFarmId": "[resourceId('Microsoft.Web/serverfarms', variables('appServicePlanPortalName'))]"
+          },
+          "dependsOn": [
```
bmoore-msft, line 83: move this above properties

101-webapp-basic-windows/azuredeploy.parameters.us.json
```diff
@@ -0,0 +1,12 @@
+{
```
bmoore-msft, line 1: remove this file

101-webapp-basic-windows/azuredeploy.parameters.json
```diff
@@ -4,6 +4,9 @@
 	"parameters": {	
 		"webAppName": {
 			"value": "GEN-UNIQUE"
-		  }
+		  },
+		"location": {
```
bmoore-msft, line 8: remove this so the defaultValue can be used

msangapu-msft:
Hi @bmoore-msft Good points on the feedback. I made the requested changes. I will follow-up soon on Linux. BTW, I've been working on the existing template. This should be replacing the old one and not creating a new one. Let me know if you're seeing differently. 

One addition is default Git repo URLs from our samples from Docs. Thanks for your help!

101-webapp-basic-windows/azuredeploy.json
```diff
@@ -1,58 +1,146 @@
...
+            "name": "[parameters('webAppName')]",
+            "location": "[parameters('location')]",
+            "dependsOn": [
+                "[resourceId('Microsoft.Web/serverfarms', variables('appServicePlanPortalName'))]"
+            ],
+            "properties": {
+                "siteConfig": "[variables('configuration')]",
+                "serverFarmId": "[resourceId('Microsoft.Web/serverfarms', variables('appServicePlanPortalName'))]"
+            },
+            "resources": [
```
bmoore-msft, line 128: Why did we add this?  the original sample did not have an app deployed - and this changes the sample from an empty web app to one that's running a work load.

Also since the repo is not a param, the branch should not be either...

If we want to keep the repo option here we should provide a language option which is "none" and conditionally deploy the sourcecontrols resource.

Reverting to the original would also be fine if it fits with the doc.

101-webapp-basic-windows/metadata.json
```diff
@@ -1,10 +1,12 @@
...
-  "summary": "This template allows you to deploy an app service plan and a basic Windows web app",
-  "githubUsername": "cloudmelon",
-  "dateUpdated": "2018-07-31"
+  "itemDisplayName": "App Service Quickstart - Windows app",
+  "description": "This template is to be used with http://docs.microsoft.com/azure/app-service/quickstart-arm-template",
+  "summary": "It allows you to deploy an App Service Plan and a Windows app",
+  "githubUsername": "msangapu-msft",
+  "dateUpdated": "2020-05-28",
+  "environments": [
+    "AzureCloud"
```
bmoore-msft, line 10: why not Azure Gov?

101-webapp-basic-windows/azuredeploy.json
```diff
@@ -1,58 +1,146 @@
...
+        "gitRepoReference":{
+            ".net":"https://github.com/Azure-Samples/app-service-web-dotnet-get-started",
+            "node":"https://github.com/Azure-Samples/nodejs-docs-hello-world",
+            "php":"https://github.com/Azure-Samples/php-docs-hello-world",
+            "html":"https://github.com/Azure-Samples/html-docs-hello-world"
+        },
+        "configuration": "[variables('configReference')[toLower(parameters('language'))]]",
+        "configReference": {
+            ".net": {
+                "type": "object",
```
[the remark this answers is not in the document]
msangapu-msft: i've removed them, but it creates empty parameters. if i don't have these parameters, i'll need additional logic. i've added comments to get around this.

101-webapp-basic-windows/azuredeploy.parameters.us.json
msangapu-msft: The Gov build fails unless I use the location of 'usgovtexas'. Since you asked to remove this, I put a workaround to limit the test to the Azure Public. Let me know what you recommend.

101-webapp-basic-windows/azuredeploy.json
bmoore-msft: depending on the parameters used, yes the same webapp name will be generated - the uniqueString() function is idempotent.

msangapu-msft:
Hi @bmoore-msft 
Thank you for your help. There's two remaining items:
- You mentioned not to use Gov parameters, but the build fails for location 'southcentralus', what's the workaround here?
- I added an if condition to the repoUrl you can optionally deploy a Git repo. Let me know if there's a better way to do it.

Thanks!

101-webapp-basic-windows/azuredeploy.json
```diff
@@ -28,6 +28,7 @@
             "type": "string",
             "defaultValue": ".net",
             "allowedValues": [
+                "none",
```
msangapu-msft, line 31: "None" is not needed. App Service creates a .Net app when no language is specified. 
Also, if we allow a "none" value then we'll need a configReference["none"] variable. 
I can add it if you think "None" should be allowed.

101-webapp-basic-windows/azuredeploy.json
```diff
@@ -55,12 +49,13 @@
...
         "appServicePlanPortalName": "[concat('AppServicePlan-', parameters('webAppName'))]",
         "gitRepoReference":{
+            "none": "",
             ".net":"https://github.com/Azure-Samples/app-service-web-dotnet-get-started",
             "node":"https://github.com/Azure-Samples/nodejs-docs-hello-world",
             "php":"https://github.com/Azure-Samples/php-docs-hello-world",
             "html":"https://github.com/Azure-Samples/html-docs-hello-world"
         },
-        "gitRepoUrl":"[if ( bool(parameters('sample')), variables('gitRepoReference')[parameters('language')], parameters('repoUrl') ) ]",
+        "gitRepoUrl":"[if(empty(parameters('repoUrl')), variables('gitRepoReference')[parameters('language')], parameters('repoUrl'))]",
```
msangapu-msft, line 58: @bmoore-msft  The scenario of creating an empty PHP, Node, and HTML app is missed if we do it this way. That is not a requirement for Docs, but was doing it for template flexibility.

If there's further concerns, can we do it over a brief call on teams? I'd like to see if we can push the sample and Docs article this week. Thanks!

msangapu-msft:
@bmoore-msft Hi Brian, can we please get this checked-in today?

101-webapp-basic-windows/azuredeploy.json
tfitzmac: If it defaults to .NET, then it seems like allowing "none" is misleading. If someone set it to none and then later added an app, is it difficult to change the site config settings to use another language?

msangapu-msft: @tfitzmac In regards to the siteConfig - yes it can be difficult, ex: Node.js requries an app setting rather than a portal drop-down selection. It's not intuitive and the customer wouldn't know w/o searching for it. Having this within the template will ensure those customers can create apps w/ the correct configurations.

PR #7725, opened 2020-06-23
update for gov cloud

markjbrown:
## Changelog

* updated for Gov Cloud
* removed dedicated throughput option
*


markjbrown:
@bmoore-msft Hey Brian. This test shows a best practice issue for an unused parameter, "Name". But I don't have any variable named that.

bmoore-msft:
It's a bug in the BP test....

PR #7762, opened 2020-07-02
Fix misplaced plan in best practices sample.

mlindstr:
# PR Checklist

## Changelog

* Fixes a misplaced `plan` key which caused the template sample to not deploy.


bmoore-msft:
the "plan" property is a peer of "properties", see: https://docs.microsoft.com/en-us/azure/templates/microsoft.compute/2019-12-01/virtualmachines

It's used for marketplace images that are PAYG - if the image you're deploying is not that type of image, deployment will fail.

It's there a particular sample in the repo that's failing?

mlindstr:
`plan` is indeed a peer of `properties`, but the current sample puts it as a child of `properties` instead, which is the reason of my patch.

bmoore-msft:
Thanks... apparently my red and green was crossed...

PR #7783, opened 2020-07-10
Create blob storage datastore in AML

achalachal2003:
# PR Checklist

## Changelog

*
*
*

### DIFF
--- a/101-datastore-create-blob/azuredeploy.json
+++ b/101-datastore-create-blob/azuredeploy.json
@@ -14,7 +14,7 @@
 				"description": "The name of the datastore, case insensitive, can only contain alphanumeric characters and underscore"
 			}
 		},
-		"accountName": {
+		"storageAccountName": {
 			"type": "string",
 			"metadata": {
 				"description": "The name of the storage account."
@@ -28,26 +28,26 @@
 		},
 		"storageAccountSubscriptionId": {
 			"type": "string",
-			"defaultValue":"",
+			"defaultValue":"[subscription().id]",
 			"metadata": {
-				"description": "Optional : The subscription ID of the storage account."
+				"description": "Optional : The subscription ID of the storage account. Defaults to selected subscription"
 			}
 		},
 		"storageAccountResourceGroup": {
 			"type": "string",
-			"defaultValue":"",
+			"defaultValue":"[resourceGroup().name]",
 			"metadata": {
-				"description": "Optional : The resource group of the storage account."
+				"description": "Optional : The resource group of the storage account. Defaults to selected resource group"
 			}
 		},
-		"accountKey": {
+		"storageAccountKey": {
 			"type": "securestring",
 			"defaultValue":"",
 			"metadata": {
-				"description": "Optional : A storage account key, defaults to None. Skip this if SAS token authentication is used."
+				"description": "Optional : A storage account key, defaults to None. This would be ignored if SAS token authentication is provided."
 			}
 		},
-		"sasToken": {
+		"storageAccountSasToken": {
 			"type": "securestring",
 			"defaultValue":"",
 			"metadata": {
@@ -96,10 +96,10 @@
 			"properties": {
 				"DataStoreType": "blob",
 				"SkipValidation": "[parameters('skipValidation')]",
-				"AccountName": "[parameters('accountName')]",
+				"AccountName": "[parameters('storageAccountName')]",
 				"ContainerName": "[parameters('containerName')]",
-				"AccountKey": "[parameters('accountKey')]",
-				"SasToken": "[parameters('sasToken')]",
+				"AccountKey": "[parameters('storageAccountKey')]",
+				"SasToken": "[parameters('storageAccountSasToken')]",
 				"Protocol": "[parameters('protocol')]",
 				"Endpoint": "[parameters('endPoint')]",
 				"StorageAccountSubscriptionId": "[parameters('storageAccountSubscriptionId')]",
--- a/101-datastore-create-blob/azuredeploy.json
+++ b/101-datastore-create-blob/azuredeploy.json
@@ -1,0 +1,110 @@
+{
+	"$schema": "https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#",
+	"contentVersion": "1.0.0.0",
+	"parameters": {
+		"workspaceName": {
+			"type": "string",
+			"metadata": {
+				"description": "Specifies the name of the Azure Machine Learning workspace which will hold this datastore target."
+			}
+		},
+		"datastoreName": {
+			"type": "string",
+			"metadata": {
+				"description": "The name of the datastore, case insensitive, can only contain alphanumeric characters and underscore"
+			}
+		},
+		"accountName": {
+			"type": "string",
+			"metadata": {
+				"description": "The name of the storage account."
+			}
+		},
+		"containerName": {
+			"type": "string",
+			"metadata": {
+				"description": "The name of the azure blob container."
+			}
+		},
+		"storageAccountSubscriptionId": {
+			"type": "string",
+			"defaultValue":"",
+			"metadata": {
+				"description": "Optional : The subscription ID of the storage account."
+			}
+		},
+		"storageAccountResourceGroup": {
+			"type": "string",
+			"defaultValue":"",
+			"metadata": {
+				"description": "Optional : The resource group of the storage account."
+			}
+		},
+		"accountKey": {
+			"type": "string",
+			"defaultValue":"",
+			"metadata": {
+				"description": "Optional : A storage account key, defaults to None. Skip this if SAS token authentication is used."
+			}
+		},
+		"sasToken": {
+			"type": "string",
+			"defaultValue":"",
+			"metadata": {
+				"description": "Optional : An account SAS token, defaults to None. Skip this if storage account key authentication is used."
+			}
+		},
+		"endPoint": {
+			"type": "string",
+			"defaultValue":"",
+			"metadata": {
+				"description": "Optional : The endpoint of the storage account. Defaults to core.windows.net"
+			}
+		},
+		"protocol": {
+			"type": "string",
+			"defaultValue":"",
+			"metadata": {
+				"description": "Optional : Protocol to use to connect to the blob container. If not specified, defaults to https."
+			}
+		},
+		"skipValidation": {
+			"type": "bool",
+			"defaultValue":false,
+			"allowedValues": [
+				false,
+				true
+			],
+			"metadata": {
+				"description": "Optional : If set to true, the call will skip Datastore validation. Defaults to false"
+			}
+		},
+		"location": {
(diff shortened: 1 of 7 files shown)


achalachal2003:
@bmoore-msft for review.

There is a warning for passing in storageAccountAccess key in output. We have to pass storageAccountAccess from outside as template should not do ListKey for datastore creation. There is a valid use case for datastore to exist without accessKey or Sas.

101-datastore-create-blob/prereqs/prereq.azuredeploy.json
```diff
@@ -124,7 +124,7 @@
       "value": "[parameters('storageAccountName')]"
     },
 	"storageAccountAccountKey": {
-      "type": "string",
+      "type": "securestring",
```
bmoore-msft, line 127: you won't be able to do this... what you need to do is output the resourceId of the storage account and then use that in the listKeys() function in azuredeploy.json (to avoid passing secrets)

achalachal2003: There is a use case where customer can create datastore without storageKey or SAS token. If we do this in azuredeploy.json it will always set AccountKey even if customer don't want it. How can we handle this usecase? There is no way to know if customer want to set it or not.

bmoore-msft: Not sure I follow - right now we're passing a secret via outputs, that's a security issue so we're phasing it out of the repo.

If the property is optional, then you can use an empty string to determine the behavior and set the property value to json('null') (if that's what's needed).

If you want to set the property, then just pass the resourceId (or account name and resourcegroup name) into the template and call listKey() there.

That help?

achalachal2003: Thank you @bmoore-msft , I updated to change the desired behavior.

101-datastore-create-blob/azuredeploy.json
```diff
@@ -0,0 +1,110 @@
...
+			"defaultValue":"",
+			"metadata": {
+				"description": "Optional : A storage account key, defaults to None. Skip this if SAS token authentication is used."
+			}
+		},
+		"sasToken": {
+			"type": "securestring",
+			"defaultValue":"",
+			"metadata": {
+				"description": "Optional : An account SAS token, defaults to None. Skip this if storage account key authentication is used."
```
bmoore-msft, line 54: what happens if both are provided or neither is provided?

101-datastore-create-blob/azuredeploy.json
```diff
@@ -0,0 +1,110 @@
...
+		"storageAccountSubscriptionId": {
+			"type": "string",
+			"defaultValue":"",
+			"metadata": {
+				"description": "Optional : The subscription ID of the storage account."
+			}
+		},
+		"storageAccountResourceGroup": {
+			"type": "string",
+			"defaultValue":"",
```
bmoore-msft, line 38: set this to "[resourceGroup().name]"

101-datastore-create-blob/azuredeploy.json
```diff
@@ -0,0 +1,110 @@
...
+		},
+		"containerName": {
+			"type": "string",
+			"metadata": {
+				"description": "The name of the azure blob container."
+			}
+		},
+		"storageAccountSubscriptionId": {
+			"type": "string",
+			"defaultValue":"",
```
bmoore-msft, line 31: set this to "[subscription().id]"

101-datastore-create-blob/azuredeploy.json
```diff
@@ -0,0 +1,110 @@
...
+				"description": "Specifies the name of the Azure Machine Learning workspace which will hold this datastore target."
+			}
+		},
+		"datastoreName": {
+			"type": "string",
+			"metadata": {
+				"description": "The name of the datastore, case insensitive, can only contain alphanumeric characters and underscore"
+			}
+		},
+		"accountName": {
```
bmoore-msft, line 17: rename this to "storageAccountName"

101-datastore-create-blob/azuredeploy.json
```diff
@@ -0,0 +1,110 @@
...
+			}
+		},
+		"storageAccountResourceGroup": {
+			"type": "string",
+			"defaultValue":"",
+			"metadata": {
+				"description": "Optional : The resource group of the storage account."
+			}
+		},
+		"accountKey": {
```
bmoore-msft, line 43: rename to "storageAccountKey"

101-datastore-create-blob/azuredeploy.json
```diff
@@ -0,0 +1,110 @@
...
+			"type": "securestring",
+			"defaultValue":"",
+			"metadata": {
+				"description": "Optional : A storage account key, defaults to None. Skip this if SAS token authentication is used."
+			}
+		},
+		"sasToken": {
```
bmoore-msft, line 50: rename to "storageAccountSasToken"

101-datastore-create-blob/azuredeploy.json
```diff
@@ -0,0 +1,110 @@
...
+			"defaultValue":"",
+			"metadata": {
+				"description": "Optional : The endpoint of the storage account. Defaults to core.windows.net"
+			}
+		},
+		"protocol": {
+			"type": "string",
+			"defaultValue":"",
+			"metadata": {
+				"description": "Optional : Protocol to use to connect to the blob container. If not specified, defaults to https."
```
bmoore-msft, line 68: what are the options for this?  Would we ever not want to use https on Azure Storage?

101-datastore-create-blob/azuredeploy.json
```diff
@@ -0,0 +1,110 @@
...
+		},
+		"skipValidation": {
+			"type": "bool",
+			"defaultValue":false,
+			"allowedValues": [
+				false,
+				true
+			],
+			"metadata": {
+				"description": "Optional : If set to true, the call will skip Datastore validation. Defaults to false"
```
bmoore-msft, line 79: what does validation do?  (and why are we skipping it in CI?)

101-datastore-create-blob/azuredeploy.json
```diff
@@ -0,0 +1,110 @@
...
+			"defaultValue":"",
+			"metadata": {
+				"description": "Optional : An account SAS token, defaults to None. Skip this if storage account key authentication is used."
+			}
+		},
+		"endPoint": {
+			"type": "string",
+			"defaultValue":"",
+			"metadata": {
+				"description": "Optional : The endpoint of the storage account. Defaults to core.windows.net"
```
bmoore-msft, line 61: does the user have to provide this?  we could get it using reference or the environment() function...

101-datastore-create-blob/azuredeploy.json
achalachal2003: For both scenerio:  We ignore account key and only consider SAS token. Updated the metadata for "storageAcountKey" to reflect this.

For neither scenerio: We create credential-less datastore object. This is a valid customer scenario for datastore object.

101-datastore-create-blob/azuredeploy.json
achalachal2003: Don't think so, we expose this to customer in our other products like [CLI ](https://docs.microsoft.com/en-us/cli/azure/ext/azure-cli-ml/ml/datastore?view=azure-cli-latest#ext-azure-cli-ml-az-ml-datastore-attach-blob)so exposing it here too.

101-datastore-create-blob/azuredeploy.json
achalachal2003: This is optional for customers, we set defaults on our service side to avoid making changes on multiple clients. (UI, CLI, SDK)

101-datastore-create-blob/azuredeploy.json
achalachal2003: Thanks

101-datastore-create-blob/azuredeploy.json
achalachal2003: Thanks

101-datastore-create-blob/azuredeploy.json
achalachal2003: Flag skips connection settings & data validation in ADLS store, ADLS Gen2 file-syste,  file share, SQL, PostgreSQL, MySQL & blob container set up by customer. We would not be able to set it up for customer in CI. I am creating separate PR's for other type of stores.

achalachal2003:
Thanks @bmoore-msft for review and suggestions. Made the changes and left some comments for things we couldn't change. 

Going to make suggested changes in other PR's too that I am working on.

101-datastore-create-blob/azuredeploy.json
bmoore-msft: Let's add allowedValues for the allowed values and set an explicit default

101-datastore-create-blob/azuredeploy.json
bmoore-msft: I think we need to split the params for this:
1) to specify the authn type (none, sas, key)
2) one to contain the secureString (sasToken or Key)

And then use appropriate logic on the resource to set the properties as appropriate

101-datastore-create-blob/azuredeploy.json
bmoore-msft: What's blocking us from setting this up in CI?

achalachal2003: Could we upload data in blob through CI?

101-datastore-create-blob/azuredeploy.json
achalachal2003: ok going to change

101-datastore-create-blob/azuredeploy.json
achalachal2003: I am not sure of allowed values. Could we set explicit default?

bmoore-msft: Let's start with "http" and "https" then

101-datastore-create-blob/azuredeploy.json
bmoore-msft: We can do quite a bit - what's the requirement?  Do we need any random blob? specific blob?  CI already has some blobs available...

achalachal2003: We could use random blob but with valid data. Can we upload data in CI blobs manually? 

Please let me know how to use it.

101-datastore-create-blob/azuredeploy.json
achalachal2003: Fixed this

101-datastore-create-blob/azuredeploy.json
achalachal2003: Removed this as discussed in call

PR #7807, opened 2020-07-13
Azure Cognitive Search ARM template updates

HeidiSteen:
Changed existing strings.  Service name is now Azure Cognitive Search.  Added SKUs. Revised URL for schema reference doc.

### DIFF
--- a/101-azure-search-create/azuredeploy.json
+++ b/101-azure-search-create/azuredeploy.json
@@ -23,7 +23,7 @@
         "storage_optimized_l2"
       ],
       "metadata": {
-        "description": "The pricing tier of the search service you want to create. You can have one free service per subscription."
+        "description": "The pricing tier of the search service you want to create (for example, basic or standard)."
       }
     },
     "replicaCount": {
@@ -32,7 +32,7 @@
       "maxValue": 12,
       "defaultValue": 1,
       "metadata": {
-        "description": "Replicas distribute search workloads across the service. You need two or more to support high availability (applies to Basic and Standard only)."
+        "description": "Replicas distribute search workloads across the service. You need at least two replicas to support high availability of query workloads (not applicable to the free tier)."
       }
     },
     "partitionCount": {
@@ -47,7 +47,7 @@
       ],
       "defaultValue": 1,
       "metadata": {
-        "description": "Partitions allow for scaling of document count as well as faster indexing by sharding your index over multiple Azure Search units."
+        "description": "Partitions allow for scaling of document count as well as faster indexing by sharding your index over multiple search units."
       }
     },
     "hostingMode": {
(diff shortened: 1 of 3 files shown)


HeidiSteen:
#sign-off

HeidiSteen:
@bmoore-msft Hello, can you review/merge this PR?  It's mostly string changes per service name rebranding, plus a version bump on our management API.  Thank you!!

101-azure-search-create/azuredeploy.json
```diff
@@ -30,7 +32,7 @@
       "maxValue": 12,
       "defaultValue": 1,
       "metadata": {
-        "description": "Replicas distribute search workloads across the service. You need 2 or more to support high availability (applies to Basic and Standard only)."
+        "description": "Replicas distribute search workloads across the service. You need two or more to support high availability (applies to Basic and Standard only)."
```
miwelsh, line 35: Should we be more explicit about what "high availability" means here or are we comfortable with the other documentation clarifying that?

101-azure-search-create/azuredeploy.json
```diff
@@ -18,10 +18,12 @@
...
         "standard",
         "standard2",
-        "standard3"
+        "standard3",
+        "storage_optimized_l1",
+        "storage_optimized_l2"
       ],
       "metadata": {
-        "description": "The SKU of the search service you want to create. E.g. free or standard"
+        "description": "The pricing tier of the search service you want to create. You can have one free service per subscription."
```
miwelsh, line 26: Why are we explicitly stating "You can have one free service per subscription."?

101-azure-search-create/azuredeploy.json
HeidiSteen: Is this better? The revision mentions that "high availability" applies to queries:   "Replicas distribute search workloads across the service. You need two or more to support high availability for query workloads (applies to Basic and Standard only)".

101-azure-search-create/azuredeploy.json
HeidiSteen: It's the only tier that has this constraint so I thought we should mention it.  But on second thought, who is going to use a template to roll out a free tier?  No one.  I'll change it back to mentioning a few tier examples (basic or standard).

101-azure-search-create/azuredeploy.json
miwelsh: High availability means different things to different people. We provide guarantees around our quality of service based on the number of replicas so we may as well be explicit with that. I would say something more along the lines of "if you need read SLA then you need to have at least 2 replicas, if you need read/write SLA then you need to have at least 3 replicas."

101-azure-search-create/azuredeploy.json
miwelsh: It's fine to lead this in, especially if we're trying to clear up customer confusion around quota for free SKUs. I just don't have any context of why this was added.

PR #7826, opened 2020-07-16
UrlSigning quick start sample

gball14:
# PR Checklist

## Changelog

*
*
*

### DIFF
--- a/201-cdn-with-ruleseengine-urlsigning/README.md
+++ b/201-cdn-with-ruleseengine-urlsigning/README.md
@@ -1,0 +1,20 @@
+# Simple deployment of rewrite and redirect via rules engine for a CDN end point
+
+<IMG SRC="https://azurequickstartsservice.blob.core.windows.net/badges/201-cdn-with-ruleseengine-rewriteandredirect/PublicLastTestDate.svg" />&nbsp;
+<IMG SRC="https://azurequickstartsservice.blob.core.windows.net/badges/201-cdn-with-ruleseengine-rewriteandredirect/PublicDeployment.svg" />&nbsp;
+
+<IMG SRC="https://azurequickstartsservice.blob.core.windows.net/badges/201-cdn-with-ruleseengine-rewriteandredirect/FairfaxLastTestDate.svg" />&nbsp;
+<IMG SRC="https://azurequickstartsservice.blob.core.windows.net/badges/201-cdn-with-ruleseengine-rewriteandredirect/FairfaxDeployment.svg" />&nbsp;
+
+<IMG SRC="https://azurequickstartsservice.blob.core.windows.net/badges/201-cdn-with-ruleseengine-rewriteandredirect/BestPracticeResult.svg" />&nbsp;
+<IMG SRC="https://azurequickstartsservice.blob.core.windows.net/badges/201-cdn-with-ruleseengine-rewriteandredirect/CredScanResult.svg" />&nbsp;
+
+<a href="https://portal.azure.com/#create/Microsoft.Template/uri/https%3A%2F%2Fraw.githubusercontent.com%2FAzure%2Fazure-quickstart-templates%2Fmaster%2F201-cdn-with-ruleseengine-rewriteandredirect%2Fazuredeploy.json" target="_blank">
+    <img src="https://raw.githubusercontent.com/Azure/azure-quickstart-templates/master/1-CONTRIBUTION-GUIDE/images/deploytoazure.png"/>
+</a>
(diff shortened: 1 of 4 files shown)


201-cdn-with-ruleseengine-urlsigning/azuredeploy.json
```diff
@@ -0,0 +1,183 @@
...
+                "name": "origin1",
+                "properties": {
+                  "hostName": "[parameters('originUrl')]"
+                }
+              }
+            ],
+			"urlSigningKeys": [
+				{
+				  "keyId": "key1",
+				  "keySourceParameters":{
```
bmoore-msft, line 86: Seems like these values should be provided by the user?

gball14: @bmoore-msft , Yes. these should ideally be provided by customer pointing to their keyvault. As the repro try to submit the template to Azure RP, we created a sample key vault and referenced here. Let me know if there are better way to handle this. Users who use the template will not know what they keys are as they don't have permissions. Only Azure CDN APP (Principal) will have access to this key vault.

gball14: @bmoore-msft , Thanks for the clear ask. I updated the readme file. Please take a look.

gball14: Hi @bmoore-msft , we are waiting to share this example in our user documentation. Can you please look into this PR?

201-cdn-with-ruleseengine-urlsigning/README.md
```diff
@@ -18,3 +18,10 @@
 
 This template creates a CDN Profile and a CDN Endpoint with a user specified origin and all of our most commonly used settings on CDN. This template also configures rules engine UrlSigning action for default and override parameters.
 
+Before using the template, make sure to do the following:
+
+1. Create a key vault in the same subscription or in a different subscription.
+2. Define two secrets and set up secret values manually (at least 32 size)
```
bmoore-msft, line 24: does the secret value matter?

201-cdn-with-ruleseengine-urlsigning/README.md
```diff
@@ -18,3 +18,10 @@
...
+3. References the two secrets in urlSigningKeys of the template (change subscriptionId, resourceGroupName, vaultName, secretName, secretVersion)
+4. Make sure the keyvault is given "Get Permissions for Secret" access to "Microsoft.Azure.CDN" Principal.
```
bmoore-msft, line 26: how do I do this?  the principal does not exist in the tenant...

201-cdn-with-ruleseengine-urlsigning/azuredeploy.json
bmoore-msft: A couple comments on the readme - also, the values below need to be parameters for the template.

201-cdn-with-ruleseengine-urlsigning/README.md
gball14: @bmoore-msft , Please see below. 
image

201-cdn-with-ruleseengine-urlsigning/README.md
gball14: Min length is 32. We allow any value.

201-cdn-with-ruleseengine-urlsigning/README.md
bmoore-msft: the principal does not exist in the tenant...

201-cdn-with-ruleseengine-urlsigning/azuredeploy.json
bmoore-msft: These still need to be moved to parameters

gball14: Parametrized the urlsigning keys.

201-cdn-with-ruleseengine-urlsigning/README.md
gball14: @bmoore-msft , what tenant are you using? I checked with both COPR and AME tenant. I could see the principal.

gball14: You may need to register if it doesn't exist. https://docs.microsoft.com/en-us/azure/cdn/cdn-custom-ssl?tabs=option-2-enable-https-with-your-own-certificate

image

gball14: @bmoore-msft , can you please check if all looks good for you?

bmoore-msft: let's add this step to the readme... so others don't get hung up.  It looks like the SP is given contributor perms to the sub by default so I suspect that step will always be required.

Everything else looks good.

Issue #7767, opened 2020-07-06
Create subscription for 'Azure subscription as an Event Grid source'


realrubberduckdev:
[Create Event Grid subscription for resource events](https://github.com/Azure/azure-quickstart-templates/tree/master/101-event-grid-resource-events-to-webhook)

### Issue Details

The readme on the template above says `This template creates Event Grid subscription for either a resource group or Azure subscription` but it always creates a resource group subscription. We can see this on Azure portal on Event Subscriptions, that the subscription appears only on `Topic Type: Resource Groups`.

The ARM template [documentation](https://docs.microsoft.com/en-us/azure/templates/microsoft.eventgrid/2019-01-01/eventsubscriptions#property-values) only says

```
Use Microsoft.EventGrid/eventSubscriptions when subscribing to resource group events. Use <provider-namespace>/<resource-type>/providers/eventSubscriptions when subscribing to custom events or events for supported resource publishers.
```

But it has nothing on how to subscribe to [Azure subscriptions events](https://docs.microsoft.com/en-us/azure/event-grid/event-schema-subscriptions) for resources across the subscription. How do we create such a subscription using ARM templates?

PR #7872, opened 2020-07-28
Fix issue #7767

MCKLMT:
Fix issue #7767

- The template now deploys an Event Grid template with subscription events

### DIFF
--- a/101-event-grid-resource-events-to-webhook/azuredeploy.json
+++ b/101-event-grid-resource-events-to-webhook/azuredeploy.json
@@ -17,7 +17,10 @@
         },
         "systemTopicName": {
             "defaultValue": "systemTopicName",
-            "type": "String"
+            "type": "String",
+            "metadata": {
+                "description": "The name of the topic to create."
+            }
         }
     },
     "resources": [
@@ -27,7 +30,7 @@
             "name": "[parameters('systemTopicName')]",
             "location": "global",
             "properties": {
-                "source": "/subscriptions/5648d12a-ba49-4789-8793-6f7cc54e1556",
+                "source": "[concat('/subscriptions/', subscription().subscriptionId)]",
                 "topicType": "Microsoft.Resources.Subscriptions"
             }
         },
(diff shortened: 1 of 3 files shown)


MCKLMT:
@bmoore-msft I don't understand why the deployment of the template fails for this one because I can deploy it successfully on my environment. 🤷🏻‍♂️
Can you have a look please?

101-event-grid-resource-events-to-webhook/azuredeploy.json
```diff
@@ -14,27 +14,48 @@
...
         {
-            "type": "Microsoft.EventGrid/eventSubscriptions",
-            "name": "[parameters('eventSubName')]",
-            "apiVersion": "2018-01-01",
+            "type": "Microsoft.EventGrid/systemTopics",
+            "apiVersion": "2020-04-01-preview",
+            "name": "[parameters('systemTopicName')]",
+            "location": "global",
+            "properties": {
+                "source": "[concat('/subscriptions/', subscription().subscriptionId)]",
```
bmoore-msft, line 33: I think this is the change that's causing the problem... it's trying to elevate to subscription scope for some operation and the principal deploying only has permission to the resourceGroup - what's the intent of this change?

@tfitzmac - is this a doc sample? (where we care about the change) it's not marked as such in metadata but I think this sample pre-dates that approach

MCKLMT: The goal is to fix the issue mentioned. The previous template does not seem to deploy event grid topic for subscriptions but only for resource groups.

bmoore-msft: ok, the readme says "resource group or a subscription" - so if we can use the same template for either, then there should be a param for determining which should be done... and then in CI we'd choose the RG path.  If it's only for a subscription, then this needs to be moved to a subscription deployment since perms are needed at that scope to deploy it.

Still want @tfitzmac to chime in on how the sample is used in docs (if so)

tfitzmac: Yes, the doc owner should be spelluru

tfitzmac: Actually, it is referenced with the "deploy to Azure" button in the docs, but the code block is not displayed. In my opinion, we shouldn't add a doc owner for those cases, because the writer doesn't really need to be looped in for every change. However, I could be convinced otherwise.

bmoore-msft:
@MCKLMT can you take a look at the merge conflicts?

PR #7875, opened 2020-07-29
Fix SQL Server image versions

MCKLMT:
Fix SQL Server image versions on template https://github.com/azure/azure-quickstart-templates/tree/master/sqlvm-alwayson-cluster


MCKLMT:
@bmoore-msft The DS v3 series are not available in the region so the template deployment fails. Can we switch back to v2 series?

PR #7883, opened 2020-07-31
Add new deployment for ARM.

robinsh:
Add article and template for iot-hub-auto-route-messages. (missing screenshots)

### DIFF
--- a/101-iothub-auto-route-messages/README.md
+++ b/101-iothub-auto-route-messages/README.md
@@ -1,13 +1,3 @@
----
-title: Use an ARM template to create an Azure IoT Hub then send messages to the hub, routing them automatically to Azure Storage
-description: Use an ARM template to create an Azure IoT Hub then send messages to the hub, routing them automatically to Azure Storage
-author: robinsh
-ms.service: iot-hub
-services: iot-hub
-ms.author: robinsh
-ms.topic: conceptual
-ms.date: 07/31/2020
----
 # Quickstart: Deploy an Azure IoT Hub and a storage account using an ARM template
 
 In this quickstart, you use an Azure Resource Manager template (ARM template) to create an IoT Hub that will route messages to Azure Storage, and a storage account to hold them. After manually adding a virtual IoT device to the hub to submit the messages, you configure that connection information in the arm-read-write application to submit messages to the hub using that device. The hub has the routing configured, so the messages sent to the hub are automatically routed by the hub to the storage account. At the end of this quickstart, you can open the storage account and see the messages sent.
--- a/101-iothub-auto-route-messages/README.md
+++ b/101-iothub-auto-route-messages/README.md
@@ -41,9 +41,10 @@ Two Azure resources are defined in the template:
 
 1. Create the resources by deploying the ARM template.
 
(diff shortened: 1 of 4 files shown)


MCKLMT:
The deployment fails.
I think you have to fix a function name: https://github.com/robinsh/azure-quickstart-templates/blob/robinsh-arm-hub/101-iothub-auto-route-messages/azuredeploy.json#L14

Replace
```
subscription().subscriptionid
```
By
```
subscription().subscriptionId
```

robinsh:
When trying to publish that quickstart, I get errors, so I assumed it wasn’t published. I’ll take a look at it today. Thanks for letting me know.
Robin

MCKLMT:
@robinsh Let me know if you need help with the template.

robinsh:
The template actually works, at least for me. “The Powers That Be” edited it after I submitted it, so I’ll have to deploy it and check for errors (like the one you found). This is part of a “horizontal article” coming showing how to use that template to create the resources and then run an application to send messages to the hub, and see them be routed to storage. The real problem is getting the template pushed to the repo, as it gives a lot of strange error messages that I haven’t figured out how to fix yet. That’s why I was surprised to see that it’s in the repo. What good is it to give fatal error messages if it’s going to publish anyway??

robinsh:
Can you let me know where you see that? I looked at azure/azure-quickstart-templates and don’t see it in there. Is it in the private repo?
Robin

MCKLMT:
image
In the Azure Pipeline

robinsh:
Doesn’t the little red x on that line mean it failed? Why would you pull it if it failed? (Genuinely interested.)

MCKLMT:
Each template should follow best practices which are tested before the deployment. 
In your case, the deployment also fails.

robinsh:
Then why can you see it? I wouldn’t think it would how a deployment that failed. I know I’m bad at GIT, but generally that’s true!

robinsh:
So here's the deal. If you haven't published the template, and you refer to it in your readme file or in the how-to article, it won't build because your article can't find the template If you publish the template and parameter files first, thinking you will publish the how-to article later, they won't publish unless you have all the files, including a readme. This push should have the template, the parameters file, the metadata json, and a readme file (which is not the how-to unless a bunch of the other templates). Hopefully this will work.

MCKLMT:
Why opening multiple pull requests? The regular process is to open a pull request from a single branch and update this branch with your modifications. You have now 4 pull requests opened for the same template, right?

robinsh:
Thank you for your feedback.

Robin

robinsh:
Did you put in PRs against my branch(es) ? Are you mcklmt? I would appreciate it in the future if you would not edit my work while I’m still working on it. Seriously, dude.

MCKLMT:
Apologies it was a mistake.

robinsh:
Abandon in favor of the lasttry branch.

robinsh:
Ok. I have half-committed changes that I can’t see in the repo, but they come out if I merge something else in there. I’ve removed most of the extras now, and hopefully am working on the (final) changes.

Robin

Issue #7768, opened 2020-07-06
Update dokku quickstart template


benaboki:
[Deploy a Dokku Instance](https://github.com/azure/azure-quickstart-templates/tree/master/dokku-vm)

### Issue Details
The template fails to run. Version number is outdated. Deployment fails with the following messages:

- "code": "VMExtensionProvisioningError",
- "message": "VM has reported a failure when processing extension 'initdokku'. Error message: \"Enable failed: failed to execute command: command terminated with exit status=100
- The following signatures couldn't be verified because the public key is not available: NO_PUBKEY FB2B6AA421CD193F
- Unable to locate package dokku\n\"\r\n\r\nMore information on troubleshooting is available at https://aka.ms/VMExtensionCSELinuxTroubleshoot

### Solution
* change Dokku version from 0.10.3 to 0.14.2

HamzaBoukraa:
@benaboki :
Will be fixed in a PR today

benaboki:
Just tested and saw that the dokku version is still 0.10.3. I spent hours trying to troubleshoot this and the error messages I was getting were not helpful.

bmoore-msft:
thx @MCKLMT

PR #7886, opened 2020-08-01
Fix issue #7768

MCKLMT:
Fix issue #7768
- Update dokku releases base url
- Update dokku version to latest


MCKLMT:
@bmoore-msft The deployment fails in AzDo but is successful with the deploy button. Can you have a look please?

Issue #6897, opened 2019-11-26
wrong urls in readme.md 201-aci-sftp-files-existing-storage


Jonatan-V:
[201-aci-sftp-files-existing-storage](https://github.com/Azure/azure-quickstart-templates/blob/master/201-aci-sftp-files-existing-storage/README.md)

### Issue Details
The deploy to azure link is pointing to the form for creating an on-demand SFTP Server with persistent storage, but should point to the form On-demand SFTP Server using an existing storage account
I think the correct URL should be: 

https://portal.azure.com/#create/Microsoft.Template/uri/https%3A%2F%2Fraw.githubusercontent.com%2FAzure%2Fazure-quickstart-templates%2Fmaster%2F201-aci-sftp-files-existing-storage%2Fazuredeploy.json

PR #7887, opened 2020-08-02
Fix issue #6897

MCKLMT:
Fix issue #6897
- Fix URLs in README.md


MCKLMT:
@bmoore-msft The deployment fails on US Gov. Can you have a look?

MCKLMT:
@bmoore-msft Can you explain what's your change?

bmoore-msft:
The container in the prereq template wasn't creating the fileshare in USGov - not sure why since it didn't fail and worked in Public...  So I changed the prereq template to use the deploymentScript resource instead of a container, which seems to work as expected.  Also, using ACI for the prereq leaves the resource intact, so the deploymentScript resource is a more appropriate solution.

bmoore-msft:
ok, looks like deployment is working everywhere... @MCKLMT any concerns merging this?

The BP failure is coming from the fact that the apiVersion is not rolled into USGov yet (even though it's 4 months old).

MCKLMT:
OK to be merged.

PR #7909, opened 2020-08-06
Fix typos on media-service-output-primary-key

MCKLMT:
# PR Checklist

## Changelog

* Fix typos on media-service-output-primary-key
* Update API version


MCKLMT:
@bmoore-msft I need help on this one.

* The goal of the template is to demonstrate the ability to create a Media Services resource and return its primary key. However, this is not a best practice and returned as an error in the CI. What should we do?
* The API version (2020-07-01) is listed as too old but it's the last [according the documentation](https://docs.microsoft.com/en-us/azure/templates/microsoft.media/mediaservices).
* An error is returned at the deployment ```Azure Active Directory does not support Media Services account keys.```. I think this is related to the ```listkeys``` function because [the documentation](https://docs.microsoft.com/en-us/azure/azure-resource-manager/templates/template-functions-resource#implementations) does not mention it's possible to use listkeys function with Media Services.

Thanks for your thoughts

PR #8027, opened 2020-08-27
Adelghabboun/anfnfsv3vol

AdelGhabboun:
# PR Checklist

## Changelog

*
*
*


101-anf-nfs-volume/azuredeploy.json
```diff
@@ -0,0 +1,186 @@
...
+  "parameters": {
+    "location": {
+      "type": "string",
+      "defaultValue": "[resourceGroup().location]",
+      "metadata": {
+        "description": "Same Location of resource group for all resources in the same deployment."
+      }
+    },
+    "netAppAccountName": {
+      "defaultValue": "test-anf-acc",
```
bmoore-msft, line 13: will this work for all users (i.e. it doesn't need to be unique within the subscription or globally unique?)

101-anf-nfs-volume/azuredeploy.json
```diff
@@ -0,0 +1,186 @@
+{
+  "$schema": "https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#",
+  "contentVersion": "1.0.0.0",
+  "parameters": {
```
bmoore-msft, line 4: add metadata descriptions for all parameters

101-anf-nfs-volume/azuredeploy.parameters.json
```diff
@@ -0,0 +1,51 @@
+{
+  "$schema": "https://schema.management.azure.com/schemas/2015-01-01/deploymentParameters.json#",
+  "contentVersion": "1.0.0.0",
+  "parameters": {
```
bmoore-msft, line 4: remove the parameters from here, for which you have defaultValues in the template

bmoore-msft (changes requested):

101-anf-nfs-volume/azuredeploy.json
AdelGhabboun: Name unique within a subscription only

AdelGhabboun: So it will work for all users

101-anf-nfs-volume/azuredeploy.json
AdelGhabboun: added

101-anf-nfs-volume/azuredeploy.json
bmoore-msft: if it's required to be unique within a subscription, then you can only deploy one instance of this template to a sub (using the defaultValues)...  so if that's required (not sure if that's what you meant because the scenario is odd) we need to make the defaultValue unique

AdelGhabboun: got it fixed. Now every resources that needs to be unique within a sub will have a different name and template can be deployed multiple times

PR #8295, opened 2020-09-29
Update API Version on 201-vm-specialized-vhd-new-or-existing-vnet

MCKLMT:
# PR Checklist

## Changelog

* Update API version
* Add US Gov deploy button
* Minor fixes


MCKLMT:
@bmoore-msft I don't understand the error for this one. If you have an idea...

bmoore-msft:
Not sure - that property is a property on the disk resource when doing an import, but the docs aren't clear about when it's needed: 
 https://docs.microsoft.com/en-us/azure/templates/microsoft.compute/2019-11-01/disks#CreationData

This also seems to mirror other samples using a uri (and some of those work ;))

I tried a new build though not confident it will make a difference.

MCKLMT:
Solved by downgrading the API version for the disk. [According to the documentation](https://docs.microsoft.com/en-us/azure/templates/microsoft.compute/2019-11-01/disks#CreationData), the storageUri parameters is required on recent API version. It was optional before.

bmoore-msft:
Any idea what we need to do to get a working version on the new apiVersion?  Would be good to have a sample, if we can't figure it out I suspect others are going to struggle as well ;)

MCKLMT:
To get it to work with the last API version you'll have to add both _storageAccountId_ AND _sourceUri_.
In our case with the CI I think we don't have the storage account information where is stored the VHD. Correct?
Any way to do revert from the VHD uri?

bmoore-msft:
We do have the stuff needed for a resourceId for the storageAccount:

https://github.com/Azure/azure-quickstart-templates/blob/master/test/ci-gen-setup/.config.json#L15-L16

bizarre that you need both - but let's try that...  I'm wondering if it's going to need permission to that storageAccount in the control plane (that we may need to fix).

MCKLMT:
> We do have the stuff needed for a resourceId for the storageAccount:
> 
> https://github.com/Azure/azure-quickstart-templates/blob/master/test/ci-gen-setup/.config.json#L15-L16
> 
> bizarre that you need both - but let's try that... I'm wondering if it's going to need permission to that storageAccount in the control plane (that we may need to fix).

@bmoore-msft We have a permission issue on the storage account containing the VHD. The service principal deploying the template does not have the needed permissions.

MCKLMT:
@bmoore-msft The error message for the build
```Status Message: The client 'cdba50f2-9a94-471c-a12f-ae3ac06b2cfb' with object id 'cdba50f2-9a94-471c-a12f-ae3ac06b2cfb' has permission to perform action 'Microsoft.Compute/disks/write' on scope '/subscriptions/d0f4ed43-cdbb-432e-b9f1-14a9fae84201/resourcegroups/azdo-cd6c59b3-a206-4ab4-8a05-3b999e0a9801/providers/Microsoft.Compute/disks/newvm-disk'; however, it does not have permission to perform action 'Microsoft.Storage/storageAccounts/listKeys/action' on the linked scope(s) '/subscriptions/d0f4ed43-cdbb-432e-b9f1-14a9fae84201/resourceGroups/ttk-gen-artifacts/providers/Microsoft.Storage/storageAccounts/azbotstorageus' or the linked scope(s) are invalid. (Code:LinkedAuthorizationFailed)```

may1393 (approved):

may1393 (approved):

bmoore-msft:
Ok... if you're still out there.

It's odd that you would need listKeys permission to read a public blob - strange change in the new apiVersion... IDK if you want to give this a try with another/newer apiVersion but if it still requires listKeys permission to read the vhd, that seems like a bug.

MCKLMT:
@bmoore-msft Still the same issue. I think this is a bug

bmoore-msft:
Ok, if this new apiVersion requires contributor perms to the storageAccount then we need a new sample (or more changes to this one to prereq that requirement (create the account, copy the image to it).  So I'm going to close this one, feel free to reopen (with the prereqs) or just do a new one (with the prereqs)

PR #8426, opened 2020-10-15
Removes hard coded App Insights location

kyleburnsdev:
Allows Application Insights to be deployed to the same location as the other resources. Existing code appears to be from when App Insights was not widely available

# PR Checklist

## Changelog

*
*
*

### DIFF
--- a/101-function-premium-vnet-integration/azuredeploy.json
+++ b/101-function-premium-vnet-integration/azuredeploy.json
@@ -13,7 +13,7 @@
             "type": "string",
             "defaultValue": "",
             "metadata": {
-                "description": "Location for Application Insights"
+                "description": "Location for Application Insights. Leave empty for auto-selection."
             }
         },
         "appInsightsDefaultLocation": {
--- a/101-function-premium-vnet-integration/azuredeploy.json
+++ b/101-function-premium-vnet-integration/azuredeploy.json
@@ -12,7 +12,7 @@
         "appInsightsLocation": {
             "type": "string",
             "metadata": {
-                "description": "Location for Application Insights."
+                "description": "Location for Application Insights." 
             }
         },
         "runtime": {
--- a/101-function-premium-vnet-integration/azuredeploy.json
+++ b/101-function-premium-vnet-integration/azuredeploy.json
@@ -61,10 +61,6 @@
         "applicationInsightsName": "[parameters('appName')]",
         "storageAccountName": "[concat(uniquestring(resourceGroup().id), 'azfunctions')]",
         "functionWorkerRuntime": "[parameters('runtime')]",
-        "insightsLocation": {
-            "AzureCloud": "eastus",
-            "AzureUSGovernment": "usgovvirginia"
-        },
         "appInsightsResourceId": "[resourceId('Microsoft.Insights/components', variables('applicationInsightsName'))]"
     },
     "resources": [
@@ -113,7 +109,7 @@
         {
             "type": "Microsoft.Insights/components",
             "apiVersion": "2018-05-01-preview",
-            "location": "[variables('insightsLocation')[environment().name]]",
+            "location": "[parameters('location')]",
             "name": "[variables('applicationInsightsName')]",
             "kind": "web",
             "properties": {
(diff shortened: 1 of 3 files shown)


mcollier:
If this template is deployed to a region that does not yet support Application Insights, will it fail?  

Alternative, could there be a separate parameter, such as `applicationInsightsLocation` (with a default value of `resourceGroup().location`)?  This would allow the user to override the region for Application Insights to ensure they can Application Insights to a supported region.

kyleburnsdev:
> If this template is deployed to a region that does not yet support Application Insights, will it fail?
> 
> Alternative, could there be a separate parameter, such as `applicationInsightsLocation` (with a default value of `resourceGroup().location`)? This would allow the user to override the region for Application Insights to ensure they can Application Insights to a supported region.

@mcollier - reviewing the supported regions list, I think you're right. I'll update the PR

mcollier:
LGTM

101-function-premium-vnet-integration/azuredeploy.json
```diff
@@ -6,7 +6,14 @@
             "type": "string",
             "defaultValue": "[resourceGroup().location]",
             "metadata": {
-                "description": "Location for all resources."
+                "description": "Location for all resources except Application Insights."
+            }
+        },
+        "appInsightsLocation": {
+            "type": "string",
+            "defaultValue": "[resourceGroup().location]",
```
bmoore-msft, line 14: AI is not globally available so you can't have a default that can resolve to any location.  You'll need to add param files for Public and Gov to set it.

bmoore-msft (changes requested):

101-function-premium-vnet-integration/azuredeploy.json
mcollier: Ah, so an azuredeploy.parameters.us.json file - https://github.com/Azure/azure-quickstart-templates/blob/master/1-CONTRIBUTION-GUIDE/best-practices.md#sample-application-files

And then not set a default value for `appInsightsLocation`.

kyleburnsdev: @bmoore-msft  - I see what you're getting at, but also am concerned about the usability of removing the default. Most regions currently have AI or will very soon, so there's a balance to be considered on placing the burden of specifying the value (and associated risk of having it be cross region when specified) against the risk that someone on an unsupported region will initially submit the deployment accepting the default and have to re-submit with an updated value.

Do you have thoughts on this balance?

bmoore-msft: a defaultValue should never cause a deployment to fail and it's only available in about half the locations, so...  we don't want to use something that's not very likely to work.

kyleburnsdev: @bmoore-msft I've updated to favor the current deployment region while falling back to the original behavior when that region is not known to have Application Insights available

bmoore-msft: Sorry but I think we've done a 180 here... the original template was simple and could not fail due to user input.  

Now we're asking not only to select a location for AI to supply a few other params including a json array of all the locations that are supported (which is not easy to find).  This is far more complex.

I can think of a few options here:
1) the original template
2) add a distinct param for AI location that the user must supply
3) Option 2 but make the param optional and use the approach in the original template to supply the value when not provided.

kyleburnsdev: Option 3 is pretty much the approach used here. The valid values are provided in the cloud-specific param file and the user does not have to supply. 

- If the user specifies a value for AI location, it is used. 
- If the user does not supply a value and the location is in the list of valid values, the location is used (reducing latency and egress exposure).
- If the user does not supply a value and the location is _not_ in the list of valid values, the values from the original template are used as a default

The only change that the user of the template experiences is the ability to optionally specify a location for AI without breaking any current scripts that may have been created

bmoore-msft: It's not really the same as what I'm suggesting - the approach in the current template has 2 additional required parameters that need to be supplied by the user and even then it's not very clear which location is actually going to be used... also, the parameter files is not part of the template...  This needs to be simplified if you want to merge the PR.

PR #8447, opened 2020-10-20
Check API versions

mgblythe:
# PR Checklist

## Changelog

*
*
*


bmoore-msft:
@mgblythe - can you take a look at the pipeline result and advise on how customers fix that error?

mgblythe:
@bmoore-msft - some parts of gov require an earlier API version. Should we scope to public only then I can explain that you can use an earlier API version for gov?

bmoore-msft:
It's better to have Gov support than not, so if it's possible we should target the sample to support it.  Better to not pass BP checks than not support a cloud.  Also, eventually that should be fixed and then we have to remember to come back and fix these.

mgblythe:
@bmoore-msft - I left the latest API version in and I added a note to the readme that you might require an earlier version for gov.

bmoore-msft:
But that's the opposite of what I was suggesting - does this work in FF with the older API version?

mgblythe:
@bmoore-msft - yeah I see that now that I re-read :laughing: Closing this, will leave as old version for now.

PR #8468, opened 2020-10-28
Configure private DNS zone

farehar:
# PR Checklist

## Changelog

*
*
*

### DIFF
--- a/101-cloud-shell-vnet/azuredeploy.json
+++ b/101-cloud-shell-vnet/azuredeploy.json
@@ -81,7 +81,11 @@
         "relayNamespaceRef": "[resourceId('Microsoft.Relay/namespaces/', parameters('relayNamespaceName'))]",
         "relaySubnetRef": "[resourceId('Microsoft.Network/virtualNetworks/subnets', parameters('existingVNETName'), parameters('relaySubnetName'))]",
         "contributorRoleDefinitionId": "[resourceId('Microsoft.Authorization/roleDefinitions', 'b24988ac-6180-42a0-ab88-20f7382dd24c')]",
-        "networkRoleDefinitionId": "[resourceId('Microsoft.Authorization/roleDefinitions', '4d97b98b-1d4f-4787-a291-c67834d212e7')]"
+        "networkRoleDefinitionId": "[resourceId('Microsoft.Authorization/roleDefinitions', '4d97b98b-1d4f-4787-a291-c67834d212e7')]",
+        "privateDnsZoneName": "privatelink.servicebus.windows.net",
+        "networkRef": "[resourceId('Microsoft.Network/virtualNetworks', parameters('existingVNETName'))]",
+        "privateIpAddress": "10.0.2.6",
+        "privateEndpointRef": "[resourceId('Microsoft.Network/privateEndpoints', parameters('privateEndpointName'))]"
     },
     "resources": [
         {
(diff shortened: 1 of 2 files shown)


101-cloud-shell-vnet/azuredeploy.json
```diff
@@ -81,12 +81,15 @@
         "relayNamespaceRef": "[resourceId('Microsoft.Relay/namespaces/', parameters('relayNamespaceName'))]",
         "relaySubnetRef": "[resourceId('Microsoft.Network/virtualNetworks/subnets', parameters('existingVNETName'), parameters('relaySubnetName'))]",
         "contributorRoleDefinitionId": "[resourceId('Microsoft.Authorization/roleDefinitions', 'b24988ac-6180-42a0-ab88-20f7382dd24c')]",
-        "networkRoleDefinitionId": "[resourceId('Microsoft.Authorization/roleDefinitions', '4d97b98b-1d4f-4787-a291-c67834d212e7')]"
+        "networkRoleDefinitionId": "[resourceId('Microsoft.Authorization/roleDefinitions', '4d97b98b-1d4f-4787-a291-c67834d212e7')]",
+        "privateDnsZoneName": "privatelink.servicebus.windows.net",
```
bmoore-msft, line 85: this cannot be hardcoded - you can pass it in as a param

101-cloud-shell-vnet/azuredeploy.json
```diff
@@ -232,19 +235,86 @@
...
+            "name": "[concat(variables('privateDnsZoneName'), '/@')]",
+            "dependsOn": [
+                "[resourceId('Microsoft.Network/privateDnsZones', variables('privateDnsZoneName'))]"
+            ],
+            "properties": {
+                "ttl": 3600,
+                "soaRecord": {
+                    "email": "azureprivatedns-host.microsoft.com",
+                    "expireTime": 2419200,
+                    "host": "azureprivatedns.net",
```
bmoore-msft, line 283: this cannot be hardcoded

101-cloud-shell-vnet/azuredeploy.json
```diff
@@ -232,19 +235,86 @@
...
+            "type": "Microsoft.Network/privateDnsZones/SOA",
+            "apiVersion": "2018-09-01",
+            "name": "[concat(variables('privateDnsZoneName'), '/@')]",
+            "dependsOn": [
+                "[resourceId('Microsoft.Network/privateDnsZones', variables('privateDnsZoneName'))]"
+            ],
+            "properties": {
+                "ttl": 3600,
+                "soaRecord": {
+                    "email": "azureprivatedns-host.microsoft.com",
```
bmoore-msft, line 281: remove this or make it a param

bmoore-msft (changes requested):

101-cloud-shell-vnet/azuredeploy.json
farehar: removed SOA record as it'll be automatically generated.

101-cloud-shell-vnet/azuredeploy.json
farehar: removed SOA record as it'll be automatically generated.

101-cloud-shell-vnet/azuredeploy.json
```diff
@@ -232,19 +235,66 @@
...
+            "name": "[variables('privateDnsZoneName')]",
+            "location": "global",
+            "properties": {
+                "maxNumberOfRecordSets": 25000,
+                "maxNumberOfVirtualNetworkLinks": 1000,
+                "maxNumberOfVirtualNetworkLinksWithRegistration": 100,
+                "numberOfRecordSets": 2,
+                "numberOfVirtualNetworkLinks": 1,
+                "numberOfVirtualNetworkLinksWithRegistration": 0,
+                "provisioningState": "Succeeded"
```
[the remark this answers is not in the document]
farehar: removed.

101-cloud-shell-vnet/azuredeploy.json
bmoore-msft: Line 85 - the privateDnsZoneName

farehar: @bmoore-msft: If I make it a parameter and users plug in a different value, this wouldn't work anymore. Is there a way to make the paramter ready only?

bmoore-msft: If it can only be one possible value, why is it required that the user supply it?

farehar: User is not required to supply it. That is why I had it declared as a variable instead of parameter.

bmoore-msft: Does it have to be provided in the PUT request?

farehar: yes, it's required for PUT request

bmoore-msft: That's the part that doesn't make sense - it's required on the PUT but it cannot be changed.  If it cannot be changed, why is it required on the PUT?

farehar: We need to create a private DNS zone for our vnet. the name needs to be privatelink.servicebus.windows.net as we need to add a dns record for relay.privatelink.servicebus.winodws.net -> private ip address. If we make it a parameter and user changes it to something else , it won't be translated correctly. Does this make sense?

bmoore-msft: Yes, it sounds like that property value can only be the exact string you've specified.  If it cannot be changed, why does it have to be specified in the template?

farehar: @bmoore-msft : I don't understand. If we don't specify it in template how can i create a private dns zone resource with that name?

bmoore-msft: I think the part I'm struggling with is why we have an API in Azure that requires a name that contains an Azure endpoint (and it cannot be changed).  That's a strange/error prone pattern and  will prevent that template from being used across clouds.

farehar: @bmoore-msft : Private DNS zones can be created for different resources and would have a different value for different scenarios. For out particular scenario, it needs to be privatelink.servicebus.windows.net but it can be different depending on what resources you have.

PR #8589, opened 2020-12-08
DPS fix for USGov region

wesmc7777:
# PR Checklist

## Changelog

* Updates template files for [quickstart article](https://docs.microsoft.com/azure/iot-dps/quick-setup-auto-provision-rm).
* Adds quickstart article's author to docOwner in metadata.json.
* Updates README.md with link to the quickstart article.
* Added min TLS 1.2 property for US Gov Virginia but rolled that back. We don't want to force this on all regions.


bmoore-msft:
@wesmc7777 - check this PR for updates that may be needed to documentation that references this sample.

wesmc7777:
#sign-off

wesmc7777:
@bmoore-msft  @davidsmatlak  This should be ready for merging

wesmc7777:
I've rolled back the change that fixed the problem in USGovVirginia. That region requires TLS 1.2 property (`"minTlsVersion" : "1.2"`) as shown below. 

```
  "resources": [
    {
      "type": "Microsoft.Devices/IotHubs",
      "apiVersion": "2020-03-01",
      "name": "[parameters('iotHubName')]",
      "location": "[parameters('location')]",
      "sku": {
        "name": "[parameters('skuName')]",
        "capacity": "[parameters('skuUnits')]"
      },
      "properties": {
        "minTlsVersion" : "1.2"
      }
    },
```

However, we don't want to force all regions to use TLS 1.2. I'll just mention this property requirement in the ARM template doc

bmoore-msft:
@wesmc7777 - check this PR for updates that may be needed to documentation that references this sample.

wesmc7777:
Testing to see if we can use a condition to resolve this in the template

bmoore-msft:
@wesmc7777 - check this PR for updates that may be needed to documentation that references this sample.

wesmc7777:
@bmoore-msft I believe we can merge this. Let me know what you think

bmoore-msft:
@wesmc7777 - check this PR for updates that may be needed to documentation that references this sample.  [this is an automated message]

101-iothub-device-provisioning/azuredeploy.json
```diff
@@ -62,7 +62,7 @@
                 "capacity": "[parameters('skuUnits')]"
             },
             "properties": {
-                "minTlsVersion": "[if(equals(environment().name, 'AzureUSGoverment'), parameters('minTLS'), json('null'))]"
+                "minTlsVersion": "[if(equals(environment().name, 'AzureUSGovernment'), parameters('minTLS'), json('null'))]"
```
wesmc7777, line 65: @bmoore-msft This works but, TLS 1.2 is only required for `usgovvirginia`. It is **not** required for `usgovarizona`.

Pretty sure `AzureUSGovernment` is comprised of both  `usgovvirginia` and `usgovarizona`.  Seems like security requirements would be the same across both but it's not.

wesmc7777 (commented):
This works but, TLS 1.2 is only required for `usgovvirginia`. It is not required for `usgovarizona`.

Pretty sure `AzureUSGovernment` is comprised of both `usgovvirginia` and `usgovarizona`. Seems like security requirements would be the same across both but it's not.

101-iothub-device-provisioning/azuredeploy.json
bmoore-msft: Ok, that's a little strange... but if those are the rules, I'm inclined to say we should just doc it, instead of build it into the template.  Thoughts?

Also thinking about the scenario where I want to set the min tls version but am running in another cloud/region...

wesmc7777: I definitely have to document this.  But, that pipeline failure is a irritating to me.  My vote would be for just leaving `usgovvirginia` specified since that is a definite requirement. This way its clear in the docs and the template

bmoore-msft: We try to avoid whenever possible hard-coding locations in the templates.  Eventually... that will cause a failure when it shouldn't...  I took a look at the README and it's not clear to me what's needed.  I took a look at the [current] doc linked to in the README and it doesn't mention anything there either...

Soooo.... Here's what I think you're saying the requirement is...  

> Only TLS 1.2 will work in conjuction with DPS from other clouds/regions. iot Hubs still supports TLS 1.0 and 1.1 but DPS does not.

When DPS is provisioned, TLS 1.2 must be used.  This template unconditionally deploys DPS, so TLS 1.2 would always be required for this template.  If this is true we can just hard-code TLS minVersion to 1.2.

If, OTOH, it's only ```usgovvirginia``` that has this requirement, we should state that in the docs and let the use select the TLS version since the template may not be deployed to ```usgovvirginia``` or the use may still want to use 1.2 even though it's not required.

Am I following this correctly?

wesmc7777: No. A few clarifications:

* DPS supports using TLS (and recommends enabling it) to restrict device access but it's not required
* If TLS is enabled, DPS only supports TLS 1.2
* IoT Hubs support TLS (and recommends enabling it) to restrict device access but it's not required except in `usgovvirginina`
* IoT Hubs support TLS 1.0, 1.1, and 1.2

So, TLS will always be required when deploying IoT hubs in `usgovvirginia`. If DPS is being used with the hub then TLS 1.2 must be used.

I understand that we would want to avoid hard-coding a location. However, I would think this would be a good exception as it would help clarifiy and reinforce to the customer that this is not optional. As a starter template in the hands of a customer, I'd think customer's would appreciate a template that helps reinforce this without them having to run into problems with this. 

I'll update the doc today to reflect this. Let me know what you decide for this repo and I'll fix the PR

wesmc7777: One slight correction to the above. Based on the information I have from the PM, IoT hubs in `usgovvirginia` must support TLS 1.2. 

So no TLS 1.0 or 1.1 support in `usgovvirginia` location

bmoore-msft: Ok, thanks for the matrix...  I had a bit of a chuckle

Here's what I think we should do:
1. template should require TLS 1.0, 1.1, 1.2 - I know IoT doesn't require it but we should since it's recommended.  If someone doesn't want TLS they can use a different sample or modify the template
1. user selects a TLS version via param
1. if location is usgovvirgina (sigh...) change the param value to 1.2
1. doc all of this in the README.MD - even the nit that TLS isn't required but this sample uses it

I can make the changes to the template if you want to make the doc changes (or you can do both if that's easier)

bmoore-msft:
@wesmc7777 - check this PR for updates that may be needed to documentation that references this sample.  [this is an automated message]

bmoore-msft:
@wesmc7777 - check this PR for updates that may be needed to documentation that references this sample.  [this is an automated message]

wesmc7777:
@bmoore-msft So how can we set the `requireTLS` parameter to `true` for the `usgovvirginia` pipeline test without conditionally hardcoding it ?

101-iothub-device-provisioning/azuredeploy.json
wesmc7777: Sorry, I just seen this.  I don't want to mention TLS 1.0 and 1.1 in this template because this is a DPS quickstart template.  Things will not work if they go that route.

So I just set a parameter to enable TLS

bmoore-msft:
Check my last commit...  essentially:

1. always set TLS (secure by default, if you want to disable use another sample (or edit this one)
1. force to 1.2 in usgovvirginia (yes we have to hardcode it)

I think that will pass the checks...

wesmc7777:
TLS 1.0 and 1.1 are not valid options with DPS.  Device won't be able to connect

wesmc7777:
Also, defaulting to having TLS turned on is probably not a good idea.  Too many legacy devices that may not support TLS 1.2.  It's because of legacy devices that IoT Hub has not been able to get away from TLS 1.0 and 1.1.  We should default to leaving TLS off and have the user turn it on if their devices support TLS 1.2.  Otherwise, our DPS quickstart templates will deploy resources that a large portion of our user can not connect their devices to

101-iothub-device-provisioning/azuredeploy.json
```diff
@@ -35,17 +35,22 @@
...
             }
         },
-        "requireTLS": {
+        "minTlsVersion": {
             "type": "bool",
-            "defaultValue": false,
+            "defaultValue": "1.2",
+            "allowedValues": [
+                "1.0",
+                "1.1",
```
wesmc7777, line 43: 1.0 and 1.1 won't be valid options in a DPS quickstart template

bmoore-msft:
@wesmc7777 - check this PR for updates that may be needed to documentation that references this sample.  [this is an automated message]

101-iothub-device-provisioning/azuredeploy.json
```diff
@@ -58,7 +63,7 @@
                 "capacity": "[parameters('skuUnits')]"
             },
             "properties": {
-              "minTlsVersion": "[if(parameters('requireTLS'), variables('minTlsVersion'), json('null'))]"
+              "minTlsVersion": "[variables('minTlsVersion')]"
```
wesmc7777, line 66: Now we have hard-coded TLS support where only 1.2 will possibly work and a large part of our customers may have legacy devices that can't connect because of no device TLS support.  We should default to TLS off

bmoore-msft:
@wesmc7777 - check this PR for updates that may be needed to documentation that references this sample.  [this is an automated message]

101-iothub-device-provisioning/azuredeploy.json
bmoore-msft: ok, if this template always deploys DPS and only 1.2 is allowed for DPS, then we should just set TLS to 1.2 all the time - no conditionals based on location (or anything else)...  what am I missing?

101-iothub-device-provisioning/azuredeploy.json
bmoore-msft: Ok, let's just see if we can create a README.md that users can understand.  Once that's clear we can figure out the template.  Here's my 8th attempt at trying to understand this:

-This template always deploys DPS
-DPS only supports TLS 1.2
-DPS does not require TLS except in usgovvirginia (at least for now)

IoTHub support for TLS is somewhat orthogonal since this template uses DPS and that has stricter requirements.

bmoore-msft:

PR #8591, opened 2020-12-09
Add kentico-xperience-environment template

chladekm:
# PR Checklist

## Changelog

* Added new template - folder kentico-xperience-environment


bmoore-msft (changes requested):
Check the pipeline results for some changes that need to be made -- will do a full review after those changes.

kentico-xperience-environment/README.md
```diff
@@ -0,0 +1,99 @@
...
+
+1. Develop an [MVC](https://devnet.kentico.com/CMSPages/DocLinkMapper.ashx?version=latest&link=mvc_development) or [Core](https://devnet.kentico.com/CMSPages/DocLinkMapper.ashx?version=latest&link=core_section_root) Xperience project locally.
+2. Use this template to deploy resources necessary to host the Xperience site in Azure (click *Deploy to Azure* at the top of this document).
+	- See the **Deployed resources** section for an overview of all resources deployed by the template.
+3. [Publish](https://devnet.kentico.com/CMSPages/DocLinkMapper.ashx?version=latest&link=kentico_azure_webapps) the Xperience administration and front-end applications to the deployed Web Apps.
+4. Upload the project's database to a database created under the Azure SQL server instance deployed by the template. 
+
+For more information about site development and general best practices, visit the [Xperience documentation](https://docs.xperience.io/).
+
+## Deployment steps
```
bmoore-msft, line 74: remove this section, it's the same for every sample in the repo

kentico-xperience-environment/azuredeploy.json
```diff
@@ -0,0 +1,326 @@
...
+      "defaultValue": "[resourceGroup().location]",
+      "metadata": {
+        "description": "The geographical location to which the resources will be deployed."
+      }
+    }
+  },
+  "resources": [
+    {
+      "apiVersion": "2018-05-01",
+      "name": "pid-e3ec0034-3eaf-47a7-bf36-21bcf341b4dd",
```
bmoore-msft, line 113: is this registered?

kentico-xperience-environment/azuredeploy.json
```diff
@@ -0,0 +1,326 @@
...
+      }
+    },
+    {
+      "name": "[parameters('sqlserverName')]",
+      "type": "Microsoft.Sql/servers",
+      "location": "[parameters('location')]",
+      "tags": {
+        "displayName": "SqlServer"
+      },
+      "apiVersion": "2020-02-02-preview",
```
bmoore-msft, line 131: See: https://github.com/Azure/azure-quickstart-templates/blob/master/1-CONTRIBUTION-GUIDE/best-practices.md#sort-order-of-properties for the proper sort order of properties

fix this for all resources

bmoore-msft (changes requested):
Also check the pipeline results for changes that need to be made

kentico-xperience-environment/azuredeploy.json
```diff
@@ -0,0 +1,323 @@
...
+        "mode": "Incremental",
+        "template": {
+          "$schema": "https://schema.management.azure.com/schemas/2018-05-01/Microsoft.Resources.json",
+          "contentVersion": "1.0.0.0",
+          "resources": []
+        }
+      }
+    },
+    {
+      "type": "Microsoft.Sql/servers",
```
bmoore-msft, line 122: How is this used?  It seems to be unconditionally created, but there's no database... so when the db name is provided as a parameter, the connectionString will be set to this sql server, but no database.  Seems like we need to make this deployment conditional and add a database to the template?

bmoore-msft (changes requested):

kentico-xperience-environment/azuredeploy.json
chladekm: It's by design. We assume developers who use this template already have an instance of database (probably from local installation of our app). So it's up to them to provide the database backup.

bmoore-msft: Why not provide the back in the template then?

chladekm: We don't want to provide a database backup as part of the template, because the database is dependent on the version of the deployed app. Therefore, it would cause issues.

bmoore-msft: Where is the version of the app specified?

chladekm: Information about version of the app is available in both, the app and the database. The application verifies correct version of the database on startup.

PR #8613, opened 2020-12-17
Update 101-event-grid to use 'scope' property

anthony-c-martin:
# PR Checklist

## Changelog

*
*
*


bmoore-msft:
@johndowns - we have a few of these eventGrid samples that are failing due to http://requestb.in being unreachable - can you advise on a fix for these?

johndowns:
@bmoore-msft I had hard-coded a RequestBin URL into the parameters file because RequestBin URLs were implicitly trusted by Event Grid and didn't need an interactive handshake to validate them. However, it looks like the requestb.in domain was shut down recently. There is a replacement RequestBin service, but Event Grid doesn't appear to automatically trust it like it did with RequestBin.

This means there are a few options I can see:
1. We can switch to using a different event sink that doesn't require the validation handshake (e.g. we could use a storage queue instead of a webhook). This will let the template work, but it changes the scope of what the sample does.
2. We can take the webhook URL as a parameter and not have a default value. This would still be useful as a way to document how to use Event Grid with a webhook this in an ARM template, but it will break the template validation for this repo since it need to do a deployment of the template and this will need to have some human intervention in order to make it work. I'm not sure if there is a way to override this for special cases.
3. We can remove this template from the quickstart repo and put it into the Event Grid docs instead.

Do you have a preference?

bmoore-msft:
@johndowns - Is there an option 4 where we can replicate what requestb.in did?  TBH I have no idea what that endpoint needs to do but seems like we could host something in the subscription where we run CI that would do what it did?

johndowns:
@bmoore-msft The problem is that Event Grid's webhook validation procedure requires the server to either echo back a validation code in a specific response format, or make an outbound call. Simple services like RequestBin don't do this. Previously the Event Grid team specifically authorised RequestBin URLs to skip this validation procedure.

I can check with the Event Grid PG to see if there's another service they have also allowed to skip the validation process, but if not then I don't think we can do this very easily.

bmoore-msft:
ok, checking with them sounds good...  though I'm wondering how customers would do this in the real-world and seems like there *should* be a way to emulate that (and at the same time show them how)....

johndowns:
@bmoore-msft I'll reach out and see if I can find out anything. Replicating the real-world behaviour might be possible with an Azure Function, but this would be a bit of work to create a fully inline function within the template and plumb it through. Having an external URL would be a lot better IMHO.

bmoore-msft:
If we can do a function, I think we'd just spin it up in the sub and leave it running - so it's doesn't have to be part of the sample.

johndowns:
@bmoore-msft I have just had a response from the Event Grid team. We can replace the URL with a hookb.in URL and it will behave the same way. I've created one: `https://hookb.in/oXROgWrZ3lS1mmLaR1ek`. We can replace this in all of the template parameter files that use it.

How would you like to handle this? Should I do a separate PR to update all of these?

bmoore-msft:
if there are only a few, then a separate PR for each is better since we'll run the pipeline on them...  If that's too laborious (i.e. there are more than a few) then a single PR will work, just don't put any other changes in with it ;)

johndowns:
@bmoore-msft I've submitted https://github.com/Azure/azure-quickstart-templates/pull/8686 to fix the `101-event-grid` sample.

You mentioned that there are "a few" failing because of this - I searched through the samples and can only see that one. Are there any others you know of?

bmoore-msft:
@johndowns - ok I can't find any either - at least not by searching for requestb.in...  I guess ignore that for now, hopefully will remember the new site if there are more lurking.

bmoore-msft:
@johndowns - ok, well not exactly requestb.in, but the new one you created is returning a 404 (see the pipeline results)

johndowns:
@bmoore-msft That's a shame. It looks like hookb.in only keeps their endpoints alive for a certain amount of time.

I think the easiest resolution to this will be for me to update the example template to also deploy an Azure Functions app, which will act as the event sink. I'll try to do this in the next few days.

may1393 (approved):

bmoore-msft:
@johndowns - what are your latest thoughts on this?  function app in a prereq template?

johndowns:
@bmoore-msft I've already done this, and I included the changes that @anthony-c-martin had in this PR - so I think this PR can be closed.

Check out https://github.com/Azure/azure-quickstart-templates/tree/master/101-event-grid to see

bmoore-msft:
LGTM

PR #8615, opened 2020-12-17
Update 101-redis-cache to use 'scope' property

anthony-c-martin:
# PR Checklist

## Changelog

*
*
*

### DIFF
--- a/101-redis-cache/azuredeploy.json
+++ b/101-redis-cache/azuredeploy.json
@@ -92,9 +92,9 @@
       },
       "resources": [
         {
+          "scope": "[format('Microsoft.Cache/redis/{0}', parameters('redisCacheName'))]",
           "type": "Microsoft.Insights/diagnosticsettings",
           "apiVersion": "2017-05-01-preview",
-          "scope": "[format('Microsoft.Cache/redis/{0}', parameters('redisCacheName'))]",
           "name": "[parameters('redisCacheName')]",
           "location": "[parameters('location')]",
           "dependsOn": [
--- a/101-redis-cache/azuredeploy.json
+++ b/101-redis-cache/azuredeploy.json
@@ -92,9 +92,10 @@
       },
       "resources": [
         {
-          "type": "Microsoft.Cache/redis/providers/diagnosticsettings",
+          "type": "Microsoft.Insights/diagnosticsettings",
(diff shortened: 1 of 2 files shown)


bmoore-msft:
@yegu-ms - check this PR for updates that may be needed to documentation that references this sample.  [this is an automated message]

bmoore-msft:
@yegu-ms - check this PR for updates that may be needed to documentation that references this sample.  [this is an automated message]

bmoore-msft:
@yegu-ms - check this PR for updates that may be needed to documentation that references this sample.  [this is an automated message]

101-redis-cache/azuredeploy.json
```diff
@@ -92,9 +92,10 @@
       },
       "resources": [
         {
-          "type": "Microsoft.Cache/redis/providers/diagnosticsettings",
+          "scope": "[format('Microsoft.Cache/redis/{0}', parameters('redisCacheName'))]",
```
bmoore-msft, line 95: doesn't look like we can use scope on a nested child resource...

bmoore-msft (changes requested):

101-redis-cache/azuredeploy.json
anthony-c-martin: @bmoore-msft - should I just refactor this so that it isn't a nested child? It's not currently taking advantage of the shortened nested child type syntax.

bmoore-msft: yeah I think we should just promote it so we can use the scope property for these samples - aside, seems like we should fix it in the template engine?

anthony-c-martin: Possibly? To me that implies that you can nest a resource which isn't actually a child using this syntax. I imagine the only scenario we would actually want to support is scope == the parent resource id (e.g. you're deploying an extension resource)? Worth a discussion!

bmoore-msft: or we just get rid of nesting ;)

bmoore-msft:
@yegu-ms - check this PR for updates that may be needed to documentation that references this sample.  [this is an automated message]

bmoore-msft:
@yegu-ms - check this PR for updates that may be needed to documentation that references this sample.  [this is an automated message]

PR #8704, opened 2021-01-29
Cross-region load balancer with regional load balancers

asudbring:
# PR Checklist

- [ x] - Please check this box once you've submitted the PR if you've read through the Contribution Guide and best practices checklist.

## Changelog

*
*
*


asudbring:
#sign-off

101-load-balancer-cross-region/azuredeploy.json
```diff
@@ -0,0 +1,890 @@
...
+    "parameters": {
+        "projectName": {
+            "type": "string",
+            "metadata": {
+                "description": "Specifies a project name that is used for generating resource names."
+            }
+        },
+        "location-cr": {
+            "type": "string",
+            "allowedValues": [
```
bmoore-msft, line 13: why does this have allowedValues but the other locations do not?

101-load-balancer-cross-region/azuredeploy.json
```diff
@@ -0,0 +1,890 @@
...
+                }
+            }
+        },
+        {
+            "type": "Microsoft.Compute/virtualMachines",
+            "apiVersion": "2020-06-01",
+            "name": "[concat('vm-r2','-',copyIndex(1))]",
+            "location": "[parameters('location-r2')]",
+            "zones": [
+                "[copyIndex(1)]"
```
bmoore-msft, line 510: seems this template will fail if deployed to a region that doesn't support zones?

bmoore-msft (changes requested):

101-load-balancer-cross-region/azuredeploy.json
asudbring: This one is for the cross region load balancer, which is only available in select regions.  The other two are normal load balancers that are available everywhere.

101-load-balancer-cross-region/azuredeploy.json
asudbring: I got this part of the template from the published template for deploying a standard load balancer which has zones.
https://docs.microsoft.com/en-us/azure/load-balancer/quickstart-load-balancer-standard-public-template

bmoore-msft: @mumian - do you have a preference on how to handle these?  We can doc it or add allowedValues...

101-load-balancer-cross-region/azuredeploy.json
bmoore-msft: Let's add some of that detail to the metadata description

asudbring: Edited the metadata.json and updated the description to include that the cross region load balancer is available in limited regions.  I also added that the regional load balancers can be in any region.

101-load-balancer-cross-region/azuredeploy.json
mumian: The article has this part - "The template deployment creates three availability zones. Availability zones are supported only in certain regions. Use one of the supported regions. If you aren't sure, enter centralus."

I will add the description to the quickstart template.

asudbring:
@bmoore-msft @mumian 
Can this be merged?  It's been sitting out here for more than a week.

asudbring:
@bmoore-msft @mumian
Can I get a status on getting this merged?

Thanks!

mumian:
@bmoore-msft - I hope you were not waiting on me.  I have finished updating the other quickstart. Let me know if there is anything I need to do.

bmoore-msft:
The template uses (requires) availability zones, which are not globally available, so we need to set allowedValues on the location params as we did for crlb.

asudbring:
@bmoore-msft
Made requested changes, please merge.

Thank you!!

PR #8873, opened 2021-03-10
101-kusto-vnet: Update following subnet delegation support

basaba:


bmoore-msft:
@basaba - check the pipeline results for changes that need to be made

PR #8875, opened 2021-03-10
Update sample - Function App with Azure Storage private endpoints

gabesmsft:
# PR Checklist

## Changelog

* fixed broken image link in README.md
* removed GET-PREREQ-identityName defaultValue for identityName in azuredeploy.json and moved it to azuredeploy.parameters.json
* updated Microsoft.Network/privateDnsZones resources to use previously unused privateStorage*DnsZoneName variables (eg privateStorageFileDnsZoneName)
* rolled the two deploymentscripts resources into one via conditional & comparison functions (the script will execute either az powershell, az cli, or not be invoked, depending upon what is specified in the postDeploymentScript parameter

### DIFF
--- a/101-function-app-storage-private-endpoints/azuredeploy.parameters.json
+++ b/101-function-app-storage-private-endpoints/azuredeploy.parameters.json
@@ -4,6 +4,9 @@
   "parameters": {
     "identityName": {
       "value": "GET-PREREQ-identityName"
+    },
+    "identityNameResourceGroup": {
+      "value": "GET-PREREQ-identityNameResourceGroup"
     }
   }
 }
--- a/101-function-app-storage-private-endpoints/azuredeploy.parameters.json
+++ b/101-function-app-storage-private-endpoints/azuredeploy.parameters.json
@@ -2,5 +2,8 @@
   "$schema": "https://schema.management.azure.com/schemas/2015-01-01/deploymentParameters.json#",
   "contentVersion": "1.0.0.0",
   "parameters": {
+    "identityName": {
+      "value": "GET-PREREQ-identityName"
+    }
   }
 }
--- a/101-function-app-storage-private-endpoints/README.md
+++ b/101-function-app-storage-private-endpoints/README.md
@@ -85,5 +85,5 @@ In order to run the prereq template, you must have Microsoft.Authorization/roleA
 #### Using the deployment script in the main template
 
 To execute the inline deployment script in the main template:
-- For the postDeploymentScript parameter, specify either "azpowershell" or "azcli" depending on which modules your deployment environment has. If you leave the value as "none", the inline deployment script won't get executed.
+- For the postDeploymentScript parameter, specify either "azpowershell" or "azclibash" depending on which modules your deployment environment has. If you leave the value as "none", the inline deployment script won't get executed.
 - For the "identityName", specify an existing user-assigned managed identity that meets the prerequisites described above. By default, the "identityNameResourceGroup" parameter assumes that the managed identity is in the same resource group that the main template is being deployed to.
--- a/101-function-app-storage-private-endpoints/README.md
+++ b/101-function-app-storage-private-endpoints/README.md
@@ -70,20 +70,12 @@ After the first time you configure the Function App to talk to the private endpo
 
 The script will retrieve the site-level credentials from the publish profile and make an authenticated request to the Kudu /Debugconsole page of the Function App.
 
+To run the deployment script, set the postDeploymentScript parameter to either "azpowershell" or "azclibash" depending upon your deployment environment. If you set the value as "none", the inline deployment script won't get executed.
+
 #### Prerequisites to run the in-line deployment script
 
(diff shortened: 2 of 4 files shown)


101-function-app-storage-private-endpoints/azuredeploy.parameters.json
```diff
@@ -2,5 +2,8 @@
   "$schema": "https://schema.management.azure.com/schemas/2015-01-01/deploymentParameters.json#",
   "contentVersion": "1.0.0.0",
   "parameters": {
+    "identityName": {
```
bmoore-msft, line 5: you need to also output the resourceGroup name of the MSI in prereq.azuredeploy.json and then set the value here (the deploymentScript scenario isn't working)

bmoore-msft (changes requested):

101-function-app-storage-private-endpoints/azuredeploy.parameters.json
gabesmsft: I believe I will need to put the roledef and role assignment into the main template in order for this to be able to pass the tests, unless I am missing another way to do this:
I see that the prereqs and main template are deployed to different resource groups.
The roledef would need to be assigned to the identity either at the subscription level (which isn't possible via the principal that is used in the tests) or in the scope of the main template's rg or resources (which isn't predeterminable via the tests). 

The reason I had put the role assignment in the prereqs template was to allow for the possibility that the deployer of the main template doesn't have Microsoft.Authorization/roleAssignments/write perms (so that someone with these perms could do the role assignment separately), but I can roll these up into the main template instead.

I will also make the "azPowerShellVersion": "5.5", "azCliVersion": "2.20.0" changes as well (I overlooked that comment in my last run).

bmoore-msft: your design of putting the MSI as a prereq is spot-on... you just need to specify the resourceGroup parameter on the resourceId function that uses the MSI.  So your prereq template needs to output the name and the resourceGroup name - so in prereqs.json

outputs
 - name
 - resourcegroup

in azuredeploy.json
parameters
 - identityname
 - identityrg
...
resourceId(parameters(identityrg), Microsoft.ManagedIdentities/identities, parameters(identityname))

gabesmsft: So I did [these steps](https://github.com/Azure/azure-quickstart-templates/pull/8875#discussion_r592760854) in merge [b26d2a6](https://github.com/Azure/azure-quickstart-templates/pull/8875/commits/b26d2a671a0b14211a566135e6da628529f1eb92) , but Build [#48346](https://dev.azure.com/azurequickstarts/azure-quickstart-templates/_build/results?buildId=48346&view=logs&j=275f1d19-1bd8-5591-b06b-07d489ea915a&t=add333c3-6afd-5d0a-7233-d67f8d9bd855) failed due to the identity not having perms to do the necessary actions in the rg used by the main template (I verified this against the ARM backend logs via correlationId 4c07b8f1-8a29-4cfc-b93c-7a99c4a05f79 @ 2021-03-11 16:53 UTC ). I know that if I keep the MSI in the prereqs template and pass the identity name and identity rg to the main template, but move the role def & assignment to the main template akin to [this](https://github.com/Azure/azure-quickstart-templates/tree/master/201-machine-learning-dependencies-role-assignment) , it can succeed since the role def & assignment will be in the scope of the main rg. But I'm not seeing a way to have the role def & assignment be within the prereq but apply to the rg of the main template. Apologies in advance if I am overlooking something here.

bmoore-msft: Ok, this is what it looks like you need:

MSI that has contributor access to the web/fn app.

If that's correct you'd do the following:

- create MSI
- Assign MSI the built-in role of contributor on the web that's created in the sample
- 
Since the MSI has a very constrained scope *and* the scope is a new web app, I would just create it in the main template, no prereq.  Your roleAssignment in the main template will look something like this:

```
    {
      "scope": "[concat('Microsoft.Web/sites/', parameters('appName')]",      
      "type": "Microsoft.Authorization/roleAssignments",
      "apiVersion": "2020-04-01-preview",
      "name": "[guid(resourceId('Microsoft.ManagedIdentity/userAssignedIdentities', parameters('identityName')), variables('contributorRoleDefId'), parameters('appName'))]",
      "dependsOn": [
        "[resourceId('Microsoft.Web/sites', parameters('appName'))]"
      ],
      "properties": {
        "roleDefinitionId": "[subscriptionResourceId('Microsoft.Authorization/roleDefinitions', variables('contributorRoleDefId'))]",
        "principalId": "[reference(resourceId('Microsoft.ManagedIdentity/userAssignedIdentities', parameters('identityName')), '2018-11-30').principalId]",
        "principalType": "ServicePrincipal"
      }
    }
```

If that's still not working, just leave the commit in place so we can debug.

PR #8879, opened 2021-03-11
Add sample: Front Door Standard/Premium (Preview) with VM origin and Private Link service

johndowns:
# PR Checklist

## Changelog

* Add sample: Front Door Standard/Premium (Preview) with VM origin and Private Link service


bmoore-msft:
@johndowns - check this PR for updates that may be needed to documentation that references this sample.  [this is an automated message]

bmoore-msft:
@johndowns - check this PR for updates that may be needed to documentation that references this sample.  [this is an automated message]

bmoore-msft:
@johndowns - check this PR for updates that may be needed to documentation that references this sample.  [this is an automated message]

201-front-door-premium-vm-private-link/modules/front-door.bicep
```diff
@@ -0,0 +1,167 @@
...
+  }
+}
+
+resource origin 'Microsoft.Cdn/profiles/originGroups/origins@2020-09-01' = {
+  name: '${originGroup.name}/${originName}'
+  properties: {
+    hostName: originHostName
+    httpPort: 80
+    httpsPort: 443
+    originHostHeader: endpoint.properties.hostName // This probably isn't the right host header for most scenarios - you will need to look at what makes sense for your solution.
```
bmoore-msft, line 89: this seems an odd comment - can we be more prescriptive?  (or opinionated?)

201-front-door-premium-vm-private-link/modules/front-door.bicep
```diff
@@ -0,0 +1,167 @@
...
+    priority: 1
+    weight: 1000
+    sharedPrivateLinkResource: isPrivateLinkOrigin ? privateLinkOriginDetails : null
+  }
+}
+
+resource route 'Microsoft.Cdn/profiles/afdEndpoints/routes@2020-09-01' = {
+  name: '${endpoint.name}/${routeName}'
+  dependsOn: [
+    origin // This explicit dependency is required to ensure that the origin group is not empty when the route is created.
```
bmoore-msft, line 99: did bicep not create this for you? what happens when it doesn't?  (it's an odd pattern)

201-front-door-premium-vm-private-link/modules/load-balancer.bicep
```diff
@@ -0,0 +1,71 @@
...
+    loadBalancingRules: [
+      {
+        name: 'HttpRule'
+        properties: {
+          protocol: 'Tcp'
+          frontendPort: 80
+          backendPort: 80
+          idleTimeoutInMinutes: 15
+          frontendIPConfiguration: {
+            id: resourceId('Microsoft.Network/loadBalancers/frontendIpConfigurations', loadBalancerName, frontendIPConfigurationName)
```
bmoore-msft, line 55: have you tried the bicep version of the resourceId here (and below)?

loadBalancer.properties.frontendIpConfigurations.id

201-front-door-premium-vm-private-link/modules/network.bicep
```diff
@@ -0,0 +1,46 @@
...
+          addressPrefix: privateLinkServiceSubnetIPPrefix
+          privateLinkServiceNetworkPolicies: 'Disabled'
+        }
+      }
+    ]
+  }
+}
+
+output vnetName string = vnetName
+output vmSubnetResourceId string = resourceId('Microsoft.Network/virtualNetworks/subnets', vnetName, vmSubnetName)
```
bmoore-msft, line 45: here

vnet.properties.subnets[0].id

should work

bmoore-msft (changes requested):
I see there are still a few places where you're calling the resourceID function instead of the property reference - those *should* work in bicep if you're finding places where they don't we should start flagging them... (in bicep)

201-front-door-premium-vm-private-link/modules/load-balancer.bicep
johndowns: Yes - that causes a circular reference since this is part of the same resource declaration.

201-front-door-premium-vm-private-link/modules/network.bicep
johndowns: Agreed - but I'm generally not a fan of index-based properties, because it's brittle to changes in the ordering of the subnets. Referring to it by name instead of ordinal index seems more stable. What do you think?

201-front-door-premium-vm-private-link/modules/front-door.bicep
johndowns: No - this is a quirk of the Front Door ARM RP. There is a clear dependency between the route and the origin group, and also between the origin group and the origin - and Bicep detects these. However, this explicit dependency is required because Front Door has a rule that a route can't be created if the origin group it references is empty. I've spoken to the PG about this as I think it's not intuitive - but that's the behaviour they need to go with for now at least.

201-front-door-premium-vm-private-link/main.bicep
```diff
@@ -0,0 +1,105 @@
...
+param vmSubnetIPPrefix string = '10.0.0.0/24'
+
+@description('The IP address prefix (CIDR range) to use when deploying the Private Link service subnet within the virtual network.')
+param privateLinkServiceSubnetIPPrefix string = '10.0.1.0/24'
+
+@description('The name of the SKU to use when creating the virtual machine.')
+param vmSize string = 'Standard_DS1_v2'
+
+@description('The name of the publisher of the virtual machine image, such as \'MicrosoftWindowsServer\'.')
+param vmImagePublisher string = 'MicrosoftWindowsServer'
```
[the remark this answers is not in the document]
johndowns: Updated to a single object param - let me know if you want to move to the variable-based approach instead, but this seemed more configurable?

201-front-door-premium-vm-private-link/modules/front-door.bicep
johndowns: This is a fairly contrived sample (since the purpose of the sample isn't to configure a VM fully, but rather to show how Front Door works with Private Link service and VMs at a network topology level). I've rejigged the way this works to clarify that the way this is set up is specific to the way the sample has been designed.

This point is also mentioned in the README.md file too.

bmoore-msft:
@johndowns - check this PR for updates that may be needed to documentation that references this sample.  [this is an automated message]

201-front-door-premium-vm-private-link/modules/network.bicep
bmoore-msft: Very good point - does bicep allow this?

vnet.properties.subnets[vmSubnetName].id

ideally we want to make it easier to get to resourceIds in bicep (so obviate the resourceId function) but your concern trumps that imo

201-front-door-premium-vm-private-link/modules/load-balancer.bicep
bmoore-msft: do you mind opening an issue in bicep for this?  I get it, but seems solvable (and if we stick with it this way can never get rid of the resourceId fn)

201-front-door-premium-vm-private-link/modules/front-door.bicep
bmoore-msft: Thank you for trying (those conversations can be frustrating...)

201-front-door-premium-vm-private-link/modules/front-door.bicep
bmoore-msft: Should the comment be "this isn't the right host header for ....  see the readme/docs for more detail?"

The "probably" wording makes me ask "well is it or isn't it?"

201-front-door-premium-vm-private-link/main.bicep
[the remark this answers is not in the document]
bmoore-msft: Yeah, something like this sample: https://github.com/Azure/azure-quickstart-templates/blob/master/201-vm-copy-index-loops/azuredeploy.json#L73 - doesn't have to be windows/ubuntu, can be multiple windows flavors.  If folks want others, the pattern is easy to adapt.

params with interdependencies or json params end up creating a lot of deployment failures - so single params allowed values simplies it quite a bit.

201-front-door-premium-vm-private-link/modules/load-balancer.bicep
johndowns: Sure: https://github.com/Azure/bicep/issues/1852

I'll keep the current behaviour for this PR though, and can update it if/when Bicep gets this improvement.

201-front-door-premium-vm-private-link/modules/front-door.bicep
johndowns: I've updated the comment - the 'probably' is in an outdated version.

The new comment is:

> For the purposes of this sample, the 'Host' header value sent from Front Door to the origin must be the Front Door endpoint's hostname. In most real-world scenarios, you will need to set this to the originHostName value or a custom domain name, as described here: https://docs.microsoft.com/azure/frontdoor/standard-premium/concept-origin#hostheader

201-front-door-premium-vm-private-link/modules/network.bicep
johndowns: No it doesn't, unfortunately. I opened a Bicep issue for this: https://github.com/Azure/bicep/issues/1853

I'll close this for now, but if that gets fixed I'll adjust the template accordingly.

201-front-door-premium-vm-private-link/main.bicep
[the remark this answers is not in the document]
johndowns: Thanks - I have updated it to follow a similar pattern.

bmoore-msft:
@johndowns - check this PR for updates that may be needed to documentation that references this sample.  [this is an automated message]

PR #8914, opened 2021-03-22
Update sample: Front Door Standard/Premium (Preview) with App Service origin and private endpoint

johndowns:
# PR Checklist

## Changelog

* Update sample to use new Bicep syntax

### DIFF
--- a/201-front-door-premium-app-service-private-link/modules/front-door.bicep
+++ b/201-front-door-premium-app-service-private-link/modules/front-door.bicep
@@ -56,7 +56,8 @@ resource profile 'Microsoft.Cdn/profiles@2020-09-01' = {
 }
 
 resource endpoint 'Microsoft.Cdn/profiles/afdEndpoints@2020-09-01' = {
-  name: '${profile.name}/${endpointName}'
+  name: endpointName
+  parent: profile
   location: 'global'
   properties: {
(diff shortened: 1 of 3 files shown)


bmoore-msft:
@johndowns - check this PR for updates that may be needed to documentation that references this sample.  [this is an automated message]

201-front-door-premium-app-service-private-link/modules/front-door.bicep
```diff
@@ -56,7 +56,8 @@ resource profile 'Microsoft.Cdn/profiles@2020-09-01' = {
 }
 
 resource endpoint 'Microsoft.Cdn/profiles/afdEndpoints@2020-09-01' = {
```
bmoore-msft, line 58: Just noticed this....  if I have the parent property, do I still need the full resourceType?

johndowns: Yes - if you're using the parent property then you do need a fully qualified resourceType. If you use a nested resource then you don't - but stylistically I generally prefer to avoid nested resources (with a few exceptions) since they make the template harder to read.

bmoore-msft: ack - I don't like nesting either

PR #8938, opened 2021-03-29
Update template SharePoint

Yvand:
# PR Checklist

## Changelog

* Rename local admin on VM SQL/SP/FE to local-'adminUserName', to avoid using the local admin instead of the domain admin
* Set UserPrincipalName of all AD accounts
* Change the identity claim type to use the UPN in federated authentication
* Change the format of the realm / identifier in federated authentication
* Fix the reboot issue on SP and FE VMs when they join the AD domain
* Enable file sharing (on Domain network profile) on SP and FE VMs
* Setup an OIDC application in ADFS
* Add new SQL permissions to spsetup account to work with updated SPFarm resource
* Add a retry download logic to DSC resource cChocoInstaller to improve reliability
* Various improvements in DSC configurations
* Update apiVersion of ARM resources
* Replace outdated DSC module cADFS with AdfsDsc 1.1
* Update DSC module SharePointDSC from 4.3 to 4.5.1
* Update DSC module SqlServerDsc from 15.0 to 15.1.1
* Update DSC module NetworkingDsc from 8.1 to 8.2
* Update DSC module CertificateDsc from 4.7 to 5.1


sharepoint-adfs/CHANGELOG.md
```diff
@@ -1,8 +1,28 @@
 # Change log for Azure template SharePoint-ADFS
 
+## Unreleased
```
bmoore-msft, line 3: unreleased?

sharepoint-adfs/azuredeploy.json
```diff
@@ -370,13 +370,15 @@
       "spSvcPassword": "[parameters('serviceAccountsPassword')]",
       "spAppPoolUserName": "spapppool",
       "spAppPoolPassword": "[parameters('serviceAccountsPassword')]",
-      "spPassphrase": "[parameters('serviceAccountsPassword')]"
+      "spPassphrase": "[parameters('serviceAccountsPassword')]",
+      "localAdminUserName": "[concat('local-', parameters('adminUserName'))]"
```
bmoore-msft, line 374: how do users know the local user name?

sharepoint-adfs/CHANGELOG.md
Yvand: Oups... It's corrected now

sharepoint-adfs/azuredeploy.json
Yvand: in SharePoint, admins always want to sign-in as the domain account, using the local admin account is causing only confusion.
But when doing RDP, many admins set the user as "accountName" instead of "domain\accountName", which signs-in as the local admin (and it's really always undesired).
With this change, RDP will validate sign-in only if the admin explicitly set user as "domain\accountName", which I expect it to be a logical test if the attempt as "accountName" fails. Should I add a note in the README?

bmoore-msft: It seems like we should have 1 of 2 things... either we output the value so the user knows what it is (adding to readme is probably ok too) or we generate a random user name since we never expect them to use it.  

If it's a case where they could use the local account but shouldn't, then output and readme makes sense.  If it's a case where they should never user it, then random makes more sense.

Yvand: This is a great suggestion, I didn't think about this.
Looks like [function uniqueString](https://docs.microsoft.com/en-us/azure/azure-resource-manager/templates/template-functions-string?tabs=json#uniquestring) is the way to go.
I'm testing it now, if/when it works I'll push the commit

Yvand:
@bmoore-msft CI in AzQuickStarts-USGov failed because [SQL VM failed to deploy in ARM](https://dev.azure.com/azurequickstarts/azure-quickstart-templates/_build/results?buildId=49964&view=logs&j=275f1d19-1bd8-5591-b06b-07d489ea915a&t=14b22f05-782e-55fa-3e59-b66c82f514a9&l=825), IMO it can be ignored as it's not related to the template itself.

PR #9427, opened 2021-04-30

markjbrown:
## Changelog

* update api-version
* add bicep support


markjbrown:
@StephenWeatherford this job is throwing an error on the api-version which is preventing it from building successfully. I'm not sure why because this is using the same api-version for databaseAccount I'm using in other PR's and they are fine.

Nudging the job to run again. Will see if it builds successfully this time.

StephenWeatherford:
@markjbrown With current bicep (main), I definitely see the message that 'Microsoft.DocumentDB/databaseAccounts@2021-04-15' has no types.  I see that message in the bicep build in the pipeline, but it doesn't block the build (although personally I'm not convinced of that decision).

PR #9430, opened 2021-05-01

markjbrown:
## Changelog

* update api-version
* add bicep support


bmoore-msft (changes requested):
is the bicep file meant to be empty?

PR #9435, opened 2021-05-01

markjbrown:
## Changelog

* update api-version
* add bicep support

### DIFF
--- a/101-cosmosdb-sql-rbac/azuredeploy.json
+++ b/101-cosmosdb-sql-rbac/azuredeploy.json
@@ -56,7 +56,7 @@
     {
       "type": "Microsoft.DocumentDB/databaseAccounts",
       "name": "[variables('accountName')]",
-      "apiVersion": "2021-01-15",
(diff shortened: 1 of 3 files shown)


bmoore-msft:
@markjbrown - for the bicep samples, we need to update azuredeploy.json to be the one generated by bicep - IOW replace the old json with the bicep json.  That way we know the samples are the same.  We're going to automate that check once @StephenWeatherford gets back from vacation ;)

markjbrown:
@bmoore-msft So do you need me to do build the azuredeploy.json for each of these PR's or is this being built in your test scripts?

StephenWeatherford:
@markjbrown Yes please if you can.  I just restarted the validation against the new bicep pipeline (it should have been running before but wasn't), it should show the error now and what needs to be done.  Let us know if you run into problems or have questions, thanks!

markjbrown:
@StephenWeatherford @bmoore-msft 

Ack. Will run build on update all the arm templates and commit into all the PR's.

Thanks.

PR #10546, opened 2021-05-28
FIX-TEMPLATE: mysql-ha-pxc-zones

olga-spektra:
# PR Checklist

## Changelog

*
*
*


mysql-ha-pxc-zones/azuredeploy.json
```diff
@@ -4,13 +4,9 @@
   "parameters": {
     "location": {
       "type": "string",
-      "allowedValues": [
-        "westeurope",
-        "eastus2"
-      ],
-      "defaultValue": "westeurope",
+      "defaultValue": "[resourceGroup().location]",
```
bmoore-msft, line 7: I think the proper fix here is to restrict the regions to those that support zones (for both public and USGov) - this creates a defaultValue that may fail and the variable below simply overrides the user's selection (which will not be expected).

bmoore-msft (changes requested):

PR #10645, opened 2021-06-01
Add Enforce HTTPS and Disable public storage blob access.

shimonar1171:
When the customer enables VA we create a new storage account for storing the VA scan results. 
We have new requirements that this new storage account will enforce HTTPS and disable public storage blob access.
We update the Storage resource and add new properties for those requirements.

I test this template by running it manually.


shimonar1171:
@bmoore-msft , @talhers 

Can you please review my code?
Thanks.

a7mad-3bbas (approved):

PR #10794, opened 2021-06-04
Bicep samples: 101/application-gateway-v2-autoscale-create

StephenWeatherford:
Sync with bicep example: https://github.com/Azure/bicep/blob/main/docs/examples/101/application-gateway-v2-autoscale-create/main.bicep


quickstarts/microsoft.network/application-gateway-v2-autoscale-create/main.bicep
```diff
@@ -0,0 +1,172 @@
...
+  {
+    IpAddress: '10.0.0.5'
+  }
+]
+
+@allowed([
+  'Enabled'
+  'Disabled'
+])
+@description('Cookie based affinity')
```
bmoore-msft, line 42: move this above allowedValues

quickstarts/microsoft.network/application-gateway-v2-autoscale-create/main.bicep
```diff
@@ -0,0 +1,172 @@
...
+    publicIPAllocationMethod: 'Static'
+  }
+}
+
+resource applicationGateway 'Microsoft.Network/applicationGateways@2020-06-01' = {
+  name: applicationGatewayName
+  location: location
+  properties: {
+    sku: {
+      name: appGwSize
```
bmoore-msft, line 87: are both of these required?  they are set to the same value but one is hard-coded, and I suspect if the param changes this will break...

quickstarts/microsoft.network/application-gateway-v2-autoscale-create/main.bicep
```diff
@@ -0,0 +1,172 @@
...
+          cookieBasedAffinity: cookieBasedAffinity
+        }
+      }
+    ]
+    httpListeners: [
+      {
+        name: 'appGatewayHttpListener'
+        properties: {
+          frontendIPConfiguration: {
+            id: resourceId('Microsoft.Network/applicationGateways/frontendIPConfigurations', applicationGatewayName, 'appGatewayFrontendIP')
```
bmoore-msft, line 145: does the bicep symbolic reference work here?  (I'm guessing it fails with circular reference? if so we should probably file an issue since thie RP seems to require it)

quickstarts/microsoft.network/application-gateway-v2-autoscale-create/main.bicep
```diff
@@ -0,0 +1,172 @@
...
+    autoscaleConfiguration: {
+      minCapacity: minCapacity
+      maxCapacity: maxCapacity
+    }
+    gatewayIPConfigurations: [
+      {
+        name: 'appGatewayIpConfig'
+        properties: {
+          subnet: {
+            id: resourceId('Microsoft.Network/virtualNetworks/subnets', virtualNetwork.name, subnetName)
```
bmoore-msft, line 99: bicep-ify?

bmoore-msft (changes requested):

quickstarts/microsoft.network/application-gateway-v2-autoscale-create/main.bicep
StephenWeatherford: You tell me.  The docs don't say much:
image
https://docs.microsoft.com/en-us/azure/templates/microsoft.network/applicationgateways?tabs=json

This is straight from the original quickstart's json.

StephenWeatherford: image

quickstarts/microsoft.network/application-gateway-v2-autoscale-create/main.bicep
StephenWeatherford: you're saying i should convert to .id?  Not sure how to do that here.

quickstarts/microsoft.network/application-gateway-v2-autoscale-create/main.bicep
StephenWeatherford: Not sure how to do that here.

bmoore-msft: use subnet.id - I think bicep might still required the subnet to be defined as a child resource first...

quickstarts/microsoft.network/application-gateway-v2-autoscale-create/main.bicep
bmoore-msft: I don't think we can here - I'm pretty sure bicep won't allow .id on this resource type...

PR #10876, opened 2021-06-07
Update contribution guide for bicep

StephenWeatherford:


1-CONTRIBUTION-GUIDE/README.md
```diff
@@ -17,7 +17,7 @@ To deploy a sample using the Azure Portal, click the **Deploy to Azure** button
```
StephenWeatherford, line 17: I considered adding a main.bicep file to 100-blank-template, but that would be confusing because it would be empty and I'd need to show azdeploy.json as compiled from bicep.  But I have to have 100-blank-template be the example of where they start from, since we should encourage using bicep.  Do we need a separate example for a simple bicep sample?

1-CONTRIBUTION-GUIDE/README.md
```diff
@@ -83,9 +83,12 @@ To make sure your template is added to Azure.com index, please follow these guid
 
 + Guidelines on the README.md file below.
 
-1. The deployment template file must be named **azuredeploy.json**.
+1. The bicep deployment file (optional) must be named **main.bicep**
+1. The JSON deployment template file (required) must be named **azuredeploy.json**.
 1. There should be a parameters file named **azuredeploy.parameters.json**.
```
bmoore-msft, line 88: I think this is a *must* (I know it's not in your CL)

1-CONTRIBUTION-GUIDE/README.md
```diff
@@ -106,6 +109,30 @@ To make sure your template is added to Azure.com index, please follow these guid
...
+
+The decompiler is not guaranteed to produce correct Bicep code from JSON, so you will need to inspect and modify the **main.bicep** file afterwards (and then use `bicep build` to generate the matching **azuredeploy.json** file). Some general guidelines:
+
+1. Rename (F2 in VsCode) parameters and variables to be camel-cased.
+1. Rename resource symbolic names to a logical, short name, such as 'storage' or 'vmStorage' for a resource of type `Microsoft.Storage/storageAccounts`. Remove `Name` from the symbolic name if the decompiler creates it that way.
+1. Remove `_var`, `_param` and `_resource` prefixes if they are present in variables, parameters and resources.
+1. Use bicep concepts when possible
+1. See [decompiling](https://github.com/Azure/bicep/blob/main/docs/decompiling.md) for current limitations of the Bicep decompiler.
+1. See also [Best practices for Bicep](https://docs.microsoft.com/en-us/azure/azure-resource-manager/bicep/best-practices)
+
```
bmoore-msft, line 135: we should add a section for "sorting" - as we have for the json...  I think we'll have to work through ordering of the decorators (that were previously properties) but if you want to take a stab at it, that's a fine place to start...

1-CONTRIBUTION-GUIDE/README.md
bmoore-msft: Not sure I follow - why not just have a blank bicep file like we do for json? I'm not sure it helps much since we probably have an empty file, but if it serves some purpose to add something we can do that too...

bmoore-msft (changes requested):

1-CONTRIBUTION-GUIDE/README.md
StephenWeatherford: I don't follow.  This is taken care of by the 2nd item:

>> 2. The JSON deployment template file (required) must be named **azuredeploy.json**.

1-CONTRIBUTION-GUIDE/README.md
StephenWeatherford: If we have a blank bicep file, then the azdeploy.json will would theoretically need to match it (i.e. it would show the template hash/etc).  Would it be confusing if I just add a blank main.bicep but leave azdeploy.json as is?

1-CONTRIBUTION-GUIDE/README.md
StephenWeatherford: Took a stab...

1-CONTRIBUTION-GUIDE/README.md
StephenWeatherford: I went ahead and added blank main.bicep

PR #11038, opened 2021-06-10
add bicep support

markjbrown:
# PR Checklist

## Changelog

* add bicep support


bmoore-msft (changes requested):
let's move this to the quickstarts folder and remove the 101 from the folder name...

PR #11040, opened 2021-06-10
Migrate bicep example: 101/data-factory-v2-blob-to-blob-copy

StephenWeatherford:
Added bicep support to this sample by integrating bicep.main from the example in https://github.com/Azure/bicep/tree/main/docs/examples/101/data-factory-v2-blob-to-blob-copy

https://github.com/Azure/bicep/pull/3149

### DIFF
--- a/quickstarts/microsoft.datafactory/data-factory-v2-blob-to-blob-copy/main.bicep
+++ b/quickstarts/microsoft.datafactory/data-factory-v2-blob-to-blob-copy/main.bicep
@@ -1,7 +1,7 @@
 @description('Data Factory Name')
 param dataFactoryName string = 'datafactory${uniqueString(resourceGroup().id)}'
 
-@description('Location of the data factory. Currently, only East US, East US 2, and West Europe are supported.')
+@description('Location of the data factory.')
 param location string = resourceGroup().location
 
 @description('Name of the Azure storage account that contains the input/output data.')
@@ -11,8 +11,8 @@ param storageAccountName string = 'storage${uniqueString(resourceGroup().id)}'
 param blobContainerName string = 'blob${uniqueString(resourceGroup().id)}'
 
 var dataFactoryLinkedServiceName = 'ArmtemplateStorageLinkedService'
-var DataFactoryDataSetInName = 'ArmtemplateTestDatasetIn'
-var DataFactoryDataSetOutName = 'ArmtemplateTestDatasetOut'
+var dataFactoryDataSetInName = 'ArmtemplateTestDatasetIn'
+var dataFactoryDataSetOutName = 'ArmtemplateTestDatasetOut'
 var pipelineName = 'ArmtemplateSampleCopyPipeline'
 
 resource storageAccount 'Microsoft.Storage/storageAccounts@2021-04-01' = {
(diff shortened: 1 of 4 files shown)


azure-quickstarts:
@djpmsft - check this PR for updates that may be needed to documentation that references this sample.  [This is an automated message. You are receiving it because you are listed as the docOwner in metadata.json.]

StephenWeatherford:
FYI @JFolberth

azure-quickstarts:
@djpmsft - check this PR for updates that may be needed to documentation that references this sample.  [This is an automated message. You are receiving it because you are listed as the docOwner in metadata.json.]

djpmsft:
I am not on the ADF team anymore so I don't think I am the right contact here

StephenWeatherford:
@djpmsft was the sample referenced by documentation?  If not, perhaps we can just remove the ue contact.

quickstarts/microsoft.datafactory/data-factory-v2-blob-to-blob-copy/main.bicep
```diff
@@ -0,0 +1,142 @@
+@description('Data Factory Name')
+param dataFactoryName string = 'datafactory${uniqueString(resourceGroup().id)}'
+
+@description('Location of the data factory. Currently, only East US, East US 2, and West Europe are supported.')
+param location string = resourceGroup().location
```
bmoore-msft, line 5: if locations are restricted, remove the default and add allowedValues - though looking at the latest list I don't think that restriction is accurate.

quickstarts/microsoft.datafactory/data-factory-v2-blob-to-blob-copy/main.bicep
```diff
@@ -0,0 +1,142 @@
...
+    typeProperties: {
+      location: {
+        type: 'AzureBlobStorageLocation'
+        container: blobContainer.name
+        folderPath: 'input'
+        fileName: 'emp.txt'
+      }
+    }
+  }
+  dependsOn: [
```
bmoore-msft, line 68: Do we need this? Seems like we either don't need it or we need one for the blobContainer too...

Also the resource below has the same pattern and has no dependsOn property

quickstarts/microsoft.datafactory/data-factory-v2-blob-to-blob-copy/main.bicep
```diff
@@ -0,0 +1,142 @@
...
+        outputs: [
+          {
+            referenceName: DataFactoryDataSetOutName
+            type: 'DatasetReference'
+          }
+        ]
+      })
+    ]
+  }
+  dependsOn: [
```
bmoore-msft, line 138: same comment as above...

bmoore-msft (changes requested):

quickstarts/microsoft.datafactory/data-factory-v2-blob-to-blob-copy/main.bicep
StephenWeatherford: used .name notation

azure-quickstarts:
@djpmsft - check this PR for updates that may be needed to documentation that references this sample.  [This is an automated message. You are receiving it because you are listed as the docOwner in metadata.json.]

PR #11077, opened 2021-06-10
Migrate bicep example: 101/event-grid-servicebus-queue

StephenWeatherford:
Added bicep support to this sample by integrating bicep.main from the example in https://github.com/Azure/bicep/tree/main/docs/examples/101/event-grid-servicebus-queue

Corresponding bicep PR: https://github.com/Azure/bicep/pull/3168


StephenWeatherford:
@bmoore-msft I have to lower the apiversion of Microsoft.ServiceBus/namespaces to 2018-01-01-preview, or cause USGov doesn't support 2021-01-01-preview.  But if I use 2018-01-01-preview, then BP fails because it's too old.

Remove USGov support or let BP fail?

StephenWeatherford:
FYI @markusmeyer13 you're welcome to review if you'd like.

bmoore-msft:
@StephenWeatherford - two steps generally:
1. USGov should pass and BP should fail
1. Open an ICM for ServiceBus to update it's manifest in FF

StephenWeatherford:
activity

PR #11116, opened 2021-06-11
Migrate bicep example: 101/create-rg-lock-role-assignment

StephenWeatherford:
Added bicep support to this sample by integrating bicep.main from the example in https://github.com/Azure/bicep/tree/main/docs/examples/101/create-rg-lock-role-assignment

The corresponding PR for the modified bicep example is here: https://github.com/Azure/bicep/pull/3184

### DIFF
--- a/subscription-deployments/create-rg-lock-role-assignment/main.bicep
+++ b/subscription-deployments/create-rg-lock-role-assignment/main.bicep
@@ -25,8 +25,8 @@ resource newRg 'Microsoft.Resources/resourceGroups@2019-10-01' = {
 }
 
 module applyLock './applylock.bicep' = {
-  name: 'applyLock'
   scope: newRg
+  name: 'applyLock'
   params: {
     principalId: principalId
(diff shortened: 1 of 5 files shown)


subscription-deployments/create-rg-lock-role-assignment/main.bicep
```diff
@@ -0,0 +1,35 @@
...
+  location: rgLocation
+  tags: {
+    Note: 'subscription level deployment'
+  }
+  properties: {}
+}
+
+module applyLock './applylock.bicep' = {
+  name: 'applyLock'
+  scope: newRg
```
bmoore-msft, line 29: move scope above name (following 
See: https://github.com/Azure/azure-quickstart-templates/blob/master/1-CONTRIBUTION-GUIDE/best-practices.md#sort-order-of-properties for the proper sort order of properties)

bmoore-msft (changes requested):

PR #11134, opened 2021-06-13
multi-api add bicep support

markjbrown:
# PR Checklist

## Changelog

* add bicep support


markjbrown:
@bmoore-msft I tested this without the empty capabilities property but because it won't work because I use a formula for the capabilities property for the databaseAccount resource. If I remove it from line 105 it throws an exception when deployed.

Otherwise, this PR is good to go.

Thanks.

bmoore-msft:
So if the capabilities property is not present for SQL deployment fails (which I'd expect if there's no check for the property before referencing it) but it doesn't for MongoDB?  Looking at the sample as-is I'd expect it to fail if the user chooses mongo...

If the property is optional the solution would be something like:

```
capabilities: contains(apiType[api], 'capabilities') ? apiType[api].capabilities : null
```

markjbrown:
@bmoore-msft Oh ok. Advance stuff for me there :) I'll update the PR with that. Thanks

markjbrown:
Not sure why this failed. Looks temporal. Restarting.

markjbrown:
Hey @bmoore-msft I'm having trouble debugging this template. I took your advice for how to fix the capabilities property but the resulting template I built from the bicep file throws an exception.

Could you take a look at this possibly?

Thanks!

PR #11248, opened 2021-06-17
Migrate bicep example: 101/storage-blob-container

StephenWeatherford:
Added bicep support to this sample by integrating bicep.main from the example in https://github.com/Azure/bicep/tree/main/docs/examples/101/storage-blob-container

The corresponding PR for the modified bicep example is here: https://github.com/Azure/azure-quickstart-templates/pull/11248


StephenWeatherford:
FYI @leestott

PR #11391, opened 2021-06-22
Migrate bicep example: 201/web-app-asev2-create (IGNORE BEST PRACTICE FAILURES)

StephenWeatherford:
Added bicep support to this sample by integrating bicep.main from the example in https://github.com/Azure/bicep/tree/main/docs/examples/201/web-app-asev2-create

The corresponding PR for the modified bicep example is here: https://github.com/Azure/bicep/pull/3322


StephenWeatherford:
NOTE: See https://github.com/Azure/arm-ttk/issues/481 for the BP failure

StephenWeatherford:
It deploys correctly if I remove the empty workerPools property.  But bicep compile always puts in that line in the JSON, so I can't fix the BP complaint.

StephenWeatherford:
@bmoore-msft I think we should merge in as is

PR #11401, opened 2021-06-23
Migrate bicep example: 201/vmss-windows-autoscale

StephenWeatherford:
Added bicep support to this sample by integrating bicep.main from the example in https://github.com/Azure/bicep/tree/main/docs/examples/201/vmss-windows-autoscale

The corresponding PR for the modified bicep example is here: https://github.com/Azure/bicep/pull/3342

### DIFF
--- a/quickstarts/microsoft.compute/vmss-windows-autoscale/main.bicep
+++ b/quickstarts/microsoft.compute/vmss-windows-autoscale/main.bicep
@@ -201,7 +201,6 @@ resource autoScaleSettings 'microsoft.insights/autoscalesettings@2015-04-01' = {
           {
             metricTrigger: {
               metricName: 'Percentage CPU'
-              metricNamespace: ''
               metricResourceUri: vmss.id
               timeGrain: 'PT1M'
               timeWindow: 'PT5M'
@@ -220,7 +219,6 @@ resource autoScaleSettings 'microsoft.insights/autoscalesettings@2015-04-01' = {
           {
             metricTrigger: {
               metricName: 'Percentage CPU'
-              metricNamespace: ''
               metricResourceUri: vmss.id
               timeGrain: 'PT1M'
               timeWindow: 'PT5M'
--- a/quickstarts/microsoft.compute/vmss-windows-autoscale/main.bicep
+++ b/quickstarts/microsoft.compute/vmss-windows-autoscale/main.bicep
@@ -180,10 +180,6 @@ resource vmss 'Microsoft.Compute/virtualMachineScaleSets@2021-03-01' = {
       }
     }
   }
-  dependsOn: [
-    loadBalancer
-    virtualNetwork
-  ]
 }
 
 resource autoScaleSettings 'microsoft.insights/autoscalesettings@2015-04-01' = {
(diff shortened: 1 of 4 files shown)


quickstarts/microsoft.compute/vmss-windows-autoscale/main.bicep
```diff
@@ -0,0 +1,243 @@
...
+          {
+            name: nicName
+            properties: {
+              primary: true
+              ipConfigurations: [
+                {
+                  name: ipConfigName
+                  properties: {
+                    subnet: {
+                      id: '${virtualNetwork.id}/subnets/${subnetName}'
```
bmoore-msft, line 162: seems like we should be consistent with the construction of resourceIds that bicep doesn't handle for us... either use the resourceId() function or string interp...

resourceId() probably makes the most sense, consistent, familiar and intentional.

bmoore-msft (changes requested):

quickstarts/microsoft.compute/vmss-windows-autoscale/main.bicep
StephenWeatherford: Or better, maybe the subnet a resource so we can call subnet.id

bmoore-msft: that works too - though more of a refactor and causes downtime on redeploy.  We also have other resources that don't have children but still have resourceIds.

quickstarts/microsoft.compute/vmss-windows-autoscale/main.bicep
```diff
@@ -0,0 +1,241 @@
...
+                      id: resourceId('Microsoft.Network/virtualNetworks/subnets', virtualNetwork.name, subnetName)
+                    }
+                    loadBalancerBackendAddressPools: [
+                      {
+                        id: resourceId('Microsoft.Network/loadBalancers/backendAddressPools', loadBalancer.name, bePoolName)
+                      }
+                    ]
+                    loadBalancerInboundNatPools: [
+                      {
+                        id: '${loadBalancer.id}/inboundNatPools/${natPoolName}'
```
bmoore-msft, line 171: use resourceId here as above

bmoore-msft (changes requested):

quickstarts/microsoft.compute/vmss-windows-autoscale/main.bicep
StephenWeatherford: inboundNatPools doesn't seem to be a resource, so I don't think I can use resourceId(), left as is

bmoore-msft: Welcome to NRP - it is a resourceId - see: https://docs.microsoft.com/en-us/rest/api/compute/virtual-machine-scale-sets/create-or-update#virtualmachinescalesetipconfiguration or you can peek at the other samples in the repo

bmoore-msft (changes requested):

StephenWeatherford:
"id": "[resourceId('Microsoft.Network/loadBalancers/inboundNatPools', variables('loadBalancerName'), variables('natPoolName'))]"

PR #11402, opened 2021-06-23
Migrate bicep example: 201/web-app-sql-database

StephenWeatherford:
Added bicep support to this sample by integrating bicep.main from the example in https://github.com/Azure/bicep/tree/main/docs/examples/201/web-app-sql-database

The corresponding PR for the modified bicep example is here: https://github.com/Azure/bicep/pull/3343


StephenWeatherford:
BP appears to be false positive:

image
That apiVersion doesn't show up in docs or bicep

StephenWeatherford:
Do you want me to keep the newer, undoc'ed apiversion or revert to the one that fails BP?

PR #11403, opened 2021-06-23
Migrate bicep example: 301/nested-vms-in-virtual-network

StephenWeatherford:
Added bicep support to this sample by integrating bicep.main from the example in https://github.com/Azure/bicep/tree/main/docs/examples/301/nested-vms-in-virtual-network

The corresponding PR for the modified bicep example is here: https://github.com/Azure/bicep/pull/3344


StephenWeatherford:
@bmoore-msft BP fails because of Azure/arm-ttk#483
I fixed all the other BP issues I could, but can't tell if there are others (other than the nested templates using an outdated apiversion, which is a known bicep issue)

StephenWeatherford:
Should we merge it?

PR #11404, opened 2021-06-23
Migrate bicep example: 301/servicebus-namespace-vnet

StephenWeatherford:
Added bicep support to this sample by integrating bicep.main from the example in https://github.com/Azure/bicep/tree/main/docs/examples/301/servicebus-namespace-vnet

The corresponding PR for the modified bicep example is here: https://github.com/Azure/bicep/pull/3345


StephenWeatherford:
@bmoore-msft These BP errors:
```zsh
    [-] apiVersions Should Be Recent (163 ms) 
##[error]        Api versions must be the latest or under 2 years old (730 days) - API version 2017-04-01 of Microsoft.ServiceBus/namespaces is 1581 days old
        Valid Api Versions:
        2021-01-01-preview

##[warning]        The apiVersion 2021-02-01 was not found for the resource type: Microsoft.Network/virtualNetworks/subnets
##[warning]        The apiVersion 2018-01-01-preview was not found for the resource type: Microsoft.ServiceBus/namespaces/virtualnetworkrules
```

First is required for us-gov to run
The last two appear to be false positives by the TTK.  Bicep is fine with them, and they deploy fine.

PR #11757, opened 2021-08-03
New from bicep example: 101/deployment-script-with-storage

StephenWeatherford:
Created a new QuickStart sample from the bicep example in https://github.com/Azure/bicep/tree/main/docs/examples/101/deployment-script-with-storage


bmoore-msft (changes requested):
we won't be able to use the "built-in" MSI since that actually has no permission in the sub - we'll need to create a prereq template that creates the MSI

PR #11761, opened 2021-08-03
New from bicep example: 201/anchored-proximity-placement-group

StephenWeatherford:
Created a new QuickStart sample from the bicep example in https://github.com/Azure/bicep/tree/main/docs/examples/201/anchored-proximity-placement-group

### DIFF
--- a/.github/workflows/auto-fix.yml
+++ b/.github/workflows/auto-fix.yml
@@ -23,6 +23,10 @@ jobs:
       DOTNET_CLI_TELEMETRY_OPTOUT: true
 
     steps:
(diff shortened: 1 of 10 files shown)


.github/workflows/auto-fix.yml
```diff
@@ -23,6 +23,10 @@ jobs:
       DOTNET_CLI_TELEMETRY_OPTOUT: true
 
     steps:
+      - uses: actions/checkout@v2.3.4
```
bmoore-msft, line 26: this file needs to be removed from the PR

bmoore-msft (changes requested):

PR #11763, opened 2021-08-03
New from bicep example: 201/policy-azmonitor-agent-and-dcr-association

StephenWeatherford:
Created a new QuickStart sample from the bicep example in https://github.com/Azure/bicep/tree/main/docs/examples/201/policy-azmonitor-agent-and-dcr-association


bmoore-msft (changes requested):
this is a subscription scope template so needs to be manually validate (set the validationType property in metadata.json)

Will also need valid param values

PR #11770, opened 2021-08-03
New from bicep example: 301/modules-vwan-to-vnet-s2s-with-fw

StephenWeatherford:
Created a new QuickStart sample from the bicep example in https://github.com/Azure/bicep/tree/main/docs/examples/301/modules-vwan-to-vnet-s2s-with-fw


bmoore-msft (changes requested):
this is a subscription scoped template so will need to be manually validated - add the validationType property to metadata.json and supply valid params

PR #11772, opened 2021-08-03
New from bicep example: 301/sql-database-with-management

StephenWeatherford:
Created a new QuickStart sample from the bicep example in https://github.com/Azure/bicep/tree/main/docs/examples/301/sql-database-with-management


hansmbakker:
I see this is closed - what was the reason for that?

This example was referenced as a workaround for the issue in https://github.com/Azure/bicep/issues/3159 so it still contains value.

PR #11775, opened 2021-08-04
Updated azure-quickstart-templates/application-workloads/kemp/kemp-loadmaster-ha-pair/

daveRendon:
# PR Checklist

## Changelog

* updated parameters azuredeploy.json
* updated parameters azuredeploy.parameters.json
*


bmoore-msft:
@daveRendon - can you provide some bg for the change?  we generally don't allow defaultValues for usernames as it increases surface area...

bmoore-msft:
If it's required then we should remove the value and use a literal...  I *think* that may throw a BP warning but it will prevent the failure you describe.

daveRendon:
Thanks Brian @bmoore-msft, yes and will consider the literal type on the bicep template, since this is an ARM template, should this remain as modified in this PR? parameter(string)

bmoore-msft:
Not sure I follow - my understanding is that the workload *requires* the username you added.  If that name isn't used, then there's no access to the admin console.  If that's the case we should remove the parameter entirely - otherwise someone could provide a different value and regret it later.  So in bicep, just:

```bicep
adminUsername: 'bal'
```
(so the type of the literal is implicit)

Am I following this right?

daveRendon:
Yes Brian @bmoore-msft that is correct and thank you for providing clearance on the definition in the bicep file. So then I'll modify the ARM template and remove the parameter entirely and submit another PR, thanks!

bmoore-msft:
sounds good - thanks for clarifying...

PR #11787, opened 2021-08-12
Add module: Create a storage account and static website

johndowns:
# PR Checklist

## Changelog

* Add module: *Create a storage account and static website*

### DIFF
--- a/modules/Microsoft.Storage/static-website/1.0/README.md
+++ b/modules/Microsoft.Storage/static-website/1.0/README.md
@@ -15,7 +15,7 @@
 [![Deploy To Azure US Gov](https://raw.githubusercontent.com/Azure/azure-quickstart-templates/master/1-CONTRIBUTION-GUIDE/images/deploytoazuregov.svg?sanitize=true)](https://portal.azure.us/#create/Microsoft.Template/uri/https%3A%2F%2Fraw.githubusercontent.com%2FAzure%2Fazure-quickstart-templates%2Fmaster%2Fmodules%2FMicrosoft.Storage%2Fstatic-website%2F1.0%2Fazuredeploy.json)
 [![Visualize](https://raw.githubusercontent.com/Azure/azure-quickstart-templates/master/1-CONTRIBUTION-GUIDE/images/visualizebutton.svg?sanitize=true)](http://armviz.io/#/?load=https%3A%2F%2Fraw.githubusercontent.com%2FAzure%2Fazure-quickstart-templates%2Fmaster%2Fmodules%2FMicrosoft.Storage%2Fstatic-website%2F1.0%2Fazuredeploy.json)   
 
-This module creates a storage account and enables the static website feature.
+This module creates a storage account, enables the static website feature, and adds index and 404 error pages.
 
 ## Parameters
 
--- a/modules/Microsoft.Storage/static-website/1.0/README.md
+++ b/modules/Microsoft.Storage/static-website/1.0/README.md
@@ -19,20 +19,23 @@ This module creates a storage account and enables the static website feature.
 
 ## Parameters
 
-| Name | Type | Required | Description |
-| :------------- | :----------: | :----------: | :------------- |
-| location | string | No | Specifies the Azure location where the storageAccount should be created. |
-| accountName | string | No | Specifies the name of the storage account. This value must be globally unique. |
-| skuName | string | No | Specifies the SKU name for the storage account. |
-| indexDocument | string | No | Specifies the name of the index page for the static website. |
-| errorDocument404Path | string | No | Specifies the name of the error (404 error) page for the static website. |
-| supportsHttpsTrafficOnly |  bool | No | Allows https traffic only to storage service if set to true. |
+| Name                     | Type   | Required | Description                                                                                                                                                                   |
+|--------------------------|--------|----------|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| location                 | string | No       | The location into which the Azure Storage resources should be deployed. If not specified, defaults to the resource group's location.                                          |
+| accountName              | string | No       | The name of the Azure Storage account to create. This must be globally unique. If not specified, a unique name is generated.                                                  |
+| skuName                  | string | No       | The name of the SKU to use when creating the Azure Storage account. If not specified, defaults to `Standard_LRS`.                                                             |
+| indexDocumentPath        | string | No       | The name of the page to display when a user navigates to the root of your static website. If not specified, defaults to `index.htm`.                                          |
+| indexDocumentContents    | string | No       | The contents of the page to display when a user navigates to the root of your static website. If not specified, defaults to a simple welcome page.                            |
+| errorDocument404Path     | string | No       | The name of the page to display when a user attempts to navigate to a page that does not exist in your static website. If not specified, defaults to `404.htm`.               |
+| errorDocument404Contents | string | No       | The contents of the page to display when a user attempts to navigate to a page that does not exist in your static website. If not specified, defaults to a simple error page. |
+| supportsHttpsTrafficOnly | bool   | No       | Indicates whether the storage account should require HTTPS traffic. If not specified, defaults to `true`.                                                                     |
 
 ## Outputs
 
-| Name | Type | Description |
-| :------------- | :----------: | :------------- |
-| accountResourceId | string | The resource ID of the storage account. |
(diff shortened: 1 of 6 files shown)


modules/Microsoft.Storage/static-website/1.0/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+[![Deploy To Azure](https://raw.githubusercontent.com/Azure/azure-quickstart-templates/master/1-CONTRIBUTION-GUIDE/images/deploytoazure.svg?sanitize=true)](https://portal.azure.com/#create/Microsoft.Template/uri/https%3A%2F%2Fraw.githubusercontent.com%2FAzure%2Fazure-quickstart-templates%2Fmaster%2Fmodules%2FMicrosoft.Storage%2Fstatic-website%2F1.0%2Fazuredeploy.json)
+[![Deploy To Azure US Gov](https://raw.githubusercontent.com/Azure/azure-quickstart-templates/master/1-CONTRIBUTION-GUIDE/images/deploytoazuregov.svg?sanitize=true)](https://portal.azure.us/#create/Microsoft.Template/uri/https%3A%2F%2Fraw.githubusercontent.com%2FAzure%2Fazure-quickstart-templates%2Fmaster%2Fmodules%2FMicrosoft.Storage%2Fstatic-website%2F1.0%2Fazuredeploy.json)
+[![Visualize](https://raw.githubusercontent.com/Azure/azure-quickstart-templates/master/1-CONTRIBUTION-GUIDE/images/visualizebutton.svg?sanitize=true)](http://armviz.io/#/?load=https%3A%2F%2Fraw.githubusercontent.com%2FAzure%2Fazure-quickstart-templates%2Fmaster%2Fmodules%2FMicrosoft.Storage%2Fstatic-website%2F1.0%2Fazuredeploy.json)   
+
+This module creates a storage account and enables the static website feature.
+
+## Parameters
```
bmoore-msft, line 20: should we add detail to the description for:
- defaultValues (e.g. what does supportHttpsTrafficOnly default to?)
- examples of the output 

We're blazing trails here but perhaps good to work through some of these ideas for when we get to the registry...

modules/Microsoft.Storage/static-website/1.0/main.bicep
```diff
@@ -0,0 +1,91 @@
...
+}
+
+resource managedIdentity 'Microsoft.ManagedIdentity/userAssignedIdentities@2018-11-30' = {
+  name: managedIdentityName
+  location: location
+}
+
+resource roleAssignmentContributor 'Microsoft.Authorization/roleAssignments@2020-04-01-preview' = {
+  scope: storageAccount
+  name: guid(resourceGroup().id, managedIdentity.id, storageAccountContributorRoleDefinitionId)
```
bmoore-msft, line 50: since this has the scope of the storageAccount, the guid should as well, e.g.

guid(storageAccount.id, managedIdentity.id, storageAccountContributorRoleDefinitionId)

modules/Microsoft.Storage/static-website/1.0/main.bicep
```diff
@@ -0,0 +1,91 @@
...
+  properties: {
+    roleDefinitionId: storageAccountContributorRoleDefinitionId
+    principalId: managedIdentity.properties.principalId
+    principalType: 'ServicePrincipal'
+  }
+}
+
+resource roleAssignmentStorageBlobDataContributor 'Microsoft.Authorization/roleAssignments@2020-04-01-preview' = {
+  scope: storageAccount
+  name: guid(resourceGroup().id, managedIdentity.id, storageAccountStorageBlobDataContributorRoleDefinitionId)
```
bmoore-msft, line 60: same as above

modules/Microsoft.Storage/static-website/1.0/main.bicep
```diff
@@ -0,0 +1,91 @@
...
+    azPowerShellVersion: '5.4'
+    scriptContent: loadTextContent('scripts/enable-storage-static-website.ps1')
+    cleanupPreference: 'OnSuccess'
+    retentionInterval: 'PT4H'
+    arguments: '-ResourceGroupName ${resourceGroup().name} -StorageAccountName ${accountName} -IndexDocument ${indexDocument} -ErrorDocument404Path ${errorDocument404Path}'
+  }
+}
+
+output accountResourceId string = storageAccount.id
+output staticWebsiteHostName string = replace(replace(storageAccount.properties.primaryEndpoints.web, 'https://', ''), '/', '')
```
bmoore-msft, line 91: should this be hostname or absolute url?  (i.e. leave the https://)  Since we're create a web site, feels like we should leave the https://

bmoore-msft (changes requested):

modules/Microsoft.Storage/static-website/1.0/main.bicep
johndowns: I think both are valid for different use cases. I've added another output for the URL.

modules/Microsoft.Storage/static-website/1.0/main.bicep
johndowns: Fixed.

modules/Microsoft.Storage/static-website/1.0/main.bicep
johndowns: Good catch - fixed.

modules/Microsoft.Storage/static-website/1.0/README.md
johndowns: Yes, good idea - added.

modules/Microsoft.Storage/static-website/1.0/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+[![Deploy To Azure](https://raw.githubusercontent.com/Azure/azure-quickstart-templates/master/1-CONTRIBUTION-GUIDE/images/deploytoazure.svg?sanitize=true)](https://portal.azure.com/#create/Microsoft.Template/uri/https%3A%2F%2Fraw.githubusercontent.com%2FAzure%2Fazure-quickstart-templates%2Fmaster%2Fmodules%2FMicrosoft.Storage%2Fstatic-website%2F1.0%2Fazuredeploy.json)
+[![Deploy To Azure US Gov](https://raw.githubusercontent.com/Azure/azure-quickstart-templates/master/1-CONTRIBUTION-GUIDE/images/deploytoazuregov.svg?sanitize=true)](https://portal.azure.us/#create/Microsoft.Template/uri/https%3A%2F%2Fraw.githubusercontent.com%2FAzure%2Fazure-quickstart-templates%2Fmaster%2Fmodules%2FMicrosoft.Storage%2Fstatic-website%2F1.0%2Fazuredeploy.json)
+[![Visualize](https://raw.githubusercontent.com/Azure/azure-quickstart-templates/master/1-CONTRIBUTION-GUIDE/images/visualizebutton.svg?sanitize=true)](http://armviz.io/#/?load=https%3A%2F%2Fraw.githubusercontent.com%2FAzure%2Fazure-quickstart-templates%2Fmaster%2Fmodules%2FMicrosoft.Storage%2Fstatic-website%2F1.0%2Fazuredeploy.json)   
+
+This module creates a storage account and enables the static website feature.
```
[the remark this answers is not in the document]
johndowns: I've updated the readme to clarify that content is added.

I'm not sure about accepting a tree of content. It's a good idea in theory, but it would be a difficult thing to actually use in practice, I think? I've updated the module to take the contents of the index and error files from parameters.

For this specific module, I think that's the right way to go. However, we could look at adding another module that accepts a zip file URL. I think that's a distinct use case though.

What do you think?

johndowns:
@bmoore-msft Looks like the build is working again - over to you for further review/

johndowns:
@bmoore-msft Is there anything else you'd like to change in this PR?

modules/Microsoft.Storage/static-website/1.0/scripts/enable-storage-static-website.ps1
```diff
@@ -0,0 +1,16 @@
+$ErrorActionPreference = 'Stop'
```
bmoore-msft, line 1: thinking more about this as a module, this isn't terribly useful in that I just end up with a hello world website...  though on the flip side I can't think of a good way to populate it through the module...  Any thoughts on how we could use this and have the user provide "real" files for the site via a module?  via loadTextContent() maybe?  or zip as base64 content?

PR #11827, opened 2021-09-03
Fix typo VMMS to VMSS and add bicep template

jbpaux:
# PR Checklist

## Changelog

* Template was wrongly named (VMMS instead of VMSS for virtual machine scale set)
* Added bicep template
*


quickstarts/microsoft.compute/vmss-with-public-ip-prefix/main.bicep
```diff
@@ -0,0 +1,231 @@
...
+@description('Number of VM instances (100 or less).')
+@minValue(1)
+@maxValue(100)
+param instanceCount int = 5
+
+@description('Location for resources. Default is the current resource group location.')
+param location string = resourceGroup().location
+
+@description('Length of public IP prefix.')
+@minValue(28)
```
bmoore-msft, line 24: use allowedValues here, will make it easier to deploy in portal

quickstarts/microsoft.compute/vmss-with-public-ip-prefix/main.bicep
```diff
@@ -0,0 +1,231 @@
...
+          {
+            name: nicName
+            properties: {
+              primary: true
+              ipConfigurations: [
+                {
+                  name: ipConfigName
+                  properties: {
+                    subnet: {
+                      id: '${virtualNetwork.id}/subnets/${subnetName}'
```
bmoore-msft, line 197: use the resourceId() function

quickstarts/microsoft.compute/vmss-with-public-ip-prefix/main.bicep
```diff
@@ -0,0 +1,231 @@
...
+                      properties: {
+                        idleTimeoutInMinutes: 15
+                        publicIPPrefix: {
+                          id: publicIPPrefix.id
+                        }
+                      }
+                    }
+                    loadBalancerBackendAddressPools: [
+                      {
+                        id: '${loadBalancer.id}/backendAddressPools/${bePoolName}'
```
bmoore-msft, line 210: use the resourceId() function

quickstarts/microsoft.compute/vmss-with-public-ip-prefix/main.bicep
```diff
@@ -0,0 +1,231 @@
...
+                      }
+                    ]
+                    loadBalancerInboundNatPools: [
+                      {
+                        id: '${loadBalancer.id}/inboundNatPools/${natPoolName}'
```
bmoore-msft, line 215: use the resourceId() function

quickstarts/microsoft.compute/vmss-with-public-ip-prefix/main.bicep
```diff
@@ -0,0 +1,231 @@
...
+  version: 'latest'
+}
+var imageReference = osType
+var publicIPAddressName = '${vmssName}pip'
+var publicIPPrefixName = '${vmssName}pubipprefix'
+var virtualNetworkName = '${vmssName}vnet'
+var subnetName = '${vmssName}subnet'
+var addressPrefix = '10.0.0.0/16'
+var subnetPrefix = '10.0.0.0/24'
+var dnsName = 'dns${toLower(vmssName)}'
```
bmoore-msft, line 60: add a separate parameter for this, it needs to be globally unique

bmoore-msft (changes requested):

quickstarts/microsoft.compute/vmss-with-public-ip-prefix/main.bicep
jbpaux: Hi @bmoore-msft according to the [Bicep Best Practice page](https://docs.microsoft.com/en-us/azure/azure-resource-manager/bicep/best-practices), usage of resourceId and reference functions should be avoided to ensure bicep can generate the dependencies. Are you sure we want to use it ? 
> When possible, avoid using the reference and resourceId functions in your Bicep file. You can access any resource in Bicep by using the symbolic name.

jbpaux:
Hi, in addition to the review comment above, I was wondering the real value in this example to deploy a LB. The goal of this template is to show how to deploy a VMSS with a public IP Prefix so each VMSS instance has its own IP from the IP Prefix. Usage of a LB and its nat rules is useless isn't it ? You should connect directly to the VM using its IP.

bmoore-msft:
The LB allows requests to be loadbalanced, that would be the typical use of the VMSS.  You can skip that and connect directly to an instance, but that wouldn't be a different scenario...  So my understanding is that both are useful in a single sample.

jbpaux:
Yes I agree for LB rules but this sample only had NAT rules. I'll remove the nat rules because target is to use the instance public ip.

quickstarts/microsoft.compute/vmss-with-public-ip-prefix/main.bicep
jbpaux: @bmoore-msft any opinion on previous message ?

bmoore-msft:
You can leave them as is - and just address the other comments....

PR #11832, opened 2021-09-07
Download and run TemplateAnalyzer against new templates

VeraBE:
# PR Checklist

## Changelog

* Add a pipeline script to download and run [TemplateAnalyzer](https://github.com/Azure/template-analyzer) against the templates added in a PR
* Store whether the template analysis passed or failed, and a file with the full analysis output
* Add pipeline support for a TemplateAnalyzer badge 
* Mention in the contributing guidelines that this check will be run for new templates


test/ci-scripts/Write-TestResults.ps1
```diff
@@ -527,4 +567,4 @@ Snippet that will be placed in the README.md files
 <IMG SRC="https://azurequickstartsservice.blob.core.windows.net/badges/100-blank-template/BestPracticeResult.svg" />&nbsp;
 <IMG SRC="https://azurequickstartsservice.blob.core.windows.net/badges/100-blank-template/CredScanResult.svg" />&nbsp;
```
yane3628, line 568: I think you need to update the snippet that gets pushed to the README files with the new badge.

test/ci-scripts/Write-TestResults.ps1
```diff
@@ -441,6 +468,15 @@ switch ($CredScanResult) {
     }
 }
 
+switch ($TemplateAnalyzerResult) {
+    "PASS" { $TemplateAnalyzerResultColor = "brightgreen" }
```
anaismiller, line 472: Not sure what color this is and that in other areas this is used, but is this an accessible color? Does it have enough contrast?

test/ci-scripts/Write-TestResults.ps1
anaismiller: +1

test/ci-scripts/Run-TemplateAnalyzer.ps1
```diff
@@ -0,0 +1,76 @@
...
+    [string] $mainTemplateFilename = $ENV:MAINTEMPLATE_DEPLOYMENT_FILENAME,
+    [string] $mainParametersFilename = $ENV:GEN_PARAMETERS_FILENAME
+)
+
+$RULE_FAILED_MESSAGE = "Result: Failed"
+
+$templateAnalyzerFolderPath = "$ttkFolder\templateAnalyzer"
+New-Item -ItemType Directory -Path $templateAnalyzerFolderPath -Force
+Invoke-WebRequest -OutFile "$templateAnalyzerFolderPath\TemplateAnalyzer.zip" https://github.com/Azure/template-analyzer/releases/download/0.0.2-alpha/TemplateAnalyzer.zip
+# ^ will be replaced by https://github.com/Azure/template-analyzer/releases/latest/download/TemplateAnalyzer.zip after CLI changes are released
```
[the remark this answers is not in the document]
VeraBE: There was a discussion about this in the Teams channel, and Johnathon seemed convinced to not do a release now

test/ci-scripts/Run-TemplateAnalyzer.ps1
```diff
@@ -0,0 +1,76 @@
...
+        } 
+        $testOutput = & $templateAnalyzer @params
+    }
+    $testOutput = $testOutput -join "`n"
+
+    if($testOutput.length -ne 0 -and $LASTEXITCODE -eq 0)
+    {
+        $testOutput >> $testOutputFilePath
+
+        return $testOutput.Contains($RULE_FAILED_MESSAGE)
```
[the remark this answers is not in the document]
VeraBE: I'm not sure if I understand your question, the code is doing this: If (templateAnalyzer outputted something and that execution was successful) { output to file and return failed/pass } else { report error}

test/ci-scripts/Write-TestResults.ps1
```diff
@@ -128,6 +131,8 @@ $FairfaxDeployment = $FairfaxDeployment -ireplace [regex]::Escape("true"), "PASS
 $FairfaxDeployment = $FairfaxDeployment -ireplace [regex]::Escape("false"), "FAIL"
 $PublicDeployment = $PublicDeployment -ireplace [regex]::Escape("true"), "PASS"
 $PublicDeployment = $PublicDeployment -ireplace [regex]::Escape("false"), "FAIL"
+$TemplateAnalyzerResult = $TemplateAnalyzerReportedErrors -ireplace [regex]::Escape("true"), "FAIL"
```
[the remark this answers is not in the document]
VeraBE: I don't know, I liked ReportedErrors better because it seems more declarative to me than result

test/ci-scripts/Write-TestResults.ps1
bmoore-msft: you can see a sample here: https://github.com/Azure/azure-quickstart-templates/tree/master/100-blank-template - the color used for badges in github.

test/ci-scripts/Write-TestResults.ps1
bmoore-msft: We will need to, though I don't think we'll do this (and all the readme files) until we GA and figure out the TTK story.  So we don't have almost 2 identical badges

bmoore-msft (changes requested):

test/ci-scripts/Run-TemplateAnalyzer.ps1
```diff
@@ -0,0 +1,76 @@
...
+        Write-Error "TemplateAnalyzer failed trying to analyze: $templateFilePath $parametersFilePath"
+        exit 1
+    }
+}
+
+$reportedErrors = $false
+Get-ChildItem $sampleFolder -Directory | # To analyze all the JSON files in folders that could contain nested templates
+    ForEach-Object {
+        if ($_.Name -ne "prereqs") {
+            Get-ChildItem $_ -Recurse -Filter *.json |
```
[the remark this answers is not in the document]
VeraBE: Just outputs a message in stdout :/

test/ci-scripts/Run-TemplateAnalyzer.ps1
```diff
@@ -0,0 +1,76 @@
...
+
+        return $testOutput.Contains($RULE_FAILED_MESSAGE)
+    } else {
+        Write-Error "TemplateAnalyzer failed trying to analyze: $templateFilePath $parametersFilePath"
+        exit 1
+    }
+}
+
+$reportedErrors = $false
+Get-ChildItem $sampleFolder -Directory | # To analyze all the JSON files in folders that could contain nested templates
```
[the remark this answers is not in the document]
VeraBE: I read in the contributing guidelines that all the nested templates have to go in a folder called "nested", should I observe that they might be in the sample folder?

test/ci-scripts/Write-TestResults.ps1
```diff
@@ -142,6 +147,14 @@ if ($ValidationType -eq "Manual") {
     }
 }
 
+Write-Host "Uploading TemplateAnalyzer log file..."
+Set-AzStorageBlobContent -Container $TemplateAnalyzerLogsContainerName `
+    -File $TemplateAnalyzerOutputFilePath `
```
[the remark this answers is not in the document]
VeraBE: Oh, I assumed the environment was cleaned up before/after each run of the pipeline, is that not the case? If that was the case we would have one file for each run of the pipeline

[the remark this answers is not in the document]
bmoore-msft: It is cleaned up but we're putting it into the sample place in the storageAccount

test/ci-scripts/Run-TemplateAnalyzer.ps1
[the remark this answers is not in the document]
bmoore-msft: they might be - new samples will use a folder, some older samples will not

test/ci-scripts/Write-TestResults.ps1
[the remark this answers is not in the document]
VeraBE: I think I'm not understanding your point, we are creating a file per pipeline run, and uploading it to the bpa container using as file name the sample row key, so the file should be replaced by each run of the pipeline against the same sample, right?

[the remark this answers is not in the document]
bmoore-msft: Crap - you're right, I confused "File" and "Blob"...

test/ci-scripts/Run-TemplateAnalyzer.ps1
```diff
@@ -56,21 +56,22 @@ function Analyze-Template {
...
+        $passed = $passed -and $newAnalysisPassed # evaluation done in two lines to avoid PowerShell's lazy evaluation
     }
-$preReqsFolder = "$sampleFolder\prereqs"
-$reportedErrors = $reportedErrors -or (Analyze-Template "$preReqsFolder\$prereqTemplateFilename" "$preReqsFolder\$prereqParametersFilename")
-$reportedErrors = $reportedErrors -or (Analyze-Template "$sampleFolder\$mainTemplateFilename" "$sampleFolder\$mainParametersFilename")
 
-Write-Host "##vso[task.setvariable variable=templateAnalyzer.reportedErrors]$reportedErrors"
-Write-Host "##vso[task.setvariable variable=templateAnalyzer.output.filePath]$testOutputFilePath"
+Write-Host "##vso[task.setvariable variable=template.analyzer.result]$passed"
+Write-Host "##vso[task.setvariable variable=template.analyzer.output.filePath]$testOutputFilePath"
```
bmoore-msft, line 75: Do we need to write this var or can it be a static pipeline var?  Seems like it never changes so could be a param pulled from the pipeline?

test/ci-scripts/Write-TestResults.ps1
```diff
@@ -468,15 +468,6 @@ switch ($CredScanResult) {
     }
 }
 
-switch ($TemplateAnalyzerResult) {
```
bmoore-msft: I think we should keep this code - that way when we add the badges to the readme it's simply a doc change and the readme change - we won't need to change the pipeline or scripts (except for the docs below).  Rather than piecemeal the badges...

test/ci-scripts/Write-TestResults.ps1
```diff
@@ -508,10 +499,6 @@ $badges = @(
     @{
         "url"      = "https://img.shields.io/badge/Bicep%20Version-$BicepVersion-/?color=$BicepVersionColor";
         "filename" = "BicepVersion.svg"
-    },
-    @{
-        "url"      = "https://img.shields.io/badge/Template%20Analyzer%20Check-$TemplateAnalyzerReportedErrors-/?color=$TemplateAnalyzerResultColor";
```
bmoore-msft: uri should use $TemplateAnalyzerResult?  not $TemplateAnalyzerResportedErrors

test/ci-scripts/Write-TestResults.ps1
```diff
@@ -508,10 +499,6 @@ $badges = @(
...
-        "filename" = "TemplateAnalyzerResultColor.svg"
```
bmoore-msft: change the filename to just ```TemplateAnalyzerResult.svg```

bmoore-msft (changes requested):

test/ci-scripts/Run-TemplateAnalyzer.ps1
VeraBE: It depends on $ttk.folder, I didn't want to assume that one doesn't change, or that we will remember to update this one if ttk.folder changes

test/ci-scripts/Write-TestResults.ps1
VeraBE: Kind of completely unrelated, but, while making the change I noticed this:
```
switch ($BestPracticeResult) {
    "PASS" { $BestPracticeResultColor = "brightgreen" }
    "FAIL" { $BestPracticeResultColor = "red" }
    default {
        $BestPracticeResult = $na
        $BestPracticeResult = "inactive"  
    }
}
```
It's just overwriting the same variable instead of using a $...ResultColor variable like the rest, is this an issue that should be reported?

bmoore-msft: it is wrong - though there aren't really any designed cases where that would hit the default... still if you don't mind fixing that would be good.

test/ci-scripts/Run-TemplateAnalyzer.ps1
bmoore-msft: we can reference the ttk.folder (one var can reference another)... though TBH I don't think it matters where you put the file.  we would define the var in either case, since it's easier to keep track of them across tasks that way...

anaismiller (approved):

PR #11846, opened 2021-09-14
Imagebuilder/patch1

mgreenegit:
Fix for broken deployment script


bmoore-msft (changes requested):
you'll need to update the bicep source and then do ```bicep build``` to generate azuredeploy.json

mgreenegit:
@bmoore-msft I'm confused. I did that before pushing and the files are already in sync. What is the ask to build again? The files won't change, so I would have nothing to push (based on my understanding).

bmoore-msft:
You fixed the bug in azuredeploy.json directly - the fix needs to be done in the "source" which is bicep in this case.  Then build the json file using the ```bicep build``` command.  Right now, the bicep and the json are not the same.

mgreenegit:
What am I missing?

image

PR #11904, opened 2021-09-30
Added required parameter to param file

TarunSunkaraneni:


bmoore-msft:
The file already has a vmName param on line 5???

PR #12071, opened 2021-12-03
Fix auto-fix with merges

StephenWeatherford:
Use comparison against master to determine what files have changed


BenjaminFalcon (commented):
Cheque ado

PR #12095, opened 2021-12-27
Switch template to bicep, add UI definition and update

tfenster:
# PR Checklist

## Changelog

* Switch from ARM to Bicep
* Add a UI definition file
* Go from the old product name "NAV" to "BC" everywhere, including path name


demos/aci-bc/main.bicep
```diff
@@ -0,0 +1,180 @@
...
+param dnsPrefix string
+
+@description('The eMail address to be used when requesting a Let\'s Encrypt certificate')
+param letsEncryptMail string
+
+@description('Dynamics 365 Business Central Generic image (10.0.19041.985 is the version of Windows Server). See https://mcr.microsoft.com/v2/businesscentral/tags/list')
+param bcRelease string = 'mcr.microsoft.com/businesscentral:10.0.19041.985'
+
+@description('Dynamics 365 Business Central artifact URL. See https://freddysblog.com/2020/06/25/working-with-artifacts/ to understand how to find the right one.')
+param bcArtifactUrl string = 'https://bcartifacts.azureedge.net/onprem/19.2.32968.33504/w1'
```
bmoore-msft, line 15: who owns this url?  feels like it will be fragile if something changes/moves in 6 months...

bmoore-msft (changes requested):

demos/aci-bc/main.bicep
tfenster: This is the mechanism used by the Microsoft D365 Business Central development team to share the installation artifacts. My understanding is that this will remain stable while in this case version 19 is supported and possibly even longer. Maybe @freddydk, the technology evangelist of the Business Central development team responsible for those artifacts might be willing to also explain how stable those URLs are

tfenster: Thanks Freddy! @bmoore-msft, is that sufficient for you?

bmoore-msft: I think as long as we're aware of the dependency and if it breaks, we'll likely just pull the template since we won't be able to debug.

tfenster: Great, thanks. I'll put a reminder in my calendar to update that with every release wave by Microsoft (every 6 months)

PR #12151, opened 2022-02-01
Remove broken sample

bmoore-msft:
## Changelog

* removing broken sample
*
*


gruve-p:
Why was this removed @bmoore-msft  ?

bmoore-msft:
We could not get a successful deployment for the sample.

gruve-p:
But groestlcoin was working successfull  van that one be reverted?

bmoore-msft:
If you want submit a PR with a working variation of that (or just that variation) I'm happy to add it back...  We just don't have the bandwidth/expertise to debug all the community workloads.

PR #12313, opened 2022-03-17
Quickstart template to deploy an Azure Private 5G Core

rainbowFi:
# PR Checklist

## Changelog

* Quickstart template to deploy full Azure Private 5G Core.


rainbowFi:
@mumian our service is not available in westus, which is why the deployment test has failed. Is there any way to trigger the pipeline in a different region? We support westcentralus and eastus.

Not sure why we're hitting bicep warnings, but possibly because support for this API has only just been added to bicep so not all modules have updated yet? When I view the template in VSCode I see warnings of the form `Resource type "Microsoft.MobileNetwork/mobileNetworks@2022-03-01-preview" does not have types available`

rainbowFi:
Hi @bmoore-msft, mumian said you were the right person to speak to about getting pipelines run in different regions? The pipeline runs in westus but our service is not available in that region. We support westcentralus and eastus

PR #12363, opened 2022-04-05
Workspace PNA flag update in arm template and bicep

mingweihe:
# PR Checklist

## Changelog

* Updated workspace resource version from 2021-10-01 to 2022-01-01-preview
* Added publicNetworkAccess: 'Disabled'
* Removed allowPublicAccessWhenBehindVnet: false


bmoore-msft (changes requested):
Can you update the date in metadata.json?

mingweihe:
> Can you update the date in metadata.json?

Updated, thanks for the request.

mingweihe:
> Can you update the date in metadata.json?

Hi @bmoore-msft , I tried, but that also introduced lots of build errors, it requires updating versions for some other resources which I've never used before, can we please skip the metadata update?

mingweihe:
Hi @bmoore-msft , thanks for the request. I've updated metadata.json file.
The deployment in my local box by using the "az deployment" command succeeded,
The AzquickStarts-MAC verification succeeded. I believe the reason AzquickStarts-USGov pipeline failed is because of the region, for example the VM deployment in the template was failed in US west, while it's succeeded in US west2, the USGov is probably similar.
So would you like to approve the PR if everything looks good to you.
If the USGov verification is mandatory, then please kindly let me know the way to fix it, I was not able to fix the issue according to the error log there. Thanks a lot!

bmoore-msft:
We do need to work our what's going on in Fairfax - I don't think it's location since that error message is pretty clear and happens before deployment begins.  This is failing 10 minutes in on the ```machineLearningPrivateEndpoint``` resource.  First step would be to determine if the workload is supported in FF.

Also, the bicep files have some linter warnings, need to get those cleaned up as well.

mingweihe:
> 



> We do need to work our what's going on in Fairfax - I don't think it's location since that error message is pretty clear and happens before deployment begins. This is failing 10 minutes in on the `machineLearningPrivateEndpoint` resource. First step would be to determine if the workload is supported in FF.
> 
> Also, the bicep files have some linter warnings, need to get those cleaned up as well.

mingweihe:
> Can you update the date in metadata.json?

Done the update, please review again, thanks.

Issue #12449, opened 2022-05-13
GitHub Enterprise Template - incorrect `storageAccountType` value


MosJef:
[application-workloads/github-enterprise/github-enterprise](https://github.com/Azure/azure-quickstart-templates/tree/master/application-workloads/github-enterprise/github-enterprise)

### Issue Details

The current template sets `"storageAccountType": "Premium_LRS"` inside `azuredeploy.json` - this `storageAccountType` is not supported by GitHub Actions and causes the storage verification checks to fail when attempting to enable Actions.

The [Enabling GitHub Actions](https://docs.github.com/en/enterprise-server@3.1/admin/github-actions/enabling-github-actions-for-github-enterprise-server/enabling-github-actions-with-azure-blob-storage#prerequisites) documentation displays a warning regarding this configuration in the prerequisites section:

>Warning: Using the premium performance tier with a general-purpose storage account is not supported. The standard performance tier must be selected when creating the storage account, and it cannot be changed later.

This variable _should_ be: `"storageAccountType”: “Standard_LRS”`

### Repro steps
1. Deploy to Azure using template.
2. Attempt to enable GitHub Actions with the connection string to the allocated storage account.
3. Receive error: "Unable to connect to storage provider. Please check the form and try again."
4. Logs from GHES appliance show:

```shell
StatusMessage:Block blobs are not supported.
ErrorCode:BlobTypeNotSupported
ErrorMessage:Block blobs are not supported.
```
5. Forked template, updated variable to `"storageAccountType”: “Standard_LRS”`, re-deployed template with change applied. GitHub Actions enabled successfully connected to created storage account.

MosJef:
This has now been resolved in #12450.

PR #12450, opened 2022-05-13
GitHub Enterprise Template - update `storageAccountType` value

MosJef:
# PR Checklist

## Changelog

* Update `storageAccountType` from `Premium_LRS` to `Standard_LRS`
* Closes - https://github.com/Azure/azure-quickstart-templates/issues/12449

### DIFF
--- a/application-workloads/github-enterprise/github-enterprise/azuredeploy.json
+++ b/application-workloads/github-enterprise/github-enterprise/azuredeploy.json
@@ -70,7 +70,7 @@
     "addressPrefix": "10.0.0.0/16",
     "subnetName": "Subnet",
     "subnetPrefix": "10.0.0.0/24",
-    "storageAccountType": "Premium_LRS",
+    "storageAccountType": "Standard_LRS",
     "storageAccountName": "[concat(replace(replace(replace(parameters('accountPrefix'),'.',''),'_',''),'-',''), 'data')]",
     "publicIPAddressName": "[concat(replace(replace(parameters('accountPrefix'),'.',''),'_','-'), '-pub-ip')]",
     "publicIPAddressType": "Dynamic",


application-workloads/github-enterprise/github-enterprise/azuredeploy.json
```diff
@@ -70,7 +70,7 @@
     "addressPrefix": "10.0.0.0/16",
     "subnetName": "Subnet",
     "subnetPrefix": "10.0.0.0/24",
-    "storageAccountType": "Premium_LRS",
+    "storageAccountType": "Standard_LRS",
```
bmoore-msft, line 73: is the storageAccount even used?

bmoore-msft (changes requested):

application-workloads/github-enterprise/github-enterprise/azuredeploy.json
MosJef: GitHub Actions requires a storage account, I imagine this is why it was included. If GitHub Actions are not being enabled then the storage account is not required/utilised. 

Removal of the storage account completely from the template is certainly a solution. Removal will save those who aren't using Actions from extra storage account charges. But if it is to stay then it needs to be changed as the current setting is incompatible with GitHub Actions. 

The official documentation for [Enabling GitHub Actions](https://docs.github.com/en/enterprise-server@3.4/admin/github-actions/enabling-github-actions-for-github-enterprise-server/enabling-github-actions-with-azure-blob-storage) describes setting up a storage account as the first step so removal from the template seems appropriate.

bmoore-msft: If the expectation is that GH Actions will be enabled as a later/separate step - then it makes sense to remove it.  If Actions are enabled as part of the template, we can leave it - but then should should probably add a comment as to how it will be used, since it's never referenced in the template.  That would be my vote...

MosJef: Hi @bmoore-msft, after some discussions internally I'd like to request we update the storage account to a compatible one and leave it in place. 

This will allow customers to configure actions on first setup without needing to configure external storage separately. If Actions is not used, the the storage account will still be created but not used. I was under the impression this would involve unneeded cost for the user but this is not the case.

Can we please proceed with the initial change request to update this storage account line to prevent issues for new customers enabling actions. 

Thanks,

MosJef:
Duplicate comment from thread above to remove `stale` flag.

>Hi @bmoore-msft, after some discussions internally I'd like to request we update the storage account to a compatible one and leave it in place.
>
>This will allow customers to configure actions on first setup without needing to configure external storage separately. If Actions is not used, the the storage account will still be created but not used. I was under the impression this would involve unneeded cost for the user but this is not the case.
>
>Can we please proceed with the initial change request to update this storage account line to prevent issues for new customers enabling actions.
>
>Thanks

PR #12699, opened 2022-07-26
Deploy Container from Marketplace to ACR

dciborow:
# PR Checklist

Check these items before submitting a PR... 

[Contribution Guide](https://github.com/Azure/azure-quickstart-templates/blob/master/1-CONTRIBUTION-GUIDE/README.md)

[Best Practice Guide](https://github.com/Azure/azure-quickstart-templates/blob/master/1-CONTRIBUTION-GUIDE/best-practices.md)

- [ ] - Please check this box once you've submitted the PR if you've read through the Contribution Guide and best practices checklist.

## Changelog

*
*
*

PR #12750, opened 2022-08-03
Container Apps with Azure Container Registry

Gordonby:
New quickstart to demonstrate how to use a ACR from Container Apps, leveraging a User Assigned Managed Identity.

# PR Checklist


dciborow:
@Gordonby , i am trying to do something very similar over in this PR. https://github.com/Azure/azure-quickstart-templates/pull/12699
Except, instead of the images being truely public, I am trying to ingest images from the Azure Marketplace. 

I am still debugging some issues with my custom script action, but do you think it would make sense to include this directly in your bicep module you are using instead?

Gordonby:
> @Gordonby , i am trying to do something very similar over in this PR. #12699 Except, instead of the images being truely public, I am trying to ingest images from the Azure Marketplace.
> 
> I am still debugging some issues with my custom script action, but do you think it would make sense to include this directly in your bicep module you are using instead?

You're meaning your [container_deploy.sh](https://github.com/Azure/azure-quickstart-templates/pull/12699/files#diff-c2ea5a6c06c08491f6cc821c691ce6011b0e88d182ec08dbb58717510786485d) ?

This PR is relating primarily to Container Apps, the importing of images is just leveraging an existing module in the bicep registry.

Gordonby:
@bmoore-msft - I think the best practice checking in the CICD is being confused because there's a parameter called images and also an output by the same name.

This template deploys great locally, but the best-practice warning fails in the CICD.

Gordonby:
@bmoore-msft - thoughts on my last comment?

bmoore-msft:
the best practice test won't fail the pipeline (just a warning) but the build has been deleted so can't see the error... I'll start another build.

bmoore-msft:
@Gordonby - can you check the deployment failure?

Gordonby:
> @Gordonby - can you check the deployment failure?

### 1
> Microsoft.Authorization/roleAssignments uses a preview version ( 2020-10-01-preview ) and there are more recent versions available. Line: 114, Column: 8

Ok, new error - easy to fix....

### 2 
> [error]        Type Mismatch: Parameter 'images' in nested template 'importContainerImage' is defined as array, but the parent template defines it as string). Line: 352, Column: 14

This was the error message that was confusing me before. Images is a parameter in the bicep registry module. It's defined as array.

### 3
> Function 'format' found within 'acrHostedImageUri'

Not sure why using format is a bad thing...? 
I'm using it to manipulate the image string from the public registry to the private one that its now in.

image

Gordonby:
@bmoore-msft 

2 - ok
3 - URI sounds great, but i'm not sure if it's entirely geared towards a container registry 

> The template language function 'uri' expects its first argument to be a uri string. The provided value is 'crijuu6sairdtri.azurecr.io'

bmoore-msft:
Try: 

```bicep
  acrHostedImageUri : uri('https://${acr.properties.loginServer}', string(skip(image, indexOf(image, '/'))))
```

The last `string()` requirement seems like a bicep bug - but should work in any case...

Gordonby:
@bmoore-msft i think this is about ready now

PR #12888, opened 2022-09-30
Add darktrace vsensor standalone template

menev-dt:
# PR Checklist

## Changelog

*
*
*


menev-dt (commented):
`Generate new key pair` from a template is now working. This update is to remove the `Admin user public key for ssh authentication` from the `Considerations` section and update `azuredeploy.json` accordingly.

The architecture diagram is updated with the new company logo.

application-workloads/darktrace-vsensor-standalone/azuredeploy.json
```diff
@@ -77,6 +77,7 @@
             }
         },
         "adminPublicKey": {
+            "defaultValue": "",
```
bmoore-msft, line 80: if this is indeed optional, then we need to remove the param from the param file for the sample so the default is tested...

bmoore-msft (changes requested):

application-workloads/darktrace-vsensor-standalone/azuredeploy.json
menev-dt: thanks, that's now done

bmoore-msft: Looks like (according to CI) something still isn't right... failed on the missing key.  Is there another property that needs to be set?  otherwise the template is going to send an empty string for the value.

menev-dt: azuredeploy.json works - I tested it in 'Template deployment (deploy using custom templates) '. 

All three scenarios worked:
-  Generate new key pair;
- Use existing key stored in Azure;
- Use existing public key;

The only error I could see was with 'Use existing key stored in Azure'  and that could be caused by the empty string to the value:       

image


Once a key is selected it works:         

image


The template **before** the change works the same way  (same error) with 'Use existing key stored in Azure' and it did not fail the CI.

So, it looks like that if `"defaultValue": ""` in 'new' azuredeploy.json causes this (which probably the case is):

```
    "adminPublicKey": {
      "defaultValue": "",
...
      }
```

then `"value": "GEN-SSH-PUB-KEY"` in the 'old' azuredeploy.parameters.json does the same.

The only thing I can think of is to keep `"defaultValue": ""` in 'new' azuredeploy.json, and to revert azuredeploy.parameters.json back to have ` "adminPublicKey": {"value": "GEN-SSH-PUB-KEY"}` so the pipe line won't fail.

Any other suggestions?

Gordonby:
@menev-dt  - bump the dateUpdated field in your metadata.json please 🐱 
https://github.com/Azure/azure-quickstart-templates/blob/9007e7bec9944d368586ba52ae8398ee172edc39/application-workloads/darktrace-vsensor-standalone/metadata.json#L7

menev-dt:
> @menev-dt - bump the dateUpdated field in your metadata.json please 🐱
> 
> https://github.com/Azure/azure-quickstart-templates/blob/9007e7bec9944d368586ba52ae8398ee172edc39/application-workloads/darktrace-vsensor-standalone/metadata.json#L7

done, thanks

application-workloads/darktrace-vsensor-standalone/azuredeploy.json
bmoore-msft: A couple things:
1) portal doesn't use the parameter files when deploying (unless you manually submit them) - the param files in the repo are only used for testing, they are not used by any other "client"
2) if something isn't working correctly with the portal's handling of the parameter in the deployment experience (and the template works correctly here) then it's a portal issue - they've had trouble with trying to craft and experience around SSH keys in the past

application-workloads/darktrace-vsensor-standalone/azuredeploy.parameters.json
```diff
@@ -34,9 +34,6 @@
         },
         "osSensorHMACToken": {
             "value": "osSensorHMACToken-example"
-        },
```
bmoore-msft, line 37: you'll need to revert this change

menev-dt: change reverted

PR #12920, opened 2022-10-14
Container App Dapr update

Gordonby:
# PR Checklist

## Changelog

* Using new bicep registry module versions, which should pass the AQST best practice checks
*
*


Gordonby:
Still getting best practice violations😿

> ##[error]        Cannot bind argument to parameter 'Match' because it is an empty string.
> ##[error]        Type Mismatch: Parameter 'daprComponentScopes' in nested template '[parameters('containerAppEnvName')]' is defined as array, but the parent template defines it as string). Line: 133, Column: 14

bmoore-msft:
> Still getting best practice violations😿
> 
> > ##[error]        Cannot bind argument to parameter 'Match' because it is an empty string.
> > ##[error]        Type Mismatch: Parameter 'daprComponentScopes' in nested template '[parameters('containerAppEnvName')]' is defined as array, but the parent template defines it as string). Line: 133, Column: 14
> 
> image
> 
> Any guidance on getting the best-practices-violations label removed @bmoore-msft ?

I think it's just a case where the test is not checking the type casting...  you could skip the casting and use the same type but certainly not something we'll block on (since it's a bug in the TTK)

PR #13015, opened 2022-12-01
[Machine Learning workspace] - add Bicep and update API versions

mumian:
# PR Checklist

## Changelog

*
*
*


quickstarts/microsoft.machinelearningservices/machine-learning-workspace/main.bicep
```diff
@@ -0,0 +1,114 @@
...
+      family: 'A'
+    }
+    accessPolicies: []
+    enableSoftDelete: true
+  }
+}
+
+resource applicationInsight 'Microsoft.Insights/components@2020-02-02' = {
+  name: applicationInsightsName
+  location: (((location == 'eastus2') || (location == 'westcentralus')) ? 'southcentralus' : location)
```
bmoore-msft, line 83: I know this isn't really the point of your PR, but I think we need to retire this pattern, here's what we should do instead: https://github.com/Azure/azure-quickstart-templates/commit/f8c766ab21d9b12e62c8778e6a696e0b57aa1aa6

For the most part AI is available everywhere, but sets are actually intersecting and the intersecting part is shrinking.  So eventually not giving user control over this will cause these to fail and they won't be able to fix it without modifying.  I'd like to get it while we're going after these...

bmoore-msft (changes requested):

PR #13045, opened 2022-12-09
Update metadata.json to generate azuredeploy.json

bmoore-msft:
azuredeploy.json failed to generate on previous push


bmoore-msft:
@mumian - we still do need to add the bicep badge manually... only the YAML is automatically updated in the README.md...

I added it to this one, running ci now...

PR #13403, opened 2023-05-31
Removing Linux DSCv1 sample template

mgreenegit:
## Changelog

As documented here:
[https://learn.microsoft.com/azure/virtual-machines/extensions/dsc-linux](https://learn.microsoft.com/azure/virtual-machines/extensions/dsc-linux)

We will be shifting these samples to the machine configuration feature in Azure/Arc.


mgreenegit:
#sign-off

PR #13738, opened 2024-01-31
test migration to mc from dsc ext

mgreenegit:
## Changelog

* removes DSC extesion from template
* removes DSC v1 sample
* replaces DSC sample based on machine configuration format
* replaces template content with machine configuration sample


mgreenegit:
#sign-off
@alex-frankel

PR #13916, opened 2024-03-25
Add a database watcher with new data store sample

dimitri-furman:
# PR Checklist

## Changelog

* Add a sample that creates a new database watcher with a new data store in Azure Data Explorer.

### DIFF
--- a/quickstarts/microsoft.databasewatcher/watcher-new-datastore/azuredeploy.json
+++ b/quickstarts/microsoft.databasewatcher/watcher-new-datastore/azuredeploy.json
@@ -1,0 +1,383 @@
+{
+    "$schema": "https://schema.management.azure.com/schemas/2019-04-01/deploymentTemplate.json#",
+    "contentVersion": "1.0.0.0",
+    "parameters": {
+        "name": {
+            "type": "string",
+            "metadata": {
+                "description": "Database watcher name"
+            }
+        },
+        "identityType": {
+            "type": "string",
+            "metadata": {
+                "description": "The type of managed identity to assign to a watcher"
+            }
+        },
+        "location": {
+            "type": "string",
+            "defaultValue": "[resourceGroup().location]",
+            "metadata": {
+                "description": "The location (Azure region) of the watcher"
+            }
+
+        },
+        "createNewDatastore": {
+            "type": "bool",
+            "metadata": {
+                "description": "Set to true to create a new Azure Data Explorer cluster and database as the data store for the watcher"
+            }
+        },
+        "kustoOfferingType": {
+            "type": "string",
+            "defaultValue": "adx",
+            "metadata": {
+                "description": "The Kusto offering type of the data store. Supported values are: adx, free, fabric."
+            }
+        },
+        "clusterName": {
+            "type": "string",
+            "metadata": {
+                "description": "The name of the Azure Data Explorer cluster"
+            }
+        },
+        "databaseName": {
+            "type": "string",
+            "defaultValue": "database-watcher-data-store",
+            "metadata": {
+                "description": "The name of the Azure Data Explorer database"
+            }
+        },
+        "clusterSkuName": {
+            "type": "string",
+            "metadata": {
+                "description": "The SKU of the Azure Data Explorer cluster"
+            }
+        },
+        "clusterSkuTier": {
+            "type": "string",
+            "metadata": {
+                "description": "The SKU tier of the Azure Data Explorer cluster"
+            }
+        },
+        "databaseWatcherApiVersion": {
+            "type": "string",
+            "metadata": {
+                "description": "The API version to use with the Microsoft.DatabaseWatcher resource provider"
+            }
+
+        },
+        "targetCount": {
+            "type": "int",
+            "metadata": {
+                "description": "The total number of SQL targets to add to a watcher"
+            }
+        },
+        "targetProperties": {
+            "type": "array",
+            "metadata": {
+                "description": "An array containing all SQL targets to add to a watcher"
+            }
+        }
+    },
+    "resources": [
+        {
+            "type": "Microsoft.Resources/deployments",
+            "apiVersion": "2022-09-01",
+            "name": "[parameters('clusterName')]",
+            "properties": {
+                "mode": "Incremental",
+                "expressionEvaluationOptions": {
+                    "scope": "inner"
+                },
+                "parameters": {
+                    "location": {
+                        "value": "[parameters('location')]"
+                    },
+                    "createNewDatastore": {
+                        "value": "[parameters('createNewDatastore')]"
+                    },
+                    "clusterName": {
+                        "value": "[parameters('clusterName')]"
+                    },
+                    "clusterSkuName": {
+                        "value": "[parameters('clusterSkuName')]"
+                    },
+                    "clusterSkuTier": {
+                        "value": "[parameters('clusterSkuTier')]"
+                    }
+                },
+                "template": {
+                    "$schema": "https://schema.management.azure.com/schemas/2019-04-01/deploymentTemplate.json#",
+                    "contentVersion": "1.0.0.0",
+                    "parameters": {
+                        "location": {
+                            "type": "string"
+                        },
+                        "createNewDatastore": {
+                            "type": "bool"
+                        },
+                        "clusterName": {
+                            "type": "string",
+                            "defaultValue": ""
+                        },
+                        "clusterSkuName": {
+                            "type": "string",
+                            "defaultValue": ""
+                        },
+                        "clusterSkuTier": {
+                            "type": "string",
+                            "defaultValue": ""
+                        }
+                    },
+                    "resources": [
+                        {
+                            "condition": "[equals(parameters('createNewDatastore'), bool('true'))]",
+                            "type": "Microsoft.Kusto/clusters",
+                            "apiVersion": "2023-05-02",
+                            "name": "[parameters('clusterName')]",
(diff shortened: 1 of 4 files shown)


dimitri-furman:
Automatic deployment is failing because the subscription used by the CI engine is not registered with the  `Microsoft.DatabaseWatcher` RP. This is a new RP for a service that just went to public preview.

@alex-frankel - could you help please? Can the RP be registered in the subscription?

alex-frankel:
Hi Dimitri are you the owners of this RP? RPs should be auto-registered as part of a Template/Bicep deployment. If that's not happening, there may be something wrong with the RP implementation.

dimitri-furman:
Yes, we are the owners of the RP.

When creating a new resource from this RP in the portal, we see auto-registration happen if the user is an Owner or Contributor on the subscription, as documented [here](https://learn.microsoft.com/en-us/azure/azure-resource-manager/management/resource-providers-and-types#register-resource-provider). Does the CI engine have membership in the Owner or Contributor role on the subscription?

alex-frankel:
The CI service principal has Contributor rights on the subscription.

dimitri-furman:
This [article](https://learn.microsoft.com/en-us/azure/azure-resource-manager/troubleshooting/error-register-resource-provider) says:

> Most providers are registered automatically by the Microsoft Azure portal or the command-line interface, **but not all**. If you haven't used a particular resource provider before, you might need to register that provider.

It doesn't provide additional details on when auto-registration might occur or how to influence it.

To get us unblocked, would it be possible to register `Microsoft.DatabaseWatcher` manually in this subscription?

alex-frankel:
Hi Dmitri - before jumping to this workaround, can you try posting on the internal Stack Overflow to get guidance on how to resolve this?
https://eng.ms/docs/products/arm/troubleshooting/stackoverflow_guide

Once we are working on solving the root cause, I'll be happy to register the RP manually.

masree:
Interesting, ARM template is not auto-registering our RP - I'll look up ARM docs to see if we are missing any steps. In the meantime, I've posted a SO question for help - https://stackoverflow.microsoft.com/questions/394671

alex-frankel:
Thank you for working on this. The RP is now registered and I am re-running the CI.

masree:
@alex-frankel we created [ICM ](https://portal.microsofticm.com/imp/v3/incidents/incident/486943317/summary)on ARM team and it seems like CI principal might not have permission for register api. Can you check on your side? 

`The service principal 5991c0de-1362-4164-8ac0-0ab3b91175ee does not have the RBAC permissions to call register API.`

dimitri-furman:
@alex-frankel the last CI attempt failed saying that the deployment didn't complete within the timeout period. But looking at the logs, I see that all deployments we do in the template completed successfully in about 15 minutes. After that the pipeline keeps "Checking deployment status" for a couple of hours before failing. Are there any known CI issues around this?

Deploying the template with the same parameter file (replacing UNIQUE placeholders with values) from a custom deployment in the portal completed successfully.

alex-frankel:
@dimitri-furman - at first glance, it sounds like an issue with the RP not reaching a terminal status with the given parameter values that were provided. It does not sound like a CI issue. Can you have your team check the correlation ID (ce48a3a1-b36f-474c-8c6e-f59a12195af7) of the failed deployment to see if your RP logs have anything that could identify the root cause?

quickstarts/microsoft.databasewatcher/watcher-new-datastore/azuredeploy.json
```diff
@@ -0,0 +1,383 @@
...
+                "name": "targetcopy",
+                "count": "[length(range(0, parameters('targetCount')))]"
+            },
+            "type": "Microsoft.DatabaseWatcher/watchers/targets",
+            "apiVersion": "[parameters('databaseWatcherApiVersion')]",
+            "name": "[concat(parameters('name'), '/', guid(resourceGroup().id, parameters('name'), string(copyIndex('targetcopy'))))]",
+            "dependsOn": [
+                "[parameters('name')]"
+            ],
+            "properties": "[json(parameters('targetProperties')[copyIndex('targetcopy')])]"
```
alex-frankel, line 379: Why is the whole properties body being passed as a parameter value? The individual properties should be declared and then the values should be parameterized. If you want you can use an `object` parameter type if you only want one parameter, but as it is currently authored this is an anti-pattern that will be confusing for any consumer of this template.

quickstarts/microsoft.databasewatcher/watcher-new-datastore/azuredeploy.parameters.json
```diff
@@ -0,0 +1,44 @@
...
+        "clusterSkuTier": {
+            "value": "Standard"
+        },
+        "databaseWatcherApiVersion": {
+            "value": "2023-09-01-preview"
+        },
+        "targetCount": {
+            "value": 1
+        },
+        "targetProperties": {
```
alex-frankel, line 38: if you address my other comment and don't pass the entire properties body like this, then you can still use `GEN-UNIQUE` as you were before.

alex-frankel (changes requested):
Please address comments

quickstarts/microsoft.databasewatcher/watcher-new-datastore/azuredeploy.json
dimitri-furman: Fixed

quickstarts/microsoft.databasewatcher/watcher-new-datastore/azuredeploy.parameters.json
```diff
@@ -37,7 +37,13 @@
         },
         "targetProperties": {
             "value": [
-                "{\"targetAuthenticationType\":\"Aad\",\"connectionServerName\":\"quickstart-e1a0987a-54ee-46e0-b76a-12f36c835fe0.database.windows.net\",\"targetType\":\"SqlDb\",\"sqlDbResourceId\":\"/subscriptions/3b880560-bc3b-4d17-ab39-d37a1277b841/resourceGroups/quickstart-d2e98952-a48a-4128-b5d5-920f5c64c514/providers/Microsoft.Sql/servers/quickstart-e1a0987a-54ee-46e0-b76a-12f36c835fe0/databases/quickstart-database\",\"readIntent\":false}"
+                {
+                    "targetType": "SqlDb",
+                    "sqlDbResourceId": "/subscriptions/GEN-GUID/resourceGroups/GEN-UNIQUE/providers/Microsoft.Sql/servers/GEN-UNIQUE/databases/GEN-UNIQUE",
```
alex-frankel, line 42: This is not going to work. `GEN-*` can only be a parameter value that you use in the template. So to achieve this you would need something like the following:

```json
"serverName": {
  "value": "GEN-UNIQUE"
},
"targetProperties": {
  "value": {
    "targetType": "SqlDb",
    ...
   "connectionServerName": "[concat(parameters('serverName'), 'database.windows.net')]"
  }
}
```

Separately, you should not be generating subscription GUIDs and resourceGroup names randomly. You should instead use the [`resourceId()`](https://learn.microsoft.com/en-us/azure/azure-resource-manager/templates/template-functions-resource#resourceid) function inside of the template instead of constructing it yourself as a parameter value. The `resourceId()` function will use the subscription and resourceGroup context of the current deployment unless you need to override, but even then it should not be random.

quickstarts/microsoft.databasewatcher/watcher-new-datastore/azuredeploy.json
```diff
@@ -376,7 +376,7 @@
             "dependsOn": [
                 "[parameters('name')]"
             ],
-            "properties": "[json(parameters('targetProperties')[copyIndex('targetcopy')])]"
+            "properties": "[parameters('targetProperties')[copyIndex('targetcopy')]]"
```
alex-frankel, line 379: Instead of injecting both property name and value, you should only provide **_values_** with the parameters. Users consuming the template need to be able to see the properties being declared. In this case something like the following (assuming `targetProperties` changes to an object instead of an array with only one object in it):

```json
"properties": {
  "targetType": "[parameters('targetProperties').targetType]"
  "targetAuthenticationType": "[parameters('targetProperties').targetAuthenticationType]"
  "targetResourceId": "[resourceId('microsoft.sql/servers/databases', parameters('serverName'), parameters('databaseName'))]",
  ...
}
```

I wrote these expressions freehand, so they probably have errors, but hopefully this gets you on the right track

alex-frankel (changes requested):
New comments

quickstarts/microsoft.databasewatcher/watcher-new-datastore/azuredeploy.json
dimitri-furman: The nuance here is that there might be multiple targets in the `targetProperties` array, each with their own distinct set of properties that would need to be parameterized. Is the pattern for that to introduce numbered parameters, i.e. `targetType1`, `targetType2`, etc?

While in the current example there is only one target, I would like to expand it later to show how to add multiple targets in one deployment.

alex-frankel: In that case, it is ok to have an array of objects that you iterate through, but it would be the same principal -- only parameterize the values, not the property names. You can do that in an ARM Template with the `copy` property and `copyIndex()` functions. TBH, that is not very fun ARM Template code to write, so I would strongly encourage using Bicep instead, but that is up to you :)

alex-frankel:
Looks great - thank you for the iterations on this one

PR #14429, opened 2025-05-01
Feat/browserbox azure

o0101:
# PR Checklist

- [x] Read through the Contribution Guide and best practices checklist.

## Changelog

* Add license key field to match current BrowserBox requirement
* Add stop link for easy shutting down of BrowserBox app
* Update copy and versions
* Update OS versions for deployments to latest stable versions

Previous related PR #14344 included these changes as well as changing two subtree paths by correcting `./dosyago` -> `./dosaygo` falling foul of CI rules. Thanks @alex-frankel for the guidance to split the changes into two separate PRs, one for the BrowserBox updates (this one), and a separate one for the directory name change.

PR #14344, opened 2025-02-28
Feat/browserbox azure

o0101:
# PR Checklist

- [x] Read through the Contribution Guide and best practices checklist.

## Changelog

* Add license key field to match current BrowserBox requirement
* Add stop link for easy shutting down of BrowserBox app
* Update copy and versions
* Update OS versions for deployments to latest stable versions

PR #14430, opened 2025-05-01
Supercedes #14344, removes old dosyago name 

o0101:
# PR Checklist

- [x] Read through the Contribution Guide and best practices checklist.

## Changelog

* Remove incorrect folder name.

Previous related PR #14344 included these changes (changing two subtree paths by correcting `./dosyago` -> `./dosaygo`) falling foul of CI rules. Thanks @alex-frankel for the guidance to split the changes into two separate PRs, one for the BrowserBox updates (#14429 ), and a separate one (this one) for deleting the old directory name.


o0101:
Hey @alex-frankel I know you're busy so I'm sorry for asking, but I think the reason the MAC checks fail is because this PR deletes the old name of the folder. 

The reason I deleted rather than renamed is because if you rename, the check fails with "PR can only update a single folder" or similar error. Would it be impossible for you to review? It's okay if you're too busy, just thanks for reading!

o0101:
Hey @jim-minter not sure if I can ask you for help here, and it looks like you're super busy, so sorry for the additional request but I notice you in the other threads and I'm not sure how to fix this. 

I want to update the folder name from dosyago -> dosaygo. But the CI will break on that as it changes 2 subtrees/folders. @alex-frankel says to make 2 PRs, one to remove, one to add. But it seems this one, the remove one, also breaks CI, because CI doesn't expect a folder to be removed. 

I'm sure there must be a way there but I don't know how, would it be impossible for you to help?

PR #14465, opened 2025-05-16
Fix parameter names in internal load balancer quickstarts

ateaheko:
Changed parameter name `lbPublicIPAddress` -> `lbFrontendIPAddress` for internal Load Balancer examples (ARM template and bicep)  
Updated descriptions for the parametes

# PR Checklist

Check these items before submitting a PR... 

[Contribution Guide](https://github.com/Azure/azure-quickstart-templates/blob/master/1-CONTRIBUTION-GUIDE/README.md)

[Best Practice Guide](https://github.com/Azure/azure-quickstart-templates/blob/master/1-CONTRIBUTION-GUIDE/best-practices.md)

- [x] - Please check this box once you've submitted the PR if you've read through the Contribution Guide and best practices checklist.

## Changelog

*
*
*

PR #14475, opened 2025-05-19
Revert "Fix parameter names in internal load balancer quickstarts"

mbender-ms:
Reverts Azure/azure-quickstart-templates#14465


mbender-ms:
@alex-frankel The removal of the ARM templates has broken builds in azure-docs-pr and we are in a build freeze due to Microsoft Build.

I will update the template later but the ARM templates needs to be restored ASAP. Thanks!

mbender-ms:
#sign-off

PR #14701, opened 2026-04-09
Update Discovery chat model default to gpt-5.2

hectoralinares:
The Discovery Engine requires a chat model deployment named gpt-5-2 (model: gpt-5.2) for task validation. Update the default from gpt-5.1 to gpt-5.2 so new deployments work with the Discovery Engine out of the box.


alex-frankel (approved):
Old-rules compliant; AzQuickStarts-MAC passed. Bumps Discovery chat model defaults from gpt-5.1 to gpt-5.2 in main.bicep (+2/-2).

